PR #42, opened 2014-01-16
Added a default value of 'log' for all logger annotations.

ymauray:
All log annotations generate a field named "log". 
In order to override this name, I added the following to each annotation :
```java
String value() default "log";
```
I then used this value instead of the hardcoded "log".
The default value should make this compatible with existing code.

### DIFF
--- a/src/core/lombok/eclipse/handlers/HandleLog.java
+++ b/src/core/lombok/eclipse/handlers/HandleLog.java
@@ -48,7 +48,7 @@ public class HandleLog {
 		throw new UnsupportedOperationException();
 	}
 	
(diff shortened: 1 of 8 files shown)


rzwitserloot:
The problem with this is that the overwhelming majority of use cases involve changing _ALL_ the names of _EVERY_ logger, not just one of them. That implies that this value is itself going to turn into boilerplate.

Because of that we never added it and never intend to add it, but we do cater to the above use case via the lombok.config system which has been added after your pull request.

Put this:

    lombok.log.fieldName = foobar

which is equivalent to a hypothetical @Slf4j("foobar") in all source files in that directory and all subdirectories.

PR #46, opened 2014-04-08
Field name constants

cheelio:
I've created a new annotation that generates string constants for each field in a class. The constant contains the name of the field. This is highly usable when you serialize to json (with jackson). You van now reference a json field using the generated constants.

### DIFF
--- a/src/core/lombok/eclipse/handlers/HandleFieldNameConstants.java
+++ b/src/core/lombok/eclipse/handlers/HandleFieldNameConstants.java
@@ -1,3 +1,24 @@
+/*
+ * Copyright (C) 2012-2014 The Project Lombok Authors.
+ * 
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ * 
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ * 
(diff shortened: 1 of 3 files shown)


btiernay:
Does this work with inherited fields? If so, what about access modifiers?

gkozyryatskyy:
Very useful for me. Why it is still not merged?

paulroemer:
> But until that day, I really like this patch :)

Sounds like @rzwitserloot also agrees that the patch is worth the effort. Besides voting for it, what else can we do? Looks like there is just a conflict in the AUTHORS file...

wpongra:
Same here, would like this feature very much.

Odyldzhon:
Very useful for me. Is there any news?

boyun-yuriy:
+1

nikowitt:
+1

evser:
Why not yet merged?

araghava92:
Can someone please resolve this conflict. It silly that this is stopping the merge

huangd0ng:
This is quite useful +1

eximius313:
+1

eximius313:
Oldest and second-highest (I checked :D) voted of all open PRs and still not merged :(

auror:
Would love to have this feature live!! Any live date yet?

f1ztech:
+1, merge please

rzwitserloot:
Because it needed considerable code review and needed lots of tests.

Actually, because I thought @rspilker was in charge of pull requests and he thought I was. Whoopsie, our bad. We're working on it now.

bernhardkern:
OMG, there is still activity here. Yes. Please merge... :+1:

rzwitserloot:
@btiernay It does not work with inherited fields. Put the annotation on the parent class. If you don't control it, not much we can do (asking for the fields of the parent and then generating some fields in the current class is a complicated chicken-and-egg problem which can be solved but it is a high maintenance burden and has considerable impact on the performance of your compilation run. We don't introduce such features (features that require 'resolution') lightly).

PR #49, opened 2014-06-01
Fixed version of 842a475: Remove a wrong test for copyAnnotations return...

Maaartinus:
Concerning 842a475: You've misunderstood me. My point was that copyAnnotations _itself_ can return null and then you'll get an NPE in testing for zero length. But you were right with reverting it, as I made a mistake: copyAnnotations could return _both_ null and a zero-length array! So a null-safe emptiness test would be needed. That's plain wrong.

So I sort of un-reverted it and made `copyAnnotations` never return a zero-length array, so we don't need the test.

IMHO every method should do exactly what's needed and Eclipse being insane should not make us insane.

OK, FORGET IT. I'm pretty sure, I'm right, but it's not that important.

### DIFF
--- a/src/core/lombok/eclipse/handlers/EclipseHandlerUtil.java
+++ b/src/core/lombok/eclipse/handlers/EclipseHandlerUtil.java
@@ -1099,14 +1099,28 @@ public class EclipseHandlerUtil {
 		
 		return MemberExistsResult.NOT_EXISTS;
 	}
-	
+
+	/**
+	 * Wrapper for {@link #methodExists(String, EclipseNode, boolean, int, boolean)} with {@code caseSensitive} = {@code true} and {@code honorTolerate} = {@code false}.
+	 */
+	public static MemberExistsResult methodReallyExists(String methodName, EclipseNode node, int params) {
+		return methodExists(methodName, node, true, params, false);
+	}
+
 	/**
 	 * Wrapper for {@link #methodExists(String, EclipseNode, boolean, int)} with {@code caseSensitive} = {@code true}.
 	 */
 	public static MemberExistsResult methodExists(String methodName, EclipseNode node, int params) {
 		return methodExists(methodName, node, true, params);
 	}
-	
+
+	/**
+	 * Wrapper for {@link #methodExists(String, EclipseNode, boolean, int, boolean)} with {@code honorTolerate} = {@code true}.
+	 */
+	public static MemberExistsResult methodExists(String methodName, EclipseNode node, boolean caseSensitive, int params) {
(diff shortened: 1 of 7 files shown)


rzwitserloot:
Right, well, the problem is, this behaviour, while its a thing I can imagine one would like to do, isn't actually what Tolerate is documented as doing.

Also, this closes another exotic use-case which is:  You wrote an equals() method that is not the 'real' equals (parameter type isn't j.l.Object) and you'd like lombok to ignore it.

We'd have to document this. While the behaviour of this pull request is sensible enough, trying to document it is mostly going to read as confusing gobbledygook which is unfortunate.

If you're still interested in this feature, what's the use case? When would you like a lombok-generated equals while writing your own hashCode, and when would you like a lombok-generated hashCode while writing your own equals? What about canEqual?

Maaartinus:
**Concerning documentation**, for me adding "as much as it can makes sense" to the existing documentation would do.

Fineprint: 1. If generating a method would lead to an error and Lombok knows it, it never generates it. 2. Otherwise, it ignores existing method annotated with `@Tolerate`. 3. Otherwise, an existing equally-named method with the same number of arguments stops Lombok from generating a method. 4. Moreover, if there is any method named either `hashCode` or `equals`, regardless of return type, no methods will be generated, and a warning is emitted instead.

Note that I tried to summarize all suppressions due to existing methods. Rule 2 is the explanation of what `@Tolerate` is good for, rule 1 is just a sanity rule making it more usable. For me, both feel natural and are trivial to remember. Not knowing them can never lead to unexpected errors.

Actually, I'd probably extend the Tolerate documentation and replace all mentioning of suppressions due to an existing thing by a link.

**Concerning use cases**, I'd actually prefer my `hashCode` to be *always* deterministic. This may be a non-common need, but then it gets hard, as two important cases stay in the way, namely `enum`s and `Class`es. So while the requirement might be a bit exotic, the obstacles are rather abundant. I could imagine `@EQHC(moreDeterministic=true)` handling these two obstacles, but I'm afraid the demand is not high enough.

Also some manual caching of `hashCode` makes sense. A corresponding Lombok feature would be even better, but there are many options (precomputation or caching, support for other methods, possibly methods with arguments, etc.) and it doesn't exist yet. I'd personally be happy with a hash code precomputation (faster than caching and good enough for immutables), additionally used in `equals` as the very first thing to compare.

Someone could want to exclude some fields from `hashCode` for performance reasons, but not from `equals` (for correctness). This may be the case for large arrays or lists or dependent fields.

I can't imagine someone letting `hashCode` to be generated and writing `equals` manually (doing just the more tedious part) and `canEqual` is probably too rare to care about.

PR #51, opened 2014-06-24
support dynamic flag for @Log

sebek64:
- dynamic logging means non-static field initialized by getClass() call

This is my first commit to lombok, feel free to suggest improvements. The change has been partially tested (just @Log and @Slf4j).

### DIFF
--- a/src/core/lombok/extern/apachecommons/CommonsLog.java
+++ b/src/core/lombok/extern/apachecommons/CommonsLog.java
@@ -63,4 +63,9 @@ public @interface CommonsLog {
 	 * Sets the category of the constructed Logger. By default, it will use the type where the annotation is placed.
 	 */
 	String topic() default "";
+
+	/**
+	 * Whether to create non-static field with getClass() topic instead of default static. Useful for abstract classes.
+	 */
+	boolean dynamic() default false;
 }
--- a/src/core/lombok/extern/java/Log.java
+++ b/src/core/lombok/extern/java/Log.java
@@ -62,4 +62,9 @@ public @interface Log {
 	 * Sets the category of the constructed Logger. By default, it will use the type where the annotation is placed.
 	 */
 	String topic() default "";
+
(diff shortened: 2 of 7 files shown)


rzwitserloot:
Given that nobody's come up with a reasonable use case for changing the staticness of the logger within a single package, we're not going to add this feature; the lombok.config system is better suited for package and project-wide settings, and delivers the same functionality:

lombok.log.fieldIsStatic = false

Add that to a file called 'lombok.config' in the package or any parent directory of your source package and your loggers will be instance, not static.

mmoayyed:
It would be somewhat better if you exposed this options and let users decide. static loggers make sense for final classes. non-static for non-final and abstract classes specially if the log field is to be inherited from child classes.

rzwitserloot:
We _DO_ let users decide... via the lombok.config option.

The argument 'heeey, throw in a setting, what could it hurt' does not work for us; we think having, and I'm intentionally going overboard here, 50 separate parameter flags on an annotation is itself a bad thing. A forest of exotic, hard to explain and rarely used options will hide the few actually sensible ones. That alone is ample reason to not just chuck the kitchen sink at it.

We asked on the newsgroup for use cases that specifically make a case for having this configurable per logger (which pretty much implies: Some classes in a package want a static one, but some want an instance variant, otherwise, lombok.config can do it just fine), and we didn't hear anything that made enough sense to warrant a parameter in the annotation itself.

If you do have one, perhaps we should continue the discussion on the newsgroup:

https://groups.google.com/forum/#!topic/project-lombok/18jVJnMArvk

mmoayyed:
I totally and wholeheartedly agree with you on the issue of going overboard :) Sure. My only concern was that the ultimate issue would be to let the logger itself decide. You have global and package-level config. Most projects would probably need a combination of these for very specific components.

I'd be happy to continue with the conversation on the list. Thanks for your prompt response.

PR #52, opened 2014-08-21
Improve @Tolerate, so it really works with @EqualsAndHashCode.

Maaartinus:
It was possible to e.g. let Lombok ignore a preexisting `hashCode` using `@Tolerate`, but then both `equals` and `hashCode` were generated.

That's totally wrong. Generating the missing method only makes sense and there are two good use cases:
1. For perfect reproducibility, I need a deterministic `hashCode` (so that my HashSet gets always iterated in the same order), but there are some fields of type `Enum` or `Class`. When I compute the `hashCode` manually I can easily handle such fields specially. The generated `equals` has no such problem, so I want to use it.
2. With a value object, I precompute and store the `hashCode` for speed. Again, the generated `equals` fits perfectly. This is something Lombok could do itself, but for now there's just this workaround.

### DIFF
--- a/src/core/lombok/eclipse/handlers/EclipseHandlerUtil.java
+++ b/src/core/lombok/eclipse/handlers/EclipseHandlerUtil.java
@@ -1099,14 +1099,28 @@ public class EclipseHandlerUtil {
 		
 		return MemberExistsResult.NOT_EXISTS;
 	}
-	
+
+	/**
+	 * Wrapper for {@link #methodExists(String, EclipseNode, boolean, int, boolean)} with {@code caseSensitive} = {@code true} and {@code honorTolerate} = {@code false}.
+	 */
+	public static MemberExistsResult methodReallyExists(String methodName, EclipseNode node, int params) {
+		return methodExists(methodName, node, true, params, false);
+	}
+
 	/**
 	 * Wrapper for {@link #methodExists(String, EclipseNode, boolean, int)} with {@code caseSensitive} = {@code true}.
 	 */
 	public static MemberExistsResult methodExists(String methodName, EclipseNode node, int params) {
 		return methodExists(methodName, node, true, params);
 	}
-	
+
+	/**
+	 * Wrapper for {@link #methodExists(String, EclipseNode, boolean, int, boolean)} with {@code honorTolerate} = {@code true}.
+	 */
+	public static MemberExistsResult methodExists(String methodName, EclipseNode node, boolean caseSensitive, int params) {
+		return methodExists(methodName, node, caseSensitive, params, true);
+	}
+
 	/**
 	 * Checks if there is a method with the provided name. In case of multiple methods (overloading), only
 	 * the first method decides if EXISTS_BY_USER or EXISTS_BY_LOMBOK is returned.
@@ -1115,8 +1129,9 @@ public class EclipseHandlerUtil {
 	 * @param node Any node that represents the Type (TypeDeclaration) to look in, or any child node thereof.
 	 * @param caseSensitive If the search should be case sensitive.
 	 * @param params The number of parameters the method should have; varargs count as 0-*. Set to -1 to find any method with the appropriate name regardless of parameter count.
+	 * @param honorTolerate If true, then methods annotated with {@code Tolerate} will be ignored.
 	 */
-	public static MemberExistsResult methodExists(String methodName, EclipseNode node, boolean caseSensitive, int params) {
+	private static MemberExistsResult methodExists(String methodName, EclipseNode node, boolean caseSensitive, int params, boolean honorTolerate) {
 		while (node != null && !(node.get() instanceof TypeDeclaration)) {
 			node = node.up();
 		}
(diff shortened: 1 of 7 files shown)


rzwitserloot:
See the comments on the other pull request, but you've provided some use cases here.

I don't think @Tolerate covers the gist of what this does. We need some other way to mark: Yeah, yeah, I wrote my own hashCode. Please still just generate equals as planned.

But perhaps there are API-wise nicer ways of doing this. We can cater to the hashCode cache usecase by simply making an annotation for it. Lombok would generate a field and such automatically.

The usecase of needing a deterministic hashcode, mixed with fields where identity-equals always works out (that is, those instances have System.identityHashCode for a hashcode, but their equals() methods are nevertheless acceptable) is... exotic, to say the least. I'm not sure we should bother adding a feature to cater to that use case, which gets us back to: Hijacking tolerate to cater to it doesn't sound right.

Maaartinus:
**Concerning documentation**, for me adding "as much as it can makes sense" to the existing documentation would do.

Fineprint: 1. If generating a method would lead to an error and Lombok knows it, it never generates it. 2. Otherwise, it ignores existing method annotated with `@Tolerate`. 3. Otherwise, an existing equally-named method with the same number of arguments stops Lombok from generating a method. 4. Moreover, if there is any method named either `hashCode` or `equals`, regardless of return type, no methods will be generated, and a warning is emitted instead.

Note that I tried to summarize all suppressions due to existing methods. Rule 2 is the explanation of what `@Tolerate` is good for, rule 1 is just a sanity rule making it more usable. For me, both feel natural and are trivial to remember. Not knowing them can never lead to unexpected errors.

Actually, I'd probably extend the Tolerate documentation and replace all mentioning of suppressions due to an existing thing by a link.

**Concerning use cases**, I'd actually prefer my `hashCode` to be *always* deterministic. This may be a non-common need, but then it gets hard, as two important cases stay in the way, namely `enum`s and `Class`es. So while the requirement might be a bit exotic, the obstacles are rather abundant. I could imagine `@EQHC(moreDeterministic=true)` handling these two obstacles, but I'm afraid the demand is not high enough.

Also some manual caching of `hashCode` makes sense. A corresponding Lombok feature would be even better, but there are many options (precomputation or caching, support for other methods, possibly methods with arguments, etc.) and it doesn't exist yet. I'd personally be happy with a hash code precomputation (faster than caching and good enough for immutables), additionally used in `equals` as the very first thing to compare.

Someone could want to exclude some fields from `hashCode` for performance reasons, but not from `equals` (for correctness). This may be the case for large arrays or lists or dependent fields.

I can't imagine someone letting `hashCode` to be generated and writing `equals` manually (doing just the more tedious part) and `canEqual` is probably too rare to care about.

rzwitserloot:
It's backwards compatibility breaking if we give enums and classes special
treatment, but I think we can cross that bridge.

The bridge we can't cross is twofold:

* We can't tell that some type is an enum without resolution, and we don't
want to mess with resolution for @Data and friends, and...

* It feels 'weird' that a field of some enum type gets weird treatment
(ordinal instead of .hashCode for hashcode), but a field of a more generic
type (such as j.l.Object) that so happens to contain enums does not. Exact
same instance, but the type of the field decides behaviour? I guess this is
already true for arrays, so there's some precedent. But it still doesn't
really feel right.



 --Reinier Zwitserloot

Maaartinus:
You comment deals with my idea of `@EQHC(moreDeterministic=true)` only and I'm fine with what you wrote, provided that I can supply my own `hashCode` which will be `@Tolerate`d. Some other notes:

---

Concerning arrays, I guess I've never assigned an array to a field declared as `Object`. Similarly for `Class` and `enum`. So I don't really care, if the treatment of fields is based on their declared or their actual type. 

---

In cases like above, you can avoid resolution. Just generate both possibilities and branch on

    private static final boolean $FOO_IS_ENUM;

initialized via

     MyClass.getDeclaredField("foo").getType().isEnum();

The price is longer and ugly code, but no runtime overhead (possibly except for missing inlining due to  a longer bytecode). It's not nice, but it may be helpful in case there's no better solution.

rzwitserloot:
Given that this adds a LOT of magic voodoo to delomboked results,
introduces discrepancies (enums and classes get magic treatment, but there
are various other types that require magic treatment which don't get it, so
where DO we draw the line?), and we can't make good on the promise (I guess
'moreDeterministic', yeah, we can fulfill that promise, but it's a
ridiculous proposal. Nobody cares about 'more' deterministic, the only
relevant question is: Is it 100% deterministic, or not? And we can't
promise determinism, because we don't control equals and hashCode impls of
the constituent fields), for relatively little gain, I'm going to leave
this as something that does not currently clear the bar.


 --Reinier Zwitserloot

rzwitserloot:
How about this redesign of the of/exclude parameters?

https://github.com/rzwitserloot/lombok/wiki/FEATURE-IDEA:-Annotation-based-Include-Exclude-for-ToString,-EqualsAHC,-etc.

I think it would solve the central problem: You can write something like:

	@EqualsAndHashCode.Exclude CardRank rank; // Let's say this is an enum
	@EqualsAndHashCode.Include private int getRankOrdinal() {
		return rank.ordinal();
	}

and now you have a deterministic hashCode.

Caching hashCode is a separate issue, and would be handled with something like `@EqualsAndHashCode(cacheHashCode = true)`.

I'm going to close this pullreq, let's continue discussion in either a new issue or on the wiki page linked above.

PR #56, opened 2014-10-18
Minor, put installer window on correct location

ebraminio:
When installer window is opened, it would be much better if it would be in a correct location specified by OS instead being at most top left.


rzwitserloot:
Good point. Thanks!

PR #58, opened 2014-12-08
Adding annotation for creating Utility Classes easily

luanpotter:
Hello! I'd like to say first that I love your project. It is amazing what it opens up with Java, and serves mainly just to show how powerful and extensible this amazing language is.  
This is a feature I am missing very much in the pack for a while now. When we make Utility Classes, i.e., classes only with static methods, we should define a private default constructor, to override the public one that is created if none is presented, and avoid misuse of the class. The best practices would even ask for it to throw and exception when called, to avoid it being used internally or via reflection.  
This class level annotation creates that for you, checks if no other constructors are defined, and also makes all fields, methods and inner classes static, as well as marking the class final for documentation purposes (even though it could not be subclassed already due to lack of constructors). Simple as that.  
I am attaching a pull request, but it is more of a proof of concept. I did it for javac and it works fine. I noticed that I need to add a handler for eclipse as well, which is very odd (doesn't it use javac in the end as well?), but I had no luck in making it work on eclipse (there is a similar but different API, and I can't seem to get any grasp of it). I'll add here any news on that. There is also no documentation (website and whatnot). I did create a ConfigurationKey, because I inspired in the other annotations, yet I don't quite understand what that is for, apart from being some sort of custom switch or something.  
Anyway, what do you guys think of this idea? Of course the name of the annotation and other details can be easily changed, e.g., it can yield an error if you create a non-static member, instead of defaulting them.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -2,6 +2,7 @@ Lombok contributors in alphabetical order:
 
 Christian Sterzl <christian.sterzl@gmail.com>
 Jappe van der Hel <jappe.vanderhel@gmail.com>
+Luan Nico <luannico27@gmail.com>
 Maarten Mulders <mthmulders@users.noreply.github.com>
 Philipp Eichhorn <peichhor@web.de>
 Reinier Zwitserloot <reinier@zwitserloot.com>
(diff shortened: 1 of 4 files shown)


rzwitserloot:
Can you add your name to the /AUTHORS file?

luanpotter:
Done :) Any thoughts on the feature? I think it would come quite handy.

rzwitserloot:
Working on this now. The changes I am making:

* Some code cleanup.
* Add tests.
* Add documentation.
* If annotating an inner class, then the class is also marked static, unless that is not possible, in which case an error is flagged and nothing happens.
* Error messages updated.
* Added eclipse implementation.

commit will follow later.

rzwitserloot:
... and it's in. I've made an edge release if you want to play around with it :)

http://projectlombok.org/download-edge.html

luanpotter:
Good idea, I hadn't thought about non-static inner classes. Thanks for the eclipse impl too, I will surely take a look to see better how it works under the hood :)

PR #61, opened 2015-03-14
Rename file to fix BuilderSingularNoAutosingularize test failing on Linux OS.

mg6maciej:


rzwitserloot:
Thanks for spotting the problem. Fixing the capitalization works on linux but not on my mac at least, where the rename gets all confused. We fixed it (just pushed to master branch) by renaming the whole thing to BuilderSingularNoAuto (lopping off the 'singularize' part).

PR #64, opened 2015-03-29
Redesign of http://projectlombok.org/

lehvolk:
Hi. I want to contribute a more modern version of project site. I think that such cool java tool needs cool project site :-)

Currently I don't integrate with build system (actually because I can't run website.ant.xml).
If community approve new concept I will fix small defects (cross-browsers) and try to integrate with build system.


rzwitserloot:
Oh wow, this is awesome. It'll be the first thing on the agenda when Roel is back (which will be this very week), and we'll take care of integrating this stuff into the build system, it's the least we can do here.

Could you add your name to the /AUTHORS file? Also, I think the link to the edge release immediately goes to download it, instead of showing the changelog first, but that's something the build system normally takes care of. We'll try to fix it ourselves, should be a nice way to get some practical insights into the amazing work you've done here.

Thanks!

lehvolk:
Great. This week I will make some minor fixes to UI. I will work only with html and css, I will not move or rename any files, so this fixes will not affect integration with build system.

Thanks for lombok)

rzwitserloot:
Roel and I have looked at in some more detail. It looks great!

A few notes. First some minor design nits:

* The centered headers look weird on large screens. We think left-aligning everything, possibly with some fixed indent amount, is going to end up looking better on more screens.

* The snippets on the features pages overlap one another on a large enough (>1100px, so that they flow next to each other instead of  vertically) screen.

And then the major one:

* We're not fans of the angular/single-page model here. I have no idea if searchability is impacted, but there's not much point to it; the only elements that remain between pages are the top bar and the footer. The site doesn't currently have, and probably never will have, interactive elements (no chat or forums directly hosted within this website project).

If this is as far as your time on this project would go, our plan would be to take the style and remove all the single-paginess of it. It's rather annoying that we are forced to copy the header/footer across all the pages in the git tree (that's how the current site works), so while we're at it, probably some sort of build-time templating solution (some basic mustache or freemarker, run by the build script, should be more than plenty here, it's primarily header and footer for the bars and the default javascript/stylesheet, that's all really).


Would you be interested in doing this work? We can help or take care of at least the build part + running a templating engine as part of the build process if you like.

lehvolk:
"The centered headers look weird on large screens. We think left-aligning everything, possibly with some fixed indent amount, is going to end up looking better on more screens."

I think that center align looks better than left align (I test it under laptop und 22 inches hd display).  Maybe I missed something. If you want to change it - it's up to you)

"The snippets on the features pages overlap one another on a large enough (>1100px, so that they flow next to each other instead of vertically) screen."

This is deffect and I fix it this week.

"We're not fans of the angular/single-page model here. I have no idea if searchability is impacted, but there's not much point to it; the only elements that remain between pages are the top bar and the footer. The site doesn't currently have, and probably never will have, interactive elements (no chat or forums directly hosted within this website project)."

I frequently faced with angiular on my daily job so angular is the fastes way for me to implement the functionality. Yes there are < a href="http://www.yearofmoo.com/2012/11/angularjs-and-seo.html" some problems</a> with SEO. So I think best solution is to use freemarker. As I see there will be new ant task into website.ant.xml which generates static html from main template and views. So other task will stay the same but will points to generated  html. In this vision I can implement this.

lehvolk:
Hi, guys.

I switched to freemarker template engine and fix some bugs.

I think I get what you mean about weird text center position. It's actaully weird on pages where not enougth text under header (lilke installation pages). So I try to use tabs here and I think UI becomes more friendly.

I write ant task for html generation in both website.ant.xml and build.xml. Please, make futher build system integration.

carlonzo:
please resolve conflicts and merge!

Kjens93:
This can probably be closed since projectlombok.org has already been redesigned in the meantime.

lehvolk:
UI looks pretty close to my version.

rspilker:
Thanks @lehvolk for creating the design.

PR #66, opened 2015-05-15
Lazy annotation for making lazy methods

gvlasov:
Added a handler for Lazy annotation. 

This annotation can be applied to methods so their result is computed only once and stored in a generated private field.

The idea behind this annotation is to emphasise object's behavior over the details about how and where the computed data is stored.

Lombok code:

```
public final class Triangle implements Bounded {
  @Lazy 
  @Override
  public Rectangle boundingRectangle() {
    return // compute bounding rectangle for a triangle
  }
}
```

Equivalent Java code:

```
public final class Triangle implements Bounded {
  private Rectangle boundingRectangle;

  @Override
  public Rectangle boundingRectangle() {
    if (boundingRectangle == null) {
      boundingRectangle = // compute bounding rectangle
    }
    return boundingRectangle;
  } 
}
```

Currently the annotation has the following deficiencies:
- Doesn't work on methods that return primitives;
- Doesn't have the Eclipse compiler implementation.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -1,6 +1,7 @@
 Lombok contributors in alphabetical order:
 
 Christian Sterzl <christian.sterzl@gmail.com>
+Georgy Vlasov <wlasowegor@gmail.com>
 Jappe van der Hel <jappe.vanderhel@gmail.com>
 Luan Nico <luannico27@gmail.com>
 Maarten Mulders <mthmulders@users.noreply.github.com>
--- a/src/core/lombok/Lazy.java
+++ b/src/core/lombok/Lazy.java
@@ -1,3 +1,24 @@
+/*
+ * Copyright (C) 2013-2015 The Project Lombok Authors.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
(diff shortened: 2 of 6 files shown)


rspilker:
Hi Suseika, thanks for taking to time to create this feature.

I have some comments on several levels:
- In general, we advice contributors for new features to first discuss them on the [project lombok group](https://groups.google.com/forum/#!forum/project-lombok). that way you can already get some feedback.
- For no-args methods, Lombok already has `@Getter(lazy=true)`. This already does exactly what you've created.
- We've already discussed Memoization before. I suggest you try to find those discussions.
- We can only accept pull requests if you also add your name to the AUTHORS file. That way we're legally covered.
- The new source files (not the test files) should have the appropriate copyright notice.

Please let us know if you want to continue working on this pull request to implement Memoization.

Suseika:
Hi Roel,

I'd argue about `@Getter(lazy=true)` that it does *exactly* what I've made. With `@Getter(lazy=true)` you need to have the field in your source code apart from the method that implements the behavior. It is field-centric, not behavior-centric. With my annotation/handler you don't have to have any fields in your source code. And I think this approach (but not the implementation yet) is superior to what `@Getter(lazy=true)` suggests, because it is more in line with the goal of Project Lombok to eliminate boilerplate code.

But I'll study the discussions on memoization in the group.

Of course I would love to continue working on this feature.

Suseika:
I'd also argue about whether what I propse deserves to be called memoization. Memoization is when a pure function maintains a cache of results where keys of that cache are the combinations of arguments, right? My Lazy annotation is for methods *without* arguments, it is more like an immutable property. This is a much less complex and narrowly targeted feature than memoization.

yegor256:
+1 would love to see this feature in Lombok

Maaartinus:
On Fri, May 15, 2015 at 3:02 PM, Georgy Vlasov <notifications@github.com>
 wrote:

> I'd argue about @Getter(lazy=true) that it does *exactly* what I've made.
> With @Getter(lazy=true) you need to have the field in your source code
> apart from the method that implements the behavior. It is field-centric,
> not behavior-centric. With my annotation/handler you don't have to have any
> fields in your source code. And I think this approach (but not the
> implementation yet) is superior to what @Getter(lazy=true) suggests,
>
IMHO it depends. You can pass arguments to

@Getter(lazy=true)
double eval43 = eval(43);

@Getter(lazy=true)
double eval44 = eval(44);

and then it gets more compact then

@Memoize
double eval43() {
    return eval(43);
}

@Memoize
double eval44() {
    return eval(44);
}

Here, memoizing eval itself would be the way to go, but that's a different
and much more complicated story.

For parameterless methods, I guess that precomputing or memoizing hashCode
<https://groups.google.com/forum/#!searchin/project-lombok/hashcode/project-lombok/ImSQ7X7PeBo/CzvK72Bqk9kJ>
and
using it in equals could be pretty useful.

For methods accepting anything beyond boolean, enum and byte, you may want
caching rather then storing everything. For anything beyond char and short,
you have no choice. IMHO for caching, Guava's Cache should be used (there's
already a precedent with lombok.singular.useGuava). But I'm afraid that
alone finding a proper syntax could be pretty difficult.

rzwitserloot:
I'm going to deny the pull request, based on the notion that, while the feature would be nice, we'd take it in a different direction: Like `@Getter(lazy = true)`, it would be based on wrapping in an AtomicReference, which lets you:

* Memoize primitives (create the appropriate java.lang.X wrapper field, using `null` as marker)
* Use AtomicReference to represent the difference between memoizing 'null' and 'haven't memoized yet'.
* Optimally speaking such a feature needs to go a lot further and allow you to have arguments. The code builds up a cache, mapping input arguments to the memoized result.

The feature will probably be denied (it's quite complicated, `@Getter(lazy=true)` covers a lot of the use cases, and the memoizing of params is going to result in lots of tricky questions about how to configure cache eviction. Please discuss it first before working on a PR.

PR #67, opened 2015-05-15
Fix for issue #801 #810 and togglable camelcase style accessors for uppercase fields

riccardobl:
I've fixed issue #810 that looks like an actual bug to me and added two features that could be useful for some people.


riccardobl:
This pr got closed when i moved the branch, i have reopened it here: https://github.com/rzwitserloot/lombok/pull/70

PR #69, opened 2015-06-11
Add callSuper to @Data and @Value

lightoze:
With this patch, `@Data(callSuper = true)` becomes an equivalent for additional `@ToString(callSuper = true)` and `@EqualsAndHashCode(callSuper = true)`, because Lombok is about avoiding boilerplate, right?

### DIFF
--- a/src/core/lombok/Data.java
+++ b/src/core/lombok/Data.java
@@ -56,4 +56,11 @@ public @interface Data {
 	 * Default: No static constructor, instead the normal constructor is public.
 	 */
 	String staticConstructor() default "";
+
+	/**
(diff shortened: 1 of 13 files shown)


rzwitserloot:
We have a pretty cool idea for how to interop `@ToString` and super, which will probably end up requiring a separate annotation for it (`@SuperToString`). If that is the superior solution for calling super, it's really just `@EqAHC` that needs it / to match the style we might introduce `@SuperData` or some such. Let's hold off on this until we're more sure about the direction of the `@SuperToString` plan.

https://github.com/rzwitserloot/lombok/wiki/FEATURE-IDEA:-@SuperToString

PR #70, opened 2015-06-25
Fix for issue #801 #810 and togglable camelcase style accessors for uppercase fields 

riccardobl:
I've fixed issue #810 that looks like an actual bug to me and added two features that could be useful for some people.

### DIFF
--- a/src/core/lombok/ConfigurationKeys.java
+++ b/src/core/lombok/ConfigurationKeys.java
@@ -38,6 +38,14 @@ public class ConfigurationKeys {
 	// ----- global -----
 	
 	/**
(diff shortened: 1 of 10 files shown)


dyorgio:
2 years... I would like to create Getter with camel case on my static UPPER_CASE field.

rzwitserloot:
These are referencing old googlecode IDs. googlecode 810 is now issue #845, and googlecode 801 is issue #836

rzwitserloot:
We've explicitly killed treating capital-B Boolean as boolean because it does not conform to the spec. Restoring that seems to be what this pullreq does. We removed it for a reason unfortunately :/

PR #72, opened 2015-07-04
JUnit: Runner should not invoke fireTestFinished if test was ignored.

Fuud:
As done in `org.junit.runners.BlockJUnit4ClassRunner.runIgnored` if test is ignored, `Runner` should not invoke `fireTestStarted` and `fireTestFinished`. 
This issue leads to problems when running tests in IntelliJ Idea (see https://youtrack.jetbrains.com/issue/IDEA-142279)

Correct alghorithm:
1. Read LombokTestSources
2. If test should be ignored - invoke fireTestIgnored
3. Otherwice: invoke fireTestStarted, run test, invoke fireTestFinished

### DIFF
--- a/test/core/src/lombok/AbstractRunTests.java
+++ b/test/core/src/lombok/AbstractRunTests.java
@@ -49,23 +49,49 @@ import lombok.javac.CapturingDiagnosticListener.CompilerMessage;
 
 public abstract class AbstractRunTests {
 	private final File dumpActualFilesHere;
-	
-	public AbstractRunTests() {
+	private final boolean ignore;
+	private final LombokTestSource sourceDirectives;
+	private final LombokTestSource expected;
+	private final DirectoryRunner.TestParams params;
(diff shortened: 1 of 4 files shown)


rspilker:
Thanks for addressing this. I don't have IntelliJ so I cannot verify if it solves the problem.

For legal reasons we also need you to add you name in the right location to the AUTHORS file.

Also, it seems that you've also used a different code format that we do. For simple ifs and fors we don't use blocks, but place the content on the same line as the if/for. But I'm happy to correct that if you don't have the time to do that.

Also, you forgot to update the copyright years. But I can do that as well.

rspilker:
We are waiting for feedback from the requester,,,

rspilker:
No need to do anything anymore. I've creatd a different implementation: https://github.com/rzwitserloot/lombok/commit/8a504f8beec691b64d1c08f629520c2443821847

PR #73, opened 2015-07-09
add excludeToString property in Data annotation for: some subclass ne…

mo-open:
add excludeToString property in Data annotation for: some subclasses need not override the toString() of Super class


rzwitserloot:
This makes lombok more convoluted (I mean, if this feature is a good idea, I can think of about 15 others, and I don't think it's a solid plan if `@Data` has 20 parameters!). What you've made is another way to write `@RequiredArgsConstructor @EqualsAndHashCode @Getter @Setter` which is longer, but not a lot longer.

PR #875, opened 2015-07-15
Update changelog.markdown

Koichi-Kobayashi:
The link of issue is changed to GitHub from Google Code.


rzwitserloot:
The issue numbers also changed. We have the mapping file and we were going to update those to the matching github URLs today. You beat us to it :)

Should be done tonight.

PR #904, opened 2015-08-26
Don't call to "Collection.toArray()" with zero length array argument

vbauer:
It's a tiny performance optimization patch

### DIFF
--- a/src/core/lombok/bytecode/PoolConstantsApp.java
+++ b/src/core/lombok/bytecode/PoolConstantsApp.java
@@ -66,7 +66,7 @@ public class PoolConstantsApp extends LombokApp {
 		CmdReader<CmdArgs> reader = CmdReader.of(CmdArgs.class);
 		CmdArgs args;
 		try {
-			args = reader.make(raw.toArray(new String[0]));
+			args = reader.make(raw.toArray(new String[raw.size()]));
 			if (args.help) {
 				System.out.println(reader.generateCommandLineHelp("java -jar lombok.jar -printpool"));
 				return 0;
--- a/src/core/lombok/bytecode/PostCompilerApp.java
+++ b/src/core/lombok/bytecode/PostCompilerApp.java
@@ -76,7 +76,7 @@ public class PostCompilerApp extends LombokApp {
 		CmdReader<CmdArgs> reader = CmdReader.of(CmdArgs.class);
 		CmdArgs args;
 		try {
-			args = reader.make(raw.toArray(new String[0]));
+			args = reader.make(raw.toArray(new String[raw.size()]));
 			if (args.help) {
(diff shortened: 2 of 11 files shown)


src/core/lombok/bytecode/PoolConstantsApp.java
```diff
@@ -66,7 +66,7 @@
 		CmdReader<CmdArgs> reader = CmdReader.of(CmdArgs.class);
 		CmdArgs args;
 		try {
-			args = reader.make(raw.toArray(new String[0]));
+			args = reader.make(raw.toArray(new String[raw.size()]));
```
twillouer: Hi,

in fact, this is a performance drop 
https://gist.github.com/twillouer/ac13eb1dadc8a270f821
and http://cr.openjdk.java.net/~shade/scratch/ToArrayBench.java

vbauer: Hi,

Actually I'm also confusing like authors of this articles, so I've decided to check it myself and run this benchmarks on my machine:

```
Benchmark                             Mode  Cnt      Score       Error   Units
ToArrayBench.defensive_copy          thrpt   20  70353.047 ±  5513.834  ops/ms
ToArrayBench.simple_toArray          thrpt   20  92047.053 ± 10031.130  ops/ms
ToArrayBench.sized_array_fixed_size  thrpt   20  12923.122 ±   811.063  ops/ms
ToArrayBench.sized_array_from_list   thrpt   20  32272.962 ±  1582.431  ops/ms
ToArrayBench.zero_sized_array        thrpt   20  40815.625 ±  4311.759  ops/ms
```

32272.962 vs 40815.625 and `ToArrayBench.sized_array_from_list` is winner. I don't see a performance drop.

twillouer: It can depend of the JVM.
In the second article, you can see the response of the creator of JMH

twillouer: And take care, the best result is ToArrayBench.zero_sized_array with 40 000 executions by ms vs ToArrayBench.sized_array_from_list who is "only" 32 000 executions by ms.

vbauer: Hmm... Yes, you're completely right! I thought that score was measured in "ms", but it was "ops/ms".

I've just re-read this articles carefully and situation with `toArray` is really interesting. Thank you for information! This PR could be closed, sorry for taking your time.

rzwitserloot:
The lesson here is this:

* You simply can't predict the performance of stuff like this. Almost always, it just won't matter, though.
* Write code to be as clean and legible as possible.

That's why we use [0]. It's cleaner and easier to read.

Thanks for looking into it though; it's easy to just preach from the pulpit, some numbers always help.

PR #937, opened 2015-10-04
Added support for builder singular for Guavas ImmutableTable

### LINKED ISSUE #917: @Builder fails with Guava's ImmutableTable
The `@Builder` annotation doesn't know that Guava's `ImmutableTable` has a Builder (`ImmutableTable.Builder`). This means that the following class fails:

<pre><code>@Builder
class Example {
    private @Singular ImmutableTable table;
}</code></pre>

Since Lombok supports the builders of `ImmutableMap`, `ImmutableSet`, etc, it should probably support `ImmutableTable`'s builder.

spacanowski:
Pull request will resolve #917

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -12,6 +12,7 @@ Reinier Zwitserloot <reinier@zwitserloot.com>
 Robbert Jan Grootjans <grootjans@gmail.com>
 Roel Spilker <r.spilker@gmail.com>
 Sander Koning <askoning@gmail.com>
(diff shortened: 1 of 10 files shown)


spacanowski:
Changes done.
I left removing of leading and only in line whitespaces (easier to read when you have whitespace showing turned on in IDE).
Method ```getTypeArgumentsCount``` (fixed typo) is required for raw collections (without specified generics).

PR #978, opened 2015-11-26
@Singleton support for javac and Eclipse

mystarrocks:
The new experimental `@Singleton` annotation will help user build thread-safe Singleton classes that ensure only a single instance of them ever exist.

The implementation comes in two flavors:
* eager (creates and initializes a `private`, `static` and `final` field by invoking the `nullary` constructor)
* lazy (resorts to creating a nested class called the `holder` to hold the singleton instance)

The documentation also includes the usage samples and various configurable keys.

### DIFF
--- a/src/core/lombok/ConfigurationKeys.java
+++ b/src/core/lombok/ConfigurationKeys.java
@@ -446,6 +446,13 @@ public class ConfigurationKeys {
 	// ----- Singleton -----
 	
 	/**
+	 * lombok configuration: {@code lombok.singleton.flagUsage} = {@code WARNING} | {@code ERROR}.
+	 * 
+	 * If set, <em>any</em> usage of {@code @Singleton} results in a warning / error.
+	 */
+	public static final ConfigurationKey<FlagUsageType> SINGLETON_FLAG_USAGE = new ConfigurationKey<FlagUsageType>("lombok.singleton.flagUsage", "Emit a warning or error if @Singleton is used.") {};
+	
+	/**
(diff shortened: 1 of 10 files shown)


rspilker:
First of all thanks for the complete feature. It included configuration keys, documentation and both an Eclipse and Javac implementation.

We love to get contributions! However, there's a but...

There is a reason why lombok does not already have `@Singleton`. We don't think it's worth the maintenance burden. If you read [this StackOverflow](http://stackoverflow.com/questions/70689/what-is-an-efficient-way-to-implement-a-singleton-pattern-in-java/73763#73763) article, you can see that's it's very complicated to do it right.

In short, since Java 5, in most cases you should just use an enum:
    
    enum Elvis {
        INSTANCE
    }

This takes care of:
- Lazy initialization
- Serialization/deserialization
- Multithreading
- Reflection invoking a constructor

There is no boilerplate to remove anymore.

Also, singleton is on its way out. Factories, Services and Value objects are the replacement for it. A singleton with a static method is either a service or a factory that you're using.

The using code shouldn't care or need to know how to create one. It gets a reference, either via some sort of injection (passed in the constructor), or can ask it's context for one.

Again, thanks for all the work you put into this. I invite you to keep making contributions. To prevent disappointment, I suggest you post your feature requests first to our [mailing list](https://groups.google.com/forum/#!forum/project-lombok).

mystarrocks:
Fair enough! I just hate Singletons as much as any one does but they are still used extensively and often incorrectly, so I was thinking an annotation would help get them wrong (or at least would make it difficult to get wrong), but I completely agree with your comments. Appreciated for the encouragement!

PR #1019, opened 2016-02-09
Allow hashCode and equals generation to be skipped completely

cezarykluczynski:
Didn't find an option to switch off the generation of hashCode and equals completely.

I'm pretty new to Lombok, so please correct me if I'm missing anything here.


rspilker:
I don't see the use case for this. This might only be relevant in combination with `@Data` or `@Value`, but in that case you can also not use those, but the _exploded_ versions: `@RequiredAgsConstructor @Getter @ToString`

But again, why would you explicitly not want to generate an equals and hashCode method?

PR #1027, opened 2016-02-17
add option to emit only changed(=delomboked) files

esuomi:
While working with https://github.com/awhitford/lombok.maven I noticed that Lombok is currently missing this quite handy feature; control with a flag whether only modified(=delomboked) are outputted or not. This has implications especially with Maven `generate-sources` phase as having control over which files are written avoids issues with duplicate source file definitions.

Tangentially this will probably also help with https://github.com/awhitford/lombok.maven/issues/7

This PR is pretty much just a stub as I haven't worked with Ant since 2007 and the project's build definition is so huge that I can't really figure out what I should do to verify this actually works. I'll be happy to add tests etc. as long as you'll help me understand what is needed to get this actually merged :)

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -16,5 +16,6 @@ Sander Koning <askoning@gmail.com>
 Szymon Pacanowski <spacanowski@gmail.com>
 Taiki Sugawara <buzz.taiki@gmail.com>
 Yun Zhi Lin <yun@yunspace.com>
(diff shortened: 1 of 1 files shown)


rzwitserloot:
Resolving the conflicts was a lot more complicated than rewriting it. Feature added in de01d1eae16a61e1d3b4cafedd32e59f98396000

Issue #1023, opened 2016-02-12
About setter and the JavaBean standard


henry1992007:
Recently I encountered a problem when using @getter function.

I have a field named "pId" in my java bean but i cannot get the origin data after the bean is deserialized.

so i removed the @getter annotation and used the defaults getters that are generated by the intellij idea instead, which is named "setpId", problem solved.

which means the deserializer could not find the correct setter for the field "pId". i supposed lombok generated a setter named "setPId".

According to the java bean standard, if the second letter of the name is a capital, the first letter should remain what it is in getters and setters.

i'll be grateful if you help to take a look at the problem.
thanks again for your contribution.

askoning:
As described in #757 and a lot of related issues, this (unfortunately) won't be changed.

rzwitserloot:
Duplicate of #2693

Issue #757, opened 2015-07-14
Names of Getters and Setters generated by Lombok are buggy for attributes like "eMail" or "aValue" etc


lombokissues:
_Migrated from Google Code (issue 722)_

shijiebei2009:
as I think, the eclipse and IDEA do the right things, why are not you in conformity with them???

o-pikozh:
Maybe, we should create `@xGetter` and `@xSetter` annotations with new behavior?

rzwitserloot:
@o-pikozh No, clearly a world where there are 18 different `@FooGetter` annotations, each with slightly different behaviour, is far worse than what we have now. In 99.99% of all cases this doesn't matter, but having multiple getter annotations is something lots of people are going to see and then they need to have their day rudely interrupted to read all about exotic, irrelevant details.

It is what it is.

Toilal:
Please make it happen without breaking change, `@Getter(strict=true)`, `@Setter(strict=true)` could do the job.

rzwitserloot:
Duplicate of #2693

Ketty0725:
Simple things have to be complicated，it's so low!

Rawi01:
@Ketty0725 If you read the linked catch-all issue, you can see that it is not as simple as it looks at first glance. And this is fixed, by the way.

PR #1030, opened 2016-02-19
fix JavaBeans naming convention #1023

hussachai:
Fix for issue #1023 

Eclipse and Intellij generate getter and setter for pId as getpId and setpld respectively
while lombok generates getPld and setPId. 
When the first character is lowercase and the second character is uppercase in this case 'I' ('I' as an Ice scream), the first character of the field name should not be capitalized.

### DIFF
--- a/test/core/src/lombok/core/TestHandlerUtil.java
+++ b/test/core/src/lombok/core/TestHandlerUtil.java
@@ -21,11 +21,34 @@
  */
 package lombok.core;
 
-import org.junit.runner.RunWith;
-import org.junit.runners.Suite;
-import org.junit.runners.Suite.SuiteClasses;
+import static lombok.core.handlers.HandlerUtil.buildAccessorName;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.fail;
+import org.junit.Test;
 
-@RunWith(Suite.class)
-@SuiteClasses({TestSingulars.class, TestHandlerUtil.class})
-public class RunCoreTests {
+public class TestHandlerUtil {
+	
+	@Test
+	public void testBuildAccessorName() {
+		
+		/* normal cases */
+		assertEquals("setFieldName", buildAccessorName("set", "fieldName"));
+		assertEquals("setI", buildAccessorName("set", "i"));
+		assertEquals("setI", buildAccessorName("set", "I"));
+		assertEquals("setUrl", buildAccessorName("set", "Url"));
+		
+		/* when the second character is uppercase, leave the first character unchanged */
+		assertEquals("setURL", buildAccessorName("set", "URL"));
+		assertEquals("setaI", buildAccessorName("set", "aI"));
+		assertEquals("setaIr", buildAccessorName("set", "aIr"));
+		
+		/* handle empty string cases */
+		assertEquals("URL", buildAccessorName("", "URL"));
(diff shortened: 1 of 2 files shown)


imzj:
Yeah, I meet the same trouble with this, I consult the naming convention, and found many existing framework and IDEs like Spring Jackson Eclipse Intellij Netbeans use the naming convention of [JavaBeans API specification](http://download.oracle.com/otn-pub/jcp/7224-javabeans-1.01-fr-spec-oth-JSpec/beans.101.pdf) which in the Chapter 8.8  explain this.Maybe Lombok use the uniform way is better.

test/core/src/lombok/core/TestHandlerUtil.java
```diff
@@ -0,0 +1,54 @@
...
+		assertEquals("setI", buildAccessorName("set", "i"));
+		assertEquals("setI", buildAccessorName("set", "I"));
+		assertEquals("setUrl", buildAccessorName("set", "Url"));
+		
+		/* when the second character is uppercase, leave the first character unchanged */
+		assertEquals("setURL", buildAccessorName("set", "URL"));
+		assertEquals("setaI", buildAccessorName("set", "aI"));
+		assertEquals("setaIr", buildAccessorName("set", "aIr"));
+		
+		/* handle empty string cases */
```
armanSingh: It is a good practice to have a separate test for exceptions. Can you please take this try-catch section, and put it in a new test?

To verify if the correct exception was thrown or not, we state the expected Exception class in the @Test annotation e.g
If we expect and ABCException to be thrown, we can specify it as 

> @Test(expected = ABCException.class)

Further Reading: [Test Annotation](http://junit.sourceforge.net/javadoc/org/junit/Test.html )

test/core/src/lombok/core/TestHandlerUtil.java
```diff
@@ -0,0 +1,54 @@
...
+	@Test
+	public void testBuildAccessorName() {
+		
+		/* normal cases */
+		assertEquals("setFieldName", buildAccessorName("set", "fieldName"));
+		assertEquals("setI", buildAccessorName("set", "i"));
+		assertEquals("setI", buildAccessorName("set", "I"));
+		assertEquals("setUrl", buildAccessorName("set", "Url"));
+		
+		/* when the second character is uppercase, leave the first character unchanged */
```
armanSingh: This can also be shifted to a new method
`testBuildAccessorNameWithUpperCaseSecondCharacterInSuffix()`

rzwitserloot:
As issue #757 explains, due to messups in the spec when javabeans were 'born', 'our' way is also a valid way. Slightly less valid but nevertheless:

* 10% of the people want geteMail().
* 6% of the people want getEMail().
* 10% of the people don't care as long as we don't change it.
* 74% of the people don't care no matter what we do, because they don't have any fields that have a single lower-case letter followed by an upper case later.


The math says 'leave it as is' wins, 16 to 10 (The current situation is getEMail).

Numbers made up, but unless you have pretty good proof I'm way off base with em I see no reason to doubt them.

rzwitserloot:
NB: Assuming a lot more people complain about this, perhaps we can consider a lombok.config key so you can opt-in to 'geteMail' mode, but check with us first. As is, a pullreq with that change would not be accepted either.

PR #1037, opened 2016-02-29
documentation: use "provided" scope for SBT build tool

vn971:


rspilker:
Since this pull request, the whole website has been changed. I wanted to incorporate your feedback but can no longer find the correct location.

vn971:
@rspilker it seems that SBT installation instructions were just removed (no SBT entry in the list below):
https://github.com/rzwitserloot/lombok/tree/master/website/templates/setup

It would of course be good to add them back (or not delete in the first place), but since I no longer use lombok, I'm not as interested for making the change as I were.

Feel free to do as you wish.

PR #1053, opened 2016-03-17
Update instructions for using lombok with ecj and Maven

marthagenaars:
As requested by Roel.

These changes add instructions for how to get lombok to hook into an Eclipse compiler in `maven-compiler-plugin`. We are trying to emulate the Eclipse problem view by parsing the logs from a Maven build that uses the Eclipse compiler in a lombok-enabled project.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -7,6 +7,7 @@ Enrique da Costa Cambio <enrique.dacostacambio@gmail.com>
 Jappe van der Hel <jappe.vanderhel@gmail.com>
 Luan Nico <luannico27@gmail.com>
 Maarten Mulders <mthmulders@users.noreply.github.com>
+Mart Hagenaars <marthagenaars@gmail.com>
 Peter Grant <petercgrant@users.noreply.github.com>
 Philipp Eichhorn <peichhor@web.de>
 Reinier Zwitserloot <reinier@zwitserloot.com>
--- a/website/index.css
+++ b/website/index.css
@@ -40,6 +40,16 @@ h1 {
 	height: 170px;
 }
 
+.separator {
+	background-image: url(buttonbar.jpg);
+	width: 800px;
+	height: 8px;
+	line-height: 8px;
+	font-size: 14px;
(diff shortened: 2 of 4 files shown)


jakub-bochenski:
What you have here as 
```
+<pre>
+mvn clean <strong>dependency:copy@get-lombok</strong>
+set MAVEN_OPTS=<strong>-javaagent:target/lombok.jar=ECJ</strong> <em class="note">(or your OS's equivalent)</em>
+mvn install
+set MAVEN_OPTS= <em class="note">(or your OS's equivalent)</em>
+</pre>
```
could be just
```
+ MAVEN_OPTS=<strong>-javaagent:target/lombok.jar=ECJ</strong> mvn clean dependency:copy@get-lombok install
```

also binding the `dependency:copy` execution to the default lifecycle sounds like a better idea than requiring it's executed manually.

marthagenaars:
The `mvn install` is pulled out of the first command to make sure that the copied `lombok.jar` is actually in place before the JVM is launched. Your solution may work, but I dare not guarantee that the JVM will always launch with the lombok agent properly attached.

Temporarily setting the `MAVEN_OPTS` environment variable is an important step, so I opted to put it on a separate line. Additionally, this cannot be inlined as fluently on a Windows command line.

The reason for having a separate `dependency:copy` execution is to be more flexible, and not get in the way of other downloads/dependencies. This way, we also avoid having to copy the lombok jar each time for your "regular" maven build. You can, of course, adapt the example to suit your specific needs at all times.

jakub-bochenski:
> The mvn install is pulled out of the first command to make sure that the copied lombok.jar is actually in place before the JVM is launched. Your solution may work, but I dare not guarantee that the JVM will always launch with the lombok agent properly attached.

Yes I went to far with this -- it would fail the first time

> This way, we also avoid having to copy the lombok jar each time for your "regular" maven build.

Just FYI: the copy task is idempotent it won't  re-download the file if it's already there

rspilker:
Thanks

Issue #1065, opened 2016-03-31
Specify prefixes for @Builder setters


programmerr47:
Hello. Sorry, maybe I missed something, but for me it seems that currently there is no possibility to customize 'set' methods in @Builder annotation. Here is an example of feature:

``` Java
@Builder(builderClassName = "Builder", withPrefix="with")   
public class SomeClass {
    private int intField;
    private String stringField;
}
```

After delombok it should look something like that:

``` Java
public class SomeClass{
    private int intField;
    private int stringField;

    SomeClass(int intField, int stringField) {
        this.intField = intField;
        this.stringField = stringField;
    }

    public static Builder builder() {
        return new Builder();
    }

    public static class Builder {
        private int intField;
        private int stringField;

        Builder() {
        }

        public SomeClass.Builder withIntField(int intField) {
            this.intField = intField;
            return this;
        }

        public SomeClass.Builder withStringField(int stringField) {
            this.stringField = stringField;
            return this;
        }

        public SomeClass build() {
            return new CarBrand(intField, stringField);
        }
    }
}
```

Currently we have no such parameter in @Builder so after delombok we will have next code:

``` Java
public class SomeClass {
    private int intField;
    private String stringField;

    SomeClass(int intField, int stringField) {
        this.intField = intField;
        this.stringField = stringField;
    }

    public static Builder builder() {
        return new Builder();
    }

    public static class Builder {
        private int intField;
        private int stringField;

        Builder() {
        }

        public SomeClass.Builder intField(int intField) {
            this.intField = intField;
            return this;
        }

        public SomeClass.Builder stringField(int stringField) {
            this.stringField = stringField;
            return this;
        }

        public SomeClass build() {
            return new CarBrand(intField, stringField);
        }
    }
}
```

P.S.: I understand that maybe it is not super feature, but It's just strange for me, why we can customize builder name, even builder 'create' function name, but can not add prefixes.

Thank you!

shahpp:
Has this been released yet?

programmerr47:
As far as I know, no.

fdw:
Does anybody know when this will be released?

askoning:
See #1209: we prefer `with` to be the prefix reserved for `@Wither`.

PR #1066, opened 2016-03-31
#1065 added "withPrefix" argument for @Builder

programmerr47:
Hello. According to my question https://github.com/rzwitserloot/lombok/issues/1065 I've wrote base implementation of "withPrefix" argument.
If you have any suggestions or remarks, please, feel free to share with me.

Thank you!

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -7,6 +7,7 @@ Enrique da Costa Cambio <enrique.dacostacambio@gmail.com>
 Jappe van der Hel <jappe.vanderhel@gmail.com>
 Luan Nico <luannico27@gmail.com>
 Maarten Mulders <mthmulders@users.noreply.github.com>
+Michael Spitsin <spitsinmichael@gmail.com>
 Peter Grant <petercgrant@users.noreply.github.com>
 Philipp Eichhorn <peichhor@web.de>
 Reinier Zwitserloot <reinier@zwitserloot.com>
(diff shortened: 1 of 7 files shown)


congtrinh:
@rzwitserloot Any reasons why this shouldnt get merged in?

jensim:
I would LOVE to see this merged to master!.. 

Might I suggest a change of "`withPerfix`" to "`setterMethodPrefix`"? Doesn't that fit the rest of the properties on the annotation? ()
Some javadoc might be nice as well..

rspilker:
`set` is for mutables, `with` is for immutables that return a copy with one modification. Builder is mutable. See #1209 for more discussions.

We do appreciate the time you've put into this.

Unfortunately, that means we're going to not merge this in.

Next time, if you want to prevent disappointment, please contact us before creating a pull request to verify if:
- somebody is already working on it
- we think it is a good idea
- discuss alternatives

rspilker:
Uhm, I might have misread part of the implementation the first time. It is not about `with` vs. `set`, but <any given prefix> vs. none.

That doesn't change the outcome though. I'm not saying that the names for the setters are set in stone, but a per-instance setting sounds a bit weird to me, as well as anything else than nothing or `set`.

So the only thing I can imagine is a boolean config key to use a `set` prefix or none. But I'm not yet convinced that the added benefits are is worth the maintenance time.

PR #1069, opened 2016-04-02
add download using gradle to README.md

humazed:


rspilker:
If you want to add gradle instructions, the right location would be on [the website](https://projectlombok.org/mavenrepo/index.html). The [source](https://github.com/rzwitserloot/lombok/blob/master/website/mavenrepo/index.html) is also on GitHub.

Issue #557, opened 2015-07-14
Allow usage of annotations as meta-annotations


lombokissues:
_Migrated from Google Code (issue 522)_

olivergierke:
@Maaartinus - Any chance you elaborate on why you thing the change to `@Target` is a bad idea?

Maaartinus:
@olivergierke Sure: 1. I find it confusing, but that's subjective. 2. In the end, you'd have to add `ElementType.ANNOTATION` to all lombok annotations as why not allow to create `@CleanupUsingDispose` or `@NoGetterNoSetter` (the latter is something I could actually use a few times).

But it's not really bad, though I'd prefer the `Metaannotation` syntax. However, I have no idea if this is possible and those who do seem to be busy.

olivergierke:
I still don't think I get you. Why is that confusing? Creating composable annotations is exactly what `@Target(ANNOTATION)` exists for. You sample from above:

```java
@lombok.Metaannotation(@__({
  @AllArgsConstructor(onConstructor = @_(@Autowired)),
  @SomeOtherFancyStuff(1, 2, 3)
)})
@interface AutowiredConstructor {}
```

Would simple down to:

```java
@AllArgsConstructor(onConstructor = @_(@Autowired))
@SomeOtherFancyStuff(1, 2, 3)
@interface AutowiredConstructor {}
```

How is that more confusing or more complex?

theigl:
+1

thomasdarimont:
As a side note: Groovy AST Transformations provide this feature in the form of an `AnnotationCollector`: http://docs.groovy-lang.org/latest/html/gapi/groovy/transform/AnnotationCollector.html

barorion:
+1

lukashinsch:
+1

brushmate:
+1

matek2305:
+1

adko-pl:
+1

franDayz:
+1

madness-inc:
+1

jcesarperez:
+1

ArturGajowy:
+1

Khoulaiz:
+1

saljuama:
+1

olivergierke:
In the meantime, this got duplicated quite a few times (#602, #774).

[...]

PR #1075, opened 2016-04-06
Introduce @Property annotation that'll generate getters and setters.

dargiri:
Please merge "@Property" functionality. In big projects it starts to be very annoying to write @Getter @Setter annotation on each model class especially if you have a lot of them. It would be very nice if it was possible to use one simple annotation for this purpose.

When working with frameworks such as Hibernate that are forcing you to have getters and setters (writing @Getter and @Setter starts to be pretty boring after some time) it's not possible to use @Data annotation (which is pretty nice although and good for other purposes), because pretty often(always in our case) you need to have:
1. custom equals/hashCode method
2. custom toString method

### DIFF
--- a/buildScripts/website.ant.xml
+++ b/buildScripts/website.ant.xml
@@ -169,6 +169,9 @@ such as converting the changelog into HTML, and creating javadoc.
 		<antcall target="-integrateSnippet">
 			<param name="transformationName" value="experimental/onX" />
 		</antcall>
+		<antcall target="-integrateSnippet">
+			<param name="transformationName" value="experimental/Property" />
+		</antcall>
 	</target>
 	
 	<target name="-website-dist">
--- a/src/core/lombok/core/handlers/HandlerUtil.java
+++ b/src/core/lombok/core/handlers/HandlerUtil.java
@@ -47,6 +47,7 @@ import lombok.core.configuration.ConfigurationKey;
 import lombok.core.configuration.FlagUsageType;
 import lombok.experimental.Accessors;
 import lombok.experimental.FieldDefaults;
+import lombok.experimental.Property;
 import lombok.experimental.Wither;
 
 /**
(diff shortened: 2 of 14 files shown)


slomkowski:
@dargiri: I understand that `@Property` annotates the class and generates getters and setters for all methods.

How about splitting `@Property` into separate annotations for getters and setters, like `@Getters`, `@Setters` or similar? Using `@Getters` would be convenient for creating immutable classes. You would even be able to create immutable Hibernate entities:

```java
@Getters
@Access(AccessType.FIELD)
class MyEntity {
```

dargiri:
@slomkowski : Stuff you mentioned is already achievable by annotating a class with `@Getter` or `@Setter` if you want to have immutable classes. But it's not my case - there are several other reasons I don't want to to that. 

Also I don't want to fall into discussion in this particular thread why I don't want to have immutable classes and use them with Hibernate or any other tool (there are several reasons for that).

It's just feature that lombok lacks in my opinion.

Maaartinus:
> it's not possible to use @Data annotation (which is pretty nice although and good for other purposes), because pretty often(always in our case) you need to have: 1. custom equals/hashCode method, 2. custom toString method

You *can* use `@Data` and define your own methods.

---

> writing `@Getter` and `@Setter` starts to be pretty boring after some time

Let's compare:

    @Property
    @Getter @Setter

Saving 6 characters *per class* IMHO does not warrant a new annotation. What's worse, I can imagine `@Property` doing something else (e.g., generating some static inner class describing fields). Good names are scarce.

*The proper solution would be [meta-annotations][557].*

Btw., half of my classes get annotated with

    @RequiredArgsConstructor(onConstructor = @__(@Inject))

Meta-annotations would solve that too.

[557]: https://github.com/rzwitserloot/lombok/issues/557

dargiri:
@Maaartinus I know, that I can use `@Data` annotation and customize it with custom equals/hashCode, to string. But it's too much effort for a simple thing.

By the way in a lot of cases Java's POJO convention required to have always getters and setters, so it's a handy shortcut in my opinion. When you'll have 50-100 model classes you'll become pretty bored to write `@Getter` `@Setter` on each class (at least I am - that's the reason). It's not only about writing 6 characters more. It's also about importing two annotations instead of one (it's making me pretty mad when I need to type more than I want to). 

I agree that introducing new annotation was forced by not having possibility to extend existing annotations. Probably meta-annotations is the true way, but I think it still has the right for living ( I can agree, that naming is probably not the best one ).

rzwitserloot:
`@Data` does just what you want. Assuming you have some constructor (or you're okay with what `@Data` makes out of the box there), and you wrote your own toString and equals/hashCode, `@Data` just generates getters and setters.

We'll use `@Property` eventually for having fields with change listener support and such.

dargiri:
I think you missed the point. Data generates equals/hashCode/toString and there's no way to disable it. And it's very bad in some cases.

dargiri:
@rspilker I'm aware of this feature, but not always you need to have all fiedls exposed.

PR #1080, opened 2016-04-11
Added counts into @ToString

librucha:
I added "counts" param into ToString annotation, which substitute all arrays, collections and maps by their items count.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -5,6 +5,7 @@ DaveLaw <project.lombok@apconsult.de>
 Dawid Rusin <dawidrusin90@gmail.com>
 Enrique da Costa Cambio <enrique.dacostacambio@gmail.com>
 Jappe van der Hel <jappe.vanderhel@gmail.com>
+Libor Ondrusek <libor.ondrusek@gmail.com>
 Luan Nico <luannico27@gmail.com>
 Maarten Mulders <mthmulders@users.noreply.github.com>
 Peter Grant <petercgrant@users.noreply.github.com>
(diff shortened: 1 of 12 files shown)


rspilker:
I like that you added your name to the AUTHORS file, that you created a config key and added End-to-end tests. 

Bad news first: we are not going to merge this. We have some guidelines we consider when adding features:
__1)__ Is the feature solving a common problem?
__2)__ Does it help developers to create better code?
__3)__ Does the code become unpredictable/easy to misunderstand?
__4)__ Is it worth the added burden of maintaining?
__5)__ Does it conflict with other ideas or solutions?

In this case, especially __(1)__ and __(4)__ are a problem. 

The after-ecj seems to indicate that it didn't work though. It looks like you didn't get the eclipse side working at all.

I'm still considering if it is going to work, especially when someone creates their own collections. Using resolution is very brittle. Especially in Eclipse. Especially when using method local classes.

Consider adding the following to your test cases:
```java
class X {
    public void method() {
        class StringList extends ArrayList<String> {}
        @ToString(count=true)
        class MethLocal {
            StringList x = new StringList ();
        }
    }
}
```

Also, it only keys off the declared type of the field, not the actual value. I think that that's a good thing, but still need to think this through.

Did you consider a different output format?
```java
class Foo {
    int[] a = {};
    String b = "x";
}
```
In your code this would result in:
```java
"Foo(a.lenght=0, b=x)"
```
But an alternative implementations could be
```java
"Foo(a[0], b=x)"
"Foo(a(0), b=x)"
"Foo(#a=0, b=x)"
"Foo(|a|=0, b=x)"
```

One other problem in your implementation is that it is an all-or-nothing property. You cannot say: this field I want to see, the other fields I want counted.

I hope that you continue to contribute to this project. Next time, please send a mail to the mailing list first, so we can discuss the guidelines, possible solutions and pitfalls.

PR #1092, opened 2016-04-20
Update Javadoc URL on Value

### LINKED ISSUE #1091: Javadoc on @Value points to experimental
In version 1.16.8.
The javadoc on lombok.Value points to https://projectlombok.org/features/experimental/Value.html
It should point to https://projectlombok.org/features/Value.html

While you are at it then the example on https://projectlombok.org/features/Value.html has this import
import lombok.experimental.Value;

I think import lombok.Value; is better.

DavidARussell:
Value is no longer experimental, so (as the person who raised the issue
suggested) the Javadoc should no longer point to the 'experimental'
documentation. Fixes #1091

### DIFF
--- a/src/core/lombok/Value.java
+++ b/src/core/lombok/Value.java
@@ -31,7 +31,7 @@ import java.lang.annotation.Target;
  * <p>
  * Equivalent to {@code @Getter @FieldDefaults(makeFinal=true, level=AccessLevel.PRIVATE) @AllArgsConstructor @ToString @EqualsAndHashCode}.
  * <p>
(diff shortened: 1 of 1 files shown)


DavidARussell:
Putting myself in 'AUTHORS' seems excessive for something so trivial, I wasn't sure whether it was a project rule to always edit it or not. I'm happy to update the PR without the AUTHORS edit (and save that for if I make a future PR with actual work in it!)

rspilker:
No need to add your name to the AUTHORS file. Thanks for the fix.

PR #1093, opened 2016-04-21
Add access level on logs annotations

antechrestos:
This change brings the possibility to add an access level on a log annotation (default: private for backward compatibility)
@rzwitserloot : I implemented it directly not knowing how to import the project correctly in my IntellIJ. I am opened to any modification/enhancement that need to be done, but I may need some help to import the project in the IDE :smile: . 
By the way, thank you for you project, it saved me and a lot of people around me a lot of time. :+1:


arnaudruffin:
👍

arnaudruffin:
too bad...

javatlacati:
that was promising :disappointed:

antechrestos:
@javatlacati  Yeah but it breaks some rules in the project. However you can use properties at a package level to obtain the same functionality

PR #1124, opened 2016-06-07
support custom annotation again

liudongmiao:
Since 1.16.X, lombok introduces `ShadowClassLoader`, which breaks custom annotation.
After digging into the codes, it's related to `inOwnBase`.
`InOwnBase` rejects any class not in `lombok.jar`, so it doesn't support custom annotation.

IMO, it's safe to use `!isSystemResource` instead of `inOwnBase`.

### DIFF
--- a/src/launch/lombok/launch/ShadowClassLoader.java
+++ b/src/launch/lombok/launch/ShadowClassLoader.java
@@ -94,6 +94,7 @@ class ShadowClassLoader extends ClassLoader {
 	private final String sclSuffix;
 	private final List<String> parentExclusion = new ArrayList<String>();
 	private final List<String> highlanders = new ArrayList<String>();
+	private final Map<String, String> extensionBases = new HashMap<String, String>();
 	
 	/**
 	 * @param source The 'parent' classloader.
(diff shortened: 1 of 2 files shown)


rzwitserloot:
I'm pretty sure isSystemResource is not actually safe here; the whole point of not trying to load 'foreign' classes into our loader is to avoid tainting such classes with the deps of lombok itself.

I've come up with an alternate strategy; check out branch sclExpansionUpdate; the trick here is:

Your classpath root with the lombok extension must contain META-INF/ShadowClassLoader which is a text file that contains the string 'lombok'. Do that, and the shadowclassloader will treat your classpath root the same as its own root (that is, will load anything in there into the lombok space so that lombok can see it, and it can see lombok).

Let me know if this works for you so it can be merged into master.

liudongmiao:
Yes, it works.

For the lombok extension, there are `META-INF/services/lombok.javac.JavacAnnotationHandler` or `META-INF/services/lombok.eclipse.EclipseAnnotationHandler`. IMO, it should be sufficient for the lombok extension without `META-INF/ShadowClassLoader`.

However, I don't know `tainting`. Is it the same with `Sealed`? http://docs.oracle.com/javase/tutorial/deployment/jar/sealman.html

liudongmiao:
I have another idea.

When `SpiLoadUtil` find services, add the resource to `ShadowClassLoader`, however, it's ugly.

menaren:
Hi, when will the sclExpansionUpdate branch modifications be added to the master branch ? 
I wrote a bunch custom annotations for the 1.14.8 which are not working anymore.

PR #1138, opened 2016-06-23
always return ShadowClassLoader.

liudongmiao:
It fixes `lombok` `1.16` doesn't works in `maven-compiler-plugin` `2.3.2` or old.
http://stackoverflow.com/questions/34358689
    
Furthermore, `maven-compiler-plugin` `2.4` requires `plexus-compiler-javac` `1.8.6`, which drops `org.codehaus.plexus.compiler.javac.IsolatedClassLoader`.
    
I love hack, but hate such hard code to fix possible problem.


rspilker:
This fix lloks good. Can you please add your name to the AUTHORS file, so we are legally covered?

liudongmiao:
Just added.

PR #1166, opened 2016-08-01
allow annotation processor to run via an other apt

jBouyoud:
Hi,

This change indicates to the java compiler that current annotation processor has altering files and he should take care about that. 
This is very usefull when you want to use lombok with another apt (like http://mapstruct.org/)

### DIFF
--- a/src/core/lombok/core/AnnotationProcessor.java
+++ b/src/core/lombok/core/AnnotationProcessor.java
@@ -163,7 +163,7 @@ public class AnnotationProcessor extends AbstractProcessor {
 		
 		for (ProcessorDescriptor proc : active) proc.process(annotations, roundEnv);
 		
-		return false;
+		return true;
 	}
 	
 	/**


shibenli:
Hi, is this usefull for Android JACK compiler? @jBouyoud

rspilker:
In cec8705c3b9fa6488d53bec5bb4eab27a6decf85 I've added the ClaimingProcessor that should take care of this. It is in lombok 1.16.10. Isn't that enough?

jBouyoud:
No sorry, it doesn't work.
I suggest to change lombok.core.AnnotationProcessor#process. This method always returns false, returning true allow to use lombok with another annotaitons processors (like mapstrcut).

shibenli:
@jBouyoud .Yes,i have got the error by using the last vesion one

rspilker:
The main processor triggers on all annotations, even non-lombok ones (there's reasons for that). If we return true, we'd be claiming all annotations, of all stripes. That sounds like it'll cause more problems than it'll solve.

According to the [API documentation](https://docs.oracle.com/javase/8/docs/api/javax/annotation/processing/Processor.html#process-java.util.Set-javax.annotation.processing.RoundEnvironment-) we should only return true if we've handled the annotation.

> Note that if a processor supports "*" and returns true, all annotations are claimed. Therefore, a universal processor being used to, for example, implement additional validity checks should return false so as to not prevent other such checkers from being able to run.

jBouyoud:
I think that's the point.
Why the main lombok annotations processor is trigger onto all (*)
annotations ?
Naively, i would have restricted the main processor should be triggered
onto 'lombok.*' annotations and then claim processed annotations.

Le mar. 22 nov. 2016 à 01:05, Roel Spilker <notifications@github.com> a
écrit :

rspilker:
We need to process * because we not only use annotations. For instance `val`, and more recently `var` can be in any file.

nlebas:
I would be willing to abandon the "val" feature in order to get just @Data-related stuff and interoperability with mapstruct/vertx/checker framework /other major annotation processors out there

rspilker:
We are close to making a fix for the interoparability with at least mapstruct.

rspilker:
We expect to be able to work with mapstruct after today. The techniques used could also be applied by other annotation processors.

PR #1196, opened 2016-09-22
added supporing for @val variables inside lambdas.

bulgakovalexander:


rspilker:
Thanks for the code!

Can you add your name and email address to the AUTHORS file so we are legally covered?

bulgakovalexander:
Hi, @rspilker.
> Can you add your name and email address to the AUTHORS file so we are legally covered?

Done.

rspilker:
I've merged in your modifications after adding a cache for the getBody method.

Thanks for adding the tests as well.

rikoe:
Hi. When is this planned for release, would be really nice to have? The latest edge release was in July, and edge releases aren't in the maven repo anyway, as far as I can tell.

PR #1217, opened 2016-10-24
Feature/var

bulgakovalexander:
Added 'var' annotation with the same functionality as the 'val' one but without mandatory 'final' behaviour.
It is possible to use in old for loops.
for(var i = 0, i < 100; ++i) ;


rspilker:
Thanks for the high quality contribution again.

Reinier and I have not  added `var` earlier for several reasons:

1. It is not as useful as 'val'. Most examples in the wild could have been implemented using val.
2. Since the variable will be reassigned (otherwise 'val' would have been used) the type of the variable should ideally be based on the right hand side of **all** assignments. This analysis is very complicated, or possibly (nearly) impossible: `var x = ""; x += x.length();`

Your solution only takes the initial assignment into account. This does not work for variables that are initially set to `null`.

We fear that the naive implementation will lead to confusion. There are also a few other concerns we have about `var`, we can elaborate if you're still interested in pursuing this feature.

Our initial response was to reject this pull request. But since you already contributed another fine patch, and the quality of this one is also high, we will give you a choice in the matter.

Reinier and I don't want to take on the responsibility for this feature. But if you want to we will allow you to under the following conditions:

- You move `var` to the experimental package.
- You add an opt-in toggle for `lombok.config`. We think we can add a value `ALLOW` to the existing flagUsage enum, and if it is missing for highly experimental features the error message should point to the website that explains that the feature is really experimental and that users can enable it by adding `lombok.var.flagUsage = ALLOW` to the appropriate `lombok.config` file.
- You handle feedback and do the maintenance.
- You maintain the documentation page. This page should inform the users it is really experimental (as in, you have to explicitly turn it on via `lombok.config`).

There is a big chance `var` will never make it to the main 'lombok' package. 

Is the above something you would like to commit to?

We look forward to more of your contributions. Next time, please contact us before you start working on something. We'd like to discuss alternatives, pitfalls, maintenance costs, etc.

Maaartinus:
> Since the variable will be reassigned (otherwise 'val' would have been used) the type of the variable should ideally be based on the right hand side of all assignments.

I hope, I'm not repeating myself, but no, this would be *in no case ideal*:

```
var x = new ArrayList();
overloadedM(x); // 2
x = new LinkedList(); // 3
overloadedM(x);
x = new HashMap(); // 5
```

- Java semantics allows reading a method top-down, except for here. What gets called at line (2)?
- At line (3), we know it's `overloadedM(List)`. But we're wrong.
- At line (5), we finally realize we declared `Cloneable & Serializable x`. Oh, did we?

A less smart analysis can be better, when it's easier to follow (Example: we all know that `if (x>0 && x<0) {}` contains unreachable code, but--for good reason--javac doesn't care).

> This does not work for variables that are initially set to null.

Sure, but the problem `val` and `var` solve, is not being explicit about the types, but *being explicit twice*. Java forces us to repeat the lengthy RHS type on the LHS, which is a completely useless boilerplate. Specifying the type once is sometimes boilerplate, too, but not that bad.

---
- I guess, this simple `var` covers most cases left over by `val`.
- When it works, you just look at the declaration line and you see the type.
- When it doesn't work, it just doesn't compile, no evil surprises.

[1]: https://groups.google.com/forum/#!topic/project-lombok/b1D_NxC0__U

bulgakovalexander:
Hi @rspilker. Thanks for so good feedback. I made changes that you requested and I ready to maintain this feature. I am not sure that feature should get out from the experimental status because I hope the JEP 286 will be approved in next java versions. But I suggest that the 'var' should cover some rare cases where the 'val' not acceptable.

rspilker:
Looks good! I do think that `var x = null;` should result in a warning though. Do you agree?

Also, your IDE settings regarding the imports and whitespace in javadoc differ from ours, resulting in too many modified lines. If it's possible, maybe you can fix those? When I merge, I can squash the commits so the intermediate steps are no longer visible.

rspilker:
`var x = null;` might even be an error, as in the example on http://openjdk.java.net/jeps/286#Risks-and-Assumptions

PR #1233, opened 2016-11-12
Feature/var

bulgakovalexander:
As a continuation of the [1217](https://github.com/rzwitserloot/lombok/pull/1217)
I accidentally deleted my branch which led to closing previous pull request.

Current changes:
- the changes of imports and white spaces in Java docs have been reverted.
- `var x = null;` will throw the error `variable initializer is 'null'`.
I thought about to make the same checking for `val` variables but have found usage of the null variables in tests, as example in the `TrickyTypeResolution.java`. So that checking can break compatibility.


rspilker:
It is in!!!!!

I did some small tweaks, but basically it's your work.

PR #1253, opened 2016-12-04
guard against possible npe

mebigfatguy:
ListBuffer<JCExpression> bounds = types.getBounds((TypeVar) param).isEmpty() ? null : new     ListBuffer<JCExpression>();
			for (Type type : types.getBounds((TypeVar) param)) {
				bounds.append(JavacResolution.typeToJCTree(type, annotation.getAst(), true));
			}
			

--> if bounds is null, npe here
    typeParams.append(maker.TypeParameter(name, bounds.toList()));


rspilker:
Thanks a lot. I love fixes that consist of only removing characters...

PR #1259, opened 2016-12-13
add missing printf anchor

mebigfatguy:


rspilker:
Thanks for the fix.

rspilker:
Uhm, turns out it was the wrong fix. The right fix was to remove the parameter, since it is already listed below.

Issue #1102, opened 2016-05-05
Is there a way to not have annotations lower test coverage?


dabraham02124:
I'm sure that lombok itself is extensively tested.  But I'm running coverage tools on my application which uses lombok extensively, and a lot of my classes are reporting coverage of 70% or so, even though every line of the code I wrote is covered (per Emma).

For instance, I frequently use @Value mostly for the getters, constructor, and private final.  I'm glad that equals, toString, & hashCode are there in case I need them, but I'm frequently not using them, much less testing them.

The simple answer is lots of boilerplate where I have a thing that runs all of the lombok functionality for a given class (I'd trust the tests on lombok itself to do the actual testing).  But adding boilerplate due to lombok seems wrong.

My first thought is an annotation to be placed on a test class.  Something like:
`@LombokTest(class=Foo.class) public class FooTest {
/* some methods that test the code I wrote */
}`

The annotation would investigate class Foo, see what lombok has added, then run those methods, thus letting my coverage report more or less reflect the coverage of my code).  I'm sure that there are better ways, but this is what I came up with.

Again thank you for lombok.  It makes my life distinctly easier.

Maaartinus:
Duplicate of [issue 1014](https://github.com/rzwitserloot/lombok/issues/1014).

I'm afraid, there's no solution for tools working on class files (because of the retention of `@javax.annotation.Generated` and because of `@lombok.Generated` not used yet).

dabraham02124:
Interesting.  So I'm late to an ongoing project.  Googling around I only found [this](https://projectlombok.org/api/lombok/Generated.html), and [this](https://groups.google.com/forum/#!topic/project-lombok/Csvi4m42yJs) (oddly enough googling on "lombok generated annotation" turns up a lot of false positives.  Is there a discussion of when this is happening, and what we'll have to do to work with it?

Again, thank you very much.

Issue #1014, opened 2016-02-04
Improve combability with cobertura


halmafest:
I use lombok to generate a lot of code and I really like it!
To measure the code coverage we use [cobertura](http://cobertura.github.io/cobertura/)

We like to use the cobertura option [ignoreAnnotations](https://github.com/cobertura/cobertura/wiki/Coverage-Annotations). Basically the feature allows to disable code coverage checks of methods annotated with a defined annotation. Unfortunately the annotation added by lombok  _javax.annotation.Generated_ does not work. 

Reasons:
- as I understand cobertura works on class files
- _javax.annotation.Generated_ has a @Retention(value=SOURCE) definition ([javadoc](https://docs.oracle.com/javase/6/docs/api/javax/annotation/Generated.html))
- means the annotation gets removed from the java compiler and is not visible during the code coverage check
- => each annotation having a @Retention(value=SOURCE) can not be used to exclude code.

Possible solutions:
- allow a user to define which annotation(s) are added to generated source
- add the already defined annotation _[lombok.@Generated](https://projectlombok.org/api/lombok/Generated.html)_ to generated source

Maaartinus:
Another solution idea: At the end, for every `@javax.annotation.Generated` annotation, add `@lombok.Generated`. This could be pretty simple and could help other code generation tools.

halmafest:
As a workaround I did a few modifications to the lombok code to write the lombok.Generated annotation in addition.

Changes: 
- lombok.Generated is missing the constructor flag in the @Target
- when javax.annotation.Generated is written I added to write in addition lombok.Generated to the code

Result:
- + Code coverage reports looks nice now. Works together with cobertura
- + only few changes required
- - issues during delombok tests, annotation isn't known except you add lombok as dependency
- - forked library, need to rebuild for every new version of lombok I want to use

rspilker:
It is possible to add an extra option to FormatPreferences either generate or not generate source code using the lombok.Generated annotation.

I think by default the lombok.Generated should not be generated in delomboked code.

Regarding the forked library: we do accept pull requests...

dabraham02124:
Yes please, @halmafest.  As the person who reported the dupe issue #1102, I'd love to see this pulled in to mainline lombok.  Thank you very much.

halmafest:
I will see if I can find some time the next weeks o make a clean pull request. 
@rspilker,  @all What is the preferred solution regarding FormatPreferences. I see two different options how to implement it:
[...]

PR #1271, opened 2016-12-24
Add config key lombok.generatedAnnotationName and handling

Gaibhne:
As per these open issues and discussions:

https://github.com/rzwitserloot/lombok/issues/1014
https://github.com/rzwitserloot/lombok/issues/1102
https://github.com/rzwitserloot/lombok/issues/624
https://groups.google.com/forum/#!topic/project-lombok/Vhj46vf2pWg
https://groups.google.com/forum/#!topic/jacoco/r_iWrlAYMp8
http://stackoverflow.com/questions/10276666/lombok-annotations-vs-code-coverage-in-cobertura-or-similar-tool

There is a need by people running code coverage tools to filter out methods generated by Lombok. Writing tests for generated methods is by most considered pointless and goes against the idea of Lombok of removing the clutter of these boiler plate methods by adding boiler plate tests.

Because the `@javax.annotation.Generated` annotation that is already added by Lombok to generated methods has a retention level of 'source', and almost all code coverage tools analyze class files, it is not possible to use the default annotation as a way to signal that those methods should be ignored.

This small PR adds a config key to supply your own annotation to be generated instead, which can then have a retention level of 'class' and can be used with code coverage tools. Example of such an annotation:

```
package com.someproject.util;

import java.lang.annotation.Retention;
import java.lang.annotation.RetentionPolicy;

@Retention(RetentionPolicy.CLASS)
public @interface LombokIgnore {}
```

I have successfully tested this PR against Cobertura with the following snipped in my pom.xml:

```
<plugin>
	<groupId>org.codehaus.mojo</groupId>
	<artifactId>cobertura-maven-plugin</artifactId>
	<version>2.7</version>
	<configuration>
		<instrumentation>
			<ignoreMethodAnnotations>
				<ignoreMethodAnnotation>com.someproject.util.LombokIgnore</ignoreMethodAnnotation>
			</ignoreMethodAnnotations>
		</instrumentation>
	</configuration>
	<executions>
		<execution>
			<goals>
				<goal>clean</goal>
			</goals>
		</execution>
	</executions>
</plugin>
```

### DIFF
--- a/src/core/lombok/ConfigurationKeys.java
+++ b/src/core/lombok/ConfigurationKeys.java
@@ -46,6 +46,13 @@ public class ConfigurationKeys {
 	public static final ConfigurationKey<Boolean> ADD_GENERATED_ANNOTATIONS = new ConfigurationKey<Boolean>("lombok.addGeneratedAnnotation", "Generate @javax.annotation.Generated on all generated code (default: true).") {};
 	
 	/**
+	 * lombok configuration: {@code lombok.generatedAnnotationName} = {@code com.fully.qualified.AnnotationName}.
+	 * 
+	 * If set, lombok generates {@code @com.fully.qualified.AnnotationName} on all fields, methods, and types that are generated. If the specified annotation is to be used by code coverage tools, {@code @Retention(RetentionPolicy.CLASS)} is usually required.
+	 */
+	public static final ConfigurationKey<String> GENERATED_ANNOTATION_NAME = new ConfigurationKey<String>("lombok.generatedAnnotationName", "Name of the annotation to generate on all generated code (default: javax.annotation.Generated).") {};
+	
+	/**
 	 * lombok configuration: {@code lombok.extern.findbugs.addSuppressFBWarnings} = {@code true} | {@code false}.
 	 * 
 	 * If {@code true}, lombok generates {@code edu.umd.cs.findbugs.annotations.SuppressFBWarnings} on all fields, methods, and types that are generated.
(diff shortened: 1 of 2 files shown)


rspilker:
Since quite some time (v1.16.2) we ship lombok with the `lombok.Generated` annotation in the jar. It has retention CLASS and targets CONSTRUCTOR, METHOD, FIELD and TYPE.

We did already plan to add `lombok.Generated` to all generated code. Would that also solve your problem?

Good thing about our own annotation
- Has the right retention
- Supports the right targets
- Does not have parameters that need to be generated as well
- Is already present for all users
- No need to configure, unless you want to suppress it

We eventually would like to use the configuration system to include type names as a separate type. That's also why I am hesitant to now abuse the String to put a class name in.

Gaibhne:
Yes, adding lombok.Generated to generated code would solve the problem as well, but it seems like you have been hesitant to make that change, as like you've said the annotation has already been shipped for a while. I don't know if there are use cases of code coverage tools being able to ignore only their own annotation without being able to specify them. I know both Cobertura and the still in-dev filtering mode of Jacoco allow you to specify which annotation(s) you would like to ignore, so specifying lombok.Generated should work fine there.

Gaibhne:
Yes, adding lombok.Generated to generated code would solve the problem as well, but it seems like you have been hesitant to make that change, as like you've said the annotation has already been shipped for a while. I don't know if there are use cases of code coverage tools being able to ignore only their own annotation without being able to specify them. I know both Cobertura and the still in-dev filtering mode of Jacoco allow you to specify which annotation(s) you would like to ignore, so specifying lombok.Generated should work fine there.

Issue #1018, opened 2016-02-09
NonNull generating incorrect null check in constructor 


linuskamb:
I'm quite new to Lombok, so forgive me if I'm doing something wrong.

It seems the emitted code is checking the value of the object field before setting the field with the value of the parameter.   

Notice the difference between the constructors' parameter names versus the setter parameter name.  In the setter, the parameter is named _id, shadowing the field, whereas in the constructor, it is named id without the underscore, yet the null check in both cases checks _id.

my lombok config includes: 
lombok.accessors.prefix+=_

I have a test case:

@Data
@RequiredArgsConstructor
@AllArgsConstructor
public class TestBok {

```
@NonNull
private Integer _id;
private String _foo;    
public static void main(String[] args) {
    try {
        Integer id = new Integer(42);
        String foo = "bar";
        TestBok b = new TestBok(id, foo);
        System.out.println(b);
    } catch (Exception e) {
        e.printStackTrace();
    }
}
```

}

Output is:
java.lang.NullPointerException: _id
    at gov.noaa.pmel.tsunami.model.TestBok.<init>(TestBok.java:10)
    at gov.noaa.pmel.tsunami.model.TestBok.main(TestBok.java:23)

Delomboked (partial) code:

import lombok.NonNull;

public class TestBok {
    @NonNull
    private Integer _id;
    private String _foo;

```
public static void main(String[] args) {
    try {
        Integer id = new Integer(42);
        String foo = "bar";
        TestBok b = new TestBok(id, foo);
        System.out.println(b);
    } catch (Exception e) {
        e.printStackTrace();
    }
}
```

[...]
    @java.beans.ConstructorProperties({"id"})
    @java.lang.SuppressWarnings("all")
    @javax.annotation.Generated("lombok")
    public TestBok(@NonNull final Integer id) {
        if (_id == null) {
            throw new java.lang.NullPointerException("_id");
        }
        this._id = id;
    }

```
@java.beans.ConstructorProperties({"id", "foo"})
@java.lang.SuppressWarnings("all")
@javax.annotation.Generated("lombok")
public TestBok(@NonNull final Integer id, final String foo) {
    if (_id == null) {
        throw new java.lang.NullPointerException("_id");
    }
    this._id = id;
    this._foo = foo;
}
@java.lang.SuppressWarnings("all")
@javax.annotation.Generated("lombok")
public void setId(@NonNull final Integer _id) {
    if (_id == null) {
[...]

Issue #869, opened 2015-07-14
@RequiredArgsConstructor causes compile error with lombok.accessors.prefix


lombokissues:
_Migrated from Google Code (issue 834)_

PR #1284, opened 2017-02-01
Fix for issues #869 and #1018 - Constructor null check bug when @NonNull with @Accessors

kchirls:
Generates the null check on the constructor parameter instead of the
instance field. Fix for issues #869 and #1018.


kchirls:
Hi rspilker@ or rzwitserloot@, Do you mind taking a quick look at this? This bug is inconvenient for persons or teams that use a custom variable name prefix style. Thanks!

rspilker:
So far, I've only looked at the source code on Github. The fix itself looks good, although I didn't see the modified after-ecj file, which means that the test for that file will fail.

Also, could you add your name the the AUTORS file so we are legally coverred?

If you cannot get the tests for ecj working or need any other help, please let us know.

kchirls:
ECJ fix and test now included. Name added to AUTHORS file. Changed one of the variables to be final, to account for that case. One minor line change (removing Generated annotation) to account for changes from the latest merge. Let me know if I missed anything. Thanks!

kchirls:
Hi @rspilker. Just a ping that I've made your recommended updates. Thanks!

rspilker:
Thanks a lot! Sorry it took so long to get merged.

PR #1292, opened 2017-02-08
Boundsetter v1.12.6 support

aiziyuer:
I fix some conflicting, and it work on tag v1.16.12, please merge it. 

https://github.com/rzwitserloot/lombok/pull/1110


rzwitserloot:
I need some context as to what problem this is trying to solve or what it does, especially considering the diff is hard to parse because of a bunch of style changes.

Gonna close it in a month without feedback.

PR #1311, opened 2017-02-22
Several minor fixes to make 'ant test' work more reliably on other systems

askoning:


rspilker:
Thanks. Better late then never...

PR #1328, opened 2017-03-08
Versionable capabilities for a DTO objects

lexfaraday:
Please see the readme for a more detailed info.

Thank you :)

### DIFF
--- a/src/core/lombok/Version.java
+++ b/src/core/lombok/Version.java
@@ -1,0 +1,26 @@
+package lombok;
+
+/**
+ * @See VersionableUtils
+ * @author Al3x G0m3z
+ *
+ */
+public enum Version
+{
+    v1,
+    v2,
+    v3,
+    v4,
+    v5,
+    v6,
+    v7,
+    v8,
+    v9,
+    v10,
+    v11,
+    v12,
+    v13,
+    v14,
+    v15,
+    v16
+}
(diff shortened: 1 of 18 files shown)


termosfera:
That would be great!

src/core/lombok/Version.java
```diff
@@ -0,0 +1,26 @@
...
+    v8,
+    v9,
+    v10,
+    v11,
+    v12,
+    v13,
+    v14,
+    v15,
+    v16
+}
```
Kjens93: Not sure if this is the most elegant solution to this problem. Using an enum like this restricts the user to only 16 versions... why don't you just have the annotation accept an `int`?

Maaartinus: I'd actually propose a String for the annotation (with int for the version number). Imagine a property present in versions 3 to 9, tentatively removed in version 10, put back temporarily in version 11 and finally removed in version 16. This sounds terribly complicated, but it's as simple as
`@Versionable(versions="3-9, 11-16")`
when using strings. With many versions, doing anything else is verbose and error-prone. For a property introduced in version 6, I'd use
`@Versionable(versions="6-")`
which means you don't have to touch the annotation as long as you keep the property.

lexfaraday: Thank you for your time guys,
The enum is a bad solution, I take a mistake so the concept of feature is the target, If you are agree with this I will fix them with int for example instead of a String.

Best Regards!
Alex

jzaruba: @lexfaraday The README does still involve the enum.

nnordrum:
I'll concede I just skimmed, but this seems like you're trying to solve a problem that could be delegated to more comprehensive systems like Avro/ProtoBuffs/Thrift/etc.

Would it not make more sense to wrap one/some/all of those instead?

lexfaraday:
Close until to find a new solution to be more flexible.

Issue #853, opened 2015-07-14
Builder should support fields from parent classes


lombokissues:
_Migrated from Google Code (issue 818)_

smougenot:
Would be nice to include the workaround in the official documentation (en @Builder)
ex :https://www.donneo.de/2015/09/16/lomboks-builder-annotation-and-inheritance/

```java
@AllArgsConstructor
public class Parent {
  private String a;
}

public class Child extends Parent {
  private String b;

  @Builder
  public Child(String a, String b){
    super(a);
    this.b = b;
  }
}

public static void main(String[] ...){
  Child.builder().a("testA").b("testB").build();
}
```

msarniak:
Builder pattern that supports inheritance without the chicken and egg problem - not the cleanest code, cluttered with generics, but since this is to be generated, maybe worth considering?

```java
public class Parent {

    private final int a;

    protected Parent(final AbstractParentBuilder builder) {
        this.a = builder.a;
    }

    public static final class Builder extends AbstractParentBuilder<Parent, Builder> {
        public Parent build() {
            return new Parent(this);
        }
    }

    protected static abstract class AbstractParentBuilder<T extends Parent, B extends AbstractParentBuilder<T, ?>> {

        private int a;

        public B withA(int a) {
            this.a = a;
            return (B) this;
        }
    }
}


public class Child extends Parent {

    private final int b;

    protected Child(final AbstractChildBuilder builder) {
        super(builder);
        this.b = builder.b;
    }

    public static final class Builder extends AbstractChildBuilder<Child, Builder> {
        public Child build() {
            return new Child(this);
        }
    }

    protected static abstract class AbstractChildBuilder<T extends Child, B extends AbstractChildBuilder<T, ?>> extends Parent.AbstractParentBuilder<T, B> {

        private int b;

        public B withB(int b) {
            this.b = b;
            return (B) this;
        }
    }
}
```

With that, code below is valid:
```
Child child = new Child.Builder().withB(1).withA(2).build();
```

We may get rid of casts, adding some abstract self() method, implemented in concrete Builders.

shakuzen:
I came here from peichhorn/lombok-pg#78, which has a ton of +1s. I hope the maintainers here are aware of this.

fahran:
+1

afedulov:
+1

ywilkof:
+1

electrobabe:
+1

gaetancollaud:
+1

[...]

PR #1337, opened 2017-03-15
Builder: Support for fields from parent classes

janrieke:
As a solution for #853 (and peichhorn/lombok-pg#78), I implemented a new feature for (type) builders, so that fields from the parent class can also be set using the builder.
The idea is to generate a constructor on the type that takes a builder as a parameter and uses the fields from the builder to set the fields of the new instance. The `Builder.build()` method calls this constructor.

The feature must be explicitly switched on, in order to not break existing implementations, which may rely on the all-args constructor generated by `@Builder`.


bitsal:
@janrieke could you fix merge conflicts and probably it will be merged?

janrieke:
@bitsal: I've tried to get a reply from @rzwitserloot several times, also via email. No success. I will not spend time fixing conflicts as long as don't know how the chances are that this will be included.

philippeboyd:
@janrieke did you get any news from @rzwitserloot ?

janrieke:
Unfortunately not.
Maybe @rspilker can say something?

philippeboyd:
@janrieke if @bitsal has write access, maybe he could merge your pull request? With the fixed conflicts of course.

bitsal:
It seems I don't have write access

KyleChamberlin:
is there a reason this PR hasn't gotten any official response from any of the lombok devs? if there are technical limitation or problems with this implementation, then we should hear about what is preventing this solution from being merged such that we can (as a community) attempt to find a better mergeable solution. 

@rzwitserloot can you make any comment on why this PR is deficient? 
@bitsal as a contributor can you help us to understand what would need to be done to move this (or any other solution to this problem) forward?

aikar:
@rzwitserloot @rspilker can you review this PR? would be great to land this.

rzwitserloot:
I have no idea why I have zero emails on this. Looking at it now.

rzwitserloot:
This needs lots of tests, documentation, changelog, etc. I've fixed the merge issues and done some codereview, as well as made a behavioural change (trying to inherit = true when you have no parent now fails instead of mostly silently doing nothing, I think - testing yet to commence).

Further development of this feature will occur in branch builderParentClass.

rzwitserloot:
@janrieke Can you write a pullreq in branch https://github.com/rzwitserloot/lombok/tree/builderParentClass where you add your name to the AUTHORS file?

rzwitserloot:
Allright, bad news. I'm pretty sure this feature is not gonna make it.

Here's the problem:

You can't use this feature unless you do NOT chain it. The whole point of builders it to chain it, so that makes this feature dead on arrival. Specifically, this must be possible:

Child  c = ChildClass.builder().fieldFromParent1(0).fieldFromChild(1).fieldFromParent2(2).build();

and there's no way to make that work; the fieldFromParent1 method returns a ParentBuilder and not a ChildBuilder. I tried some fancy generics footwork to get out of this, but [A] this gets you a boatload of unsafe casting warnings, [B] by convoluting the type of the builder, makes it basically impossible to use the builder in any way other than one-linering it, and [C] still doesn't work, for some reason the build() method doesn't wanna play ball, feels like a javac/ecj bug to me, but given that both javac and ecj exhibit this bizarre behaviour I guess it is somehow spec. It certainly highlights the usual problem with 'using fancy generics footwork to get out of a typing issue' which is that once you get beyond academic examples it generally just doesn't work well unless you add explicit type params at which point all semblance of utility is right out the window.

What am I missing, here, guys? some 90 thumbsup/hearts/whatsit and a massive pullreq.

Completely our mistake for not tracking pullreqs carefully (we're making changing on that front and will start investigating pullreqs as they come in from here on out), but how would this feature ever have worked?

Specifically, make this code not-ugly, and compile properly, and then we have something to target:

```
import java.util.List;
public class BuilderWithInherit {
	public static class Parent {
		int field1;
		
		protected Parent(final ParentBuilder<?, ?> b) {
			this.field1 = b.field1;
		}

		public static class ParentBuilder<Z extends Parent, B extends ParentBuilder<Z, B>> {
			private int field1;

			ParentBuilder() {}

			public B field1(final int field1) {
				this.field1 = field1;
				return (B) this;
			}

			public Z build() {
				return (Z) new Parent(this);
			}
		}

		public static ParentBuilder<Parent, ParentBuilder> builder() {
			return new ParentBuilder<Parent, ParentBuilder>();
		}
	}

	public static class Child extends Parent {
		double field2;

		protected Child(final ChildBuilder<?, ?> b) {
			super(b);
			this.field2 = b.field2;
		}

		public static class ChildBuilder<Z extends Child, B extends ChildBuilder<Z, B>> extends ParentBuilder<Z, B> {
			private double field2;

			ChildBuilder() {}

			public B field2(final double field2) {
				this.field2 = field2;
				return (B) this;
			}

			public Z build() {
				return (Z) new Child(this);
			}
		}

		public static ChildBuilder<ChildBuilder, Child> builder() {
			return new ChildBuilder<ChildBuilder, Child>();
		}
	}

	public static void test() {
		Child x = Child.builder().field1(5).field2(1.0).field1(2).build();
	}
}

```

aikar:
@rzwitserloot override all the getters/setters of the parent and redefine the methods with the ChildBuilder as a return.

This is valid:

```java
class A {
        private int a;
        public A a(int a) { this.a = a; return this; }
        public int getA() { return this.a; }
    }
    class B extends A {
        public B a(int a) { super.a(a); return this; }
        public int foo() { return super.getA(); }
    }
    class test {
        public void test() {
            (new B()).a(42).getA();
        }
    }```

**Comment by janrieke** [comment]
Great to have some feedback, although it took a while :)

Yes, the chaining/casting is a problem. I agree that, without solving it, this feature will look quite ugly when used in code. And as Lombok aims at making code nicer, I can understand that you do not want it merged, @rzwitserloot.

However, in the issue thread there was already a suggestion (https://github.com/rzwitserloot/lombok/issues/853#issuecomment-148284229) which seems to be implementable. It looks quite similar to what you suggested above. The generated code will be using generics heavily, which will make it difficult to understand. But as it's generated code, I think that this would be ok.

@rzwitserloot: If you think that such an approach could make it to Lombok, I would try to get that idea implemented.

**Comment by rzwitserloot** [comment]
@aikar That would require resolution which is tricky (lombok needs to run during compilation; asking the compilation process about the methods/fields in some external class is a problem; that's not done yet, and if we wait until it IS done, lombok's too late to add any methods. There are ways around it but they carry a heavy maintenance and performance burden).

@janrieke Assuming the generics can be sorted out, I'm still not convinced. This really uglies up those builders beyond all recognition. The best solution is to override the methods, but that gets back to requiring resolution which makes the maintenance burden too high for the feature.

**Comment by aikar** [comment]
Using Interfaces to help get around cyclic dependencies on the generics, I was able to get this almost working....

```java
    public interface IParentBuilder<Z extends Parent, B> {
        int field1();
        public B field1(final int field1);
        public Z build();
    }

    public static class Parent {
        int field1;

        protected <Z extends Parent, B extends IParentBuilder<Z, B>> Parent(final ParentBuilder<Z, B> b) {
            this.field1 = b.field1;
        }

        public static class ParentBuilder <Z extends Parent, B extends IParentBuilder<Z, B>> implements IParentBuilder<Z, B> {
            int field1;

            ParentBuilder() {}

            public int field1() {
                return field1;
            }

            public B field1(final int field1) {
                this.field1 = field1;
                return (B) this;
            }

            public Z build() {
                return (Z) new Parent(this);
            }
        }

        public static <T extends ParentBuilder<Child, T>> T builder() {
            return (T) new ParentBuilder();
        }
    }

    public interface IChildBuilder<Z extends Child, B> extends IParentBuilder <Z, B> {
        double field2();
        public B field2(final double field2);
        public Z build();
    }
    public static class Child extends Parent {
        double field2;

        protected <Z extends Child, B extends IChildBuilder<Z, B>> Child(final ChildBuilder<Z, B> b) {
            super(b);
            this.field2 = b.field2;
        }

        public static class ChildBuilder<Z extends Child, B extends IChildBuilder<Z, B>>
                extends ParentBuilder<Z, B> implements IChildBuilder<Z, B> {
            double field2;

            ChildBuilder() {}

            public double field2() {
                return field2;
            }
            public B field2(final double field2) {
                this.field2 = field2;
                return (B) this;
            }

            public Z build() {
                return (Z) new Child(this);
            }
        }

        public static <T extends ChildBuilder<Child, T>> T builder() {
            return (T) new ChildBuilder();
        }
    }

    public static void test() {
        Child x = Child.builder().field1(5).field2(1.0).field1(2).build();
    }

```

The issue now is clashes in erasure on the builder method :/ If you rename the parent or childs builder method, the code is then valid.

Any ideas to work around that?
@rzwitserloot

aikar:
A bridge method could solve that final detail right?

janrieke:
@aikar My solution is quite similar, except that I use abstract builder classes instead of interfaces (which I personally find a bit nicer, because you do not have to generate the setter method signature twice). Implemented in this way, there are no compiler warnings, and builder method chaining works nicely.

But I do not see a solution for overriding the `builder()` method in subclasses. Using generics there as return type compiles: `public static <B extends AbstractParentBuilder<?, ?>> B builder()`
However, you cannot call such a method from your code, as this produces two different methods, which makes the method call ambiguous for the compiler.
But maybe it's ok to require manually setting a different builder method name on subclasses?

janrieke:
Made it ;)
In this way, it compiles without any warnings, chaining works, and the builder methods can have the same name.

Superclass:
```java
public class Superclass {

    final int field1;

    protected Superclass(final AbstractSuperclassBuilder<?, ?> b) {
        this.field1 = b.field1;
    }

    public static final class Builder extends AbstractSuperclassBuilder<Superclass, Builder> {
        @Override
        public Superclass build() {
            return new Superclass(this);
        }

        @Override
        protected Builder getInstance() {
            return this;
        }
    }

    protected abstract static class AbstractSuperclassBuilder<C, B extends AbstractSuperclassBuilder<?, ?>> {
        private int field1;

        public B field1(final int field1) {
            this.field1 = field1;
            return getInstance();
        }

        protected abstract B getInstance();
        protected abstract C build();
    }

    public static AbstractSuperclassBuilder<? extends Superclass, ? extends AbstractSuperclassBuilder<? extends Superclass, ?>> builder() {
        return new Builder();
    }
}
```

Subclass:
```java
public class Subclass extends Superclass {

    final double field2;

    protected Subclass(final AbstractSubclassBuilder<?, ?> b) {
        super(b);
        this.field2 = b.field2;
    }

    public static final class Builder extends AbstractSubclassBuilder<Subclass, Builder> {
        @Override
        public Subclass build() {
            return new Subclass(this);
        }
        
        @Override
        protected Builder getInstance() {
            return this;
        }
    }

    protected abstract static class AbstractSubclassBuilder<C, B extends AbstractSubclassBuilder<Subclass, B>>
        extends Superclass.AbstractSuperclassBuilder<Subclass, B> {
        private double field2;

        public B field2(final double field2) {
            this.field2 = field2;
            return getInstance();
        }
        
        @Override
        protected abstract B getInstance();
    }

    public static AbstractSubclassBuilder<? extends Superclass, ? extends AbstractSubclassBuilder<? extends Subclass, ?>> builder() {
        return new Builder();
    }
}
```

It's really crammed with generics and thus not the nicest code. But I think this type of builders will be generatable by Lombok. Or do I miss something?
Should I give it a try implementing, @rzwitserloot?

rzwitserloot:
Yes, let's do it. I've done the following already:

* I've fixed the merge conflicts in this pullreq. We're going in a slightly different direction but it's still useful as a jumping off point I think.
* I've written a single test case + the expected delombok output.
* I've written the annotation.

TODO:
* The after-delombok SuperBuilderBasic might not be entirely correct, but it should be pretty close.
* The after-ecj of SuperBuilderBasic
* Another test case which introduces generics on Parent, carried on by Child.

changes to the spec I've made:
* Instead of having properties on `@Builder`, introduced a new annotation, `@lombok.experimental.SuperBuilder`. This has many advantages:

1. We can put it in experimental.
2. We can spec it to only be allowed on types.
3. We can get rid of builderClassName, and with that, superbuilderClassName, simplifying things considerably.
4. Every link in the chain gets the same treatment, so inherit and chainable are effectively superfluous; it really just needed 'participatesInTheExtensionSystem'. But that's an unwieldy name. Not saying 'SuperBuilder' is a homerun either, but, hey, it's less of a mouthful :)
5. renamed getInstance to self.

* toBuilder is also going to be much more complicated for superbuilders, but I've remarked it out for now, let's tackle that as a second milestone later.


The above all in branch builderParentClass (Commit 3eea3b9)

janrieke:
I came up with the same name (`@SuperBuilder`), and also figured out most of the generics generation for javac. Eclipse still needs to be done, and some problems to be solved when there are generics present on the annotated class.
So I'll create a new PR based on your `builderParentClass` branch, revert the previous changes to the existing `HandleBuilder`s, and create new `HandleSuperBuilder`s. (It will take some days, as I'm currently a bit busy.)

janrieke:
Just a short status update: Builders with generics on the annotated class turned out to be a bit more complex to generate than expected. However, javac implementation is working now. I just started with the Eclipse implementation.

wesblume23:
Any update on this?

janrieke:
Eclipse implementation is finished to most parts. I'm currently setting up several test cases to make sure everything works as expected.
You can watch the progress in the branch https://github.com/janrieke/lombok/tree/superBuilder

janrieke:
I created new pull request #1711.
Further discussions over there. Closing here.

Issue #1280, opened 2017-01-20
@SafeCall annotation on local variables as an analogue of the Safe Navigation operator in Groovy (?.)


bulgakovalexander:
Hi @rspilker. 
Java language doesn't have ?. operator for safe calling methods of class as in Groovy or Kotlin languages. 
It is impossible to compile code like this: 
`String version = computer?.getSoundcard()?.getUSB()?.getVersion();`
But I'm trying to make some analogue using annotation.
 `@SafeCall String version = computer.getSoundcard().getUSB().getVersion();`
Example of code transformation: [source code](https://raw.githubusercontent.com/bulgakovalexander/lombok/feature/safeCall/test/transform/resource/before/SafeNullVal.java), [code after delombok](https://raw.githubusercontent.com/bulgakovalexander/lombok/feature/safeCall/test/transform/resource/after-delombok/SafeNullVal.java).
Currently the feature is under development, but I hope to finish it.
What do you think about this feature? Would you like to include it in Lombok project?
 [Link to working branch](https://github.com/bulgakovalexander/lombok/tree/feature/safeCall)

Maaartinus:
Long time ago, I started a [lengthy discussion][1] with Reinier and others. There's no real conclusion there, but there's an analysis of tons of corner cases.

In the meantime, I don't believe, it's as useful as I thought initially. IMHO the real problem is the lack of a null-safe type system. Having a nice way to write null-safe expressions doesn't help with the places where I forget to use it. So I'm avoiding nulls everywhere and the possible use cases for the null-safe expressions get avoided, too.

[1]: https://groups.google.com/d/msg/project-lombok/2c9Txx8SKGc/Ab_6OwClFPcJ

rzwitserloot:
I like it, but when we did some thinking on this, we arrived at a method call instead of an annotation. Something like:

    String version = nullCarry(computer.getSoundcard().getUSB().getVersion());

or even better:

    String version = onNull(computer.getSoundcard().getUSB().getVersion(), "1.0.0");

which produces 1.0.0 if any part of the expression is null (including the final call!). onNull is a static utility method in Lombok, and lombok magically fixes things when you call it. To include it, you'd add import static lombok.Lombok.onNull;

This lets you:

[...]

PR #1402, opened 2017-05-24
Feature/safe call

m4gshm:
Further to the topic #1280, I has finished the implementation of the `@SafeCall` feature.
Except the protection from 'NullPointerException' also 'ArrayIndexOutOfBoundsException' and 'NegativeArraySizeException' have been added.
Please take a look at the test 'SafeCallArray.java'. It describes these protections.
@rzwitserloot suggests a good idea about a default value that should be assigned instead of null.
That feature is implemented using the conditional operator (?:) and can be expressed like:
`@SafeCall int intVar = getNullInteger() ? intVar : -1;`
See tests 'SafeCallDefaultValue.java' and 'SafeCallIllegalUsingDefaultValue.java'.
About the magic method-wrapper "onNull", this is an interesting idea, but I didn't analyse and didn't try to implement that feature so cannot say how hard it is. It needs time.


gixlg:
Hi, this is a very interesting feature, I'm wondering why this PR is not merged. 
It's open since 2017.

PR #1409, opened 2017-06-01
usage_examples:Removed @NotNull from Vanilla Java

ishaanbahal:
Vanilla java example used @NotNull field in the constructor and import. Removed it.


rspilker:
duplicate

Issue #1347, opened 2017-03-29
Builder.Default causes fields to be null when instantiating with new


kevcodez:
When using the `@Builder.Default` annotation on a list/set/map and instantiating the object using the new keyword, the field is null.

```java
@Data
@Builder
@AllArgsConstructor
@NoArgsConstructor
@Entity
public class MyClass {
    
   @OneToMany(mappedBy = "orderItem", cascade = CascadeType.ALL)
   @Builder.Default
   private List<Entitlement> entitlements = new ArrayList<>();

}
```

```java
// entitlements is null, should be empty ArrayList
MyClass myClass = new MyClass();
```

heapifyman:
This issue also occures with any other non-primitive type, not only list/set/map.

```#java
@Getter
@Setter
@AllArgsConstructor
@NoArgsConstructor
@Builder
@ToString(callSuper = true)
public class Test {

    public static final int DEFAULT_INT = 42;

    public enum Status {
        STATUS_ONE,
        STATUS_TWO;
    }

    @Builder.Default
    private Integer someInt = DEFAULT_INT;

    @Builder.Default
    private Status currentStatus = Status.STATUS_ONE;

    public static void main(String[] args) {
        Test test = new Test();
        System.out.println("test = " + test);
    }

}
```

TokenR1ng:
This Issue makes the Release 1.16.16 totally unusable for us. Hope this gets fixed in the next Release since we could really use `@Builder.Default` Annotation.

wgorski-antrego:
Are there any plans for fixing this? To be honest, it's a HUGE problem as it's super unintuitive and introduces bugs that are really hard to find. To be honest, the most intuitive way of generating a default value for the builder would be to take the value from the field initialization, instead of requiring an additional Builder.Default annotation.

joaodelgado:
From a quick look at the source code, I found that the `@Builder.Default` is clearing the initialization part of every annotated field. Line 585:
https://github.com/rzwitserloot/lombok/blob/e92e285d95e0fb08c32e7059f85806a30985084b/src/core/lombok/javac/handlers/HandleBuilder.java#L580-L590

What's the purpose of clearing this initialization? Is it because of potential side effects of the initialization expression?

omega09:
@joaodelgado See [here](https://github.com/rzwitserloot/lombok/pull/1429). Initialization has been moved to the builder so if you don't nullify the field it will be initialized twice.

davidje13:
[...]

PR #1429, opened 2017-06-29
Fixes issue with BuilderDefault used with Builder and NoArgsConstructor.

philliprower:
Below tests demos the issue this change fixes.  Basically if you want to use both @Builder and @NoArgsConstructor which we do because of legacy code, the @Builder.Default breaks the default initialization for @NoArgsConstructor.  We'd like the defaults values initialized when using both the builder and no arg construction of the object.

```
public class BuilderDefaultIssueUnitTest {
    @Test
    public void test() {
        Example example1 = Example.builder().build();
        Example example2 = new Example();

        assertEquals(example1.getSomeDefInitInteger(),example2.getSomeDefInitInteger());
        assertEquals(example2.getSomeDefInitString(),example1.getSomeDefInitString());
    }

    @Data
    @Builder
    @AllArgsConstructor
    @NoArgsConstructor
    public static class Example {
        @Builder.Default
        private int someDefInitInteger = 5;
        @Builder.Default
        private String someDefInitString = "Hello";
    }
}
```

### DIFF
--- a/src/core/lombok/eclipse/handlers/HandleBuilder.java
+++ b/src/core/lombok/eclipse/handlers/HandleBuilder.java
@@ -644,7 +644,6 @@ public class HandleBuilder extends EclipseAnnotationHandler<Builder> {
 		FieldDeclaration fd = (FieldDeclaration) fieldNode.get();
 		out.returnType = copyType(fd.type, source);
 		out.statements = new Statement[] {new ReturnStatement(fd.initialization, pS, pE)};
-		fd.initialization = null;
 		
 		out.traverse(new SetGeneratedByVisitor(source), ((TypeDeclaration) fieldNode.up().get()).scope);
 		return out;
--- a/src/core/lombok/javac/handlers/HandleBuilder.java
+++ b/src/core/lombok/javac/handlers/HandleBuilder.java
@@ -582,7 +582,6 @@ public class HandleBuilder extends JavacAnnotationHandler<Builder> {
 		JCVariableDecl field = (JCVariableDecl) fieldNode.get();
 		
 		JCStatement statement = maker.Return(field.init);
-		field.init = null;
 		
 		JCBlock body = maker.Block(0, List.<JCStatement>of(statement));
 		int modifiers = Flags.PRIVATE | Flags.STATIC;


rspilker:
Well, it fixes one problem and introduces another. If the initializer is a "complex operation", for every object created by the builder it is now executed twice.

Reinier and I are aware of the problem and are considering several solutions, but we have no clear winner yet. Possibly, the best solution is to duplicate the initializer code in the generated no-args constructor, and possible also in the applicable required args constructors if the fields are nullable.

For now, the workaroun is to type a no-args constructor and initialize the fields with the same expression in the constructor.

philliprower:
I see... I think your probably right generating the code in the no arg constructor is probably best.

luolong:
I was just pointed at this issue by a colleague of mine who found this behavior and was unpleasantly surprised by it.

I do believe that the argument that @rspilker pointed out of of optimizing potentially expensive initialization logic is not really in the responsibility of the tool, but rather a responsibility of the developer.

I have no concrete metrics to offer, but my gut feeling based on the kind of code bases I've seen (and believe me -- none of them have been shining examples of Clean Code or beacons of hope for believers in SOLID architecture), most of the field initialization logic has been rather trivial static default values (new empty arrays, constant literal values, null, etc.); Very few involve minimal calculation and extremely rarely is there an expensive computation involved.

Optimising for rare conditions seems like a weird choice given the widespread use and generality of the tool. And one that introduces surprising and rather unwelcome results for the end users.

So, from my completely biased and uninformed position, the current solution breaks more expectations than it solves any existing problems. 

So, I would of course prefer if Lombok generated builders would be able to solve the expensive initialization issue in a way that it would be guaranteed to be performed only once without much extra effort from developer, but barring that, I would expect the developer to know how to counter such cases without

Maaartinus:
> Very few involve minimal calculation and extremely rarely is there an expensive computation involved.

Sure, it's hardly ever possible to do *anything useful* in the initializer expression as it's executed before the constructor and can't depend on constructor arguments. Anything except initializing to suitable constant expression is very rare. The most expensive thing, I've ever seen there was a 16 kB buffer allocation and a `static AtomicInteger` instance counter increment (which is a bad design). Anyway, such cases are rather exceptional and then there's usually no use for a Builder.

citDennis:
Please fix this. I just lost a day's work when I ran into this after refactoring an extensive codebase. Its not mentioned in the documentation and I made the mistake of trusting that. Another complicating factor was that de-lombok from the GUI (plugin) didn't show it up but tests proved 1.16.16 was broken in this respect. Very very confusing altogether. 
I still love Lombok and will continue to recommend it but this really saddens me.

rzwitserloot:
Fixed in a different way, see issue #1347 .

citDennis:
Thanks Ronald, we removed most of the offending builders for now. Which maven release do you expect this to be included in?

PR #1438, opened 2017-07-11
NonNullExample_post.jpage - remove lombok from vanilla Java

shaulzorea:


rspilker:
We intentionally kept it there because you probably want to document, either in (javadoc) comments, inline comments or using an annotation that the parameter should not be null.

```java
public NonNullExample(/* non-null */ Person person) {
}
```

Since this is just a placeholder, we use an annotation we happen to have already.

PR #1466, opened 2017-09-09
remove @NonNull from "Vanilla Java" ConstructorExample

jersou:


rspilker:
We intentionally kept it there because even in vanilla java you probably would put some sort of documentation on the field. Alternatively you would use comments:

```java
// never null, required field, etc.
private T description
```

PR #1481, opened 2017-09-29
Funct rascal transformation

Adamacenog:
Using the 'rascal' tool we applied a FUNCT transformation.


rzwitserloot:
I don't know why you'd send a pull request for such a thing, but obviously we can't accept a pull request whose only purpose is to thoroughly mess up and uglify the codebase. Are you doing this for other github projects? Please, pretty please, stop.

Issue #1393, opened 2017-05-18
Allow methods to be used in @ToString and @EqualsAndHashCode


Maaartinus:
There are many requests for customizing `@ToString` and other annotations, so I'm proposing a simple way (based on ideas from this [discussion][1]) how many such requests can be easily fullfilled: Allow to exclude fields (already possible) and include methods in the computation. So the solution for #1392 would be simply

```
@ToString
class Article {
    private String author;
    private String title;
    @ToString.Exclude
    private String content;

    @ToString.Include
    private String shortenedContent() {
        return content.substring(0, Math.min(content.length, 100));
    }
}
```

This is surely more verbose than some `@ToString(maxLength=100)`, but imagine you need to append the field length. Or there's another field which should never be shortened. Or another field `signature` which is uninteresting, but you want to see its length. This all can be achieved easily with methods inclusion.

---

An alternative stringly typed syntax: `@ToString(exclude="content", include="shortenedContent()")`.

[1]: https://groups.google.com/d/msg/project-lombok/dc5nXTI3xMU/tkgk0OfbAQAJ

omega09:
Includes #1385?

Maaartinus:
@omega09 I don't think, it "includes" it, especially it wouldn't allow you to call the Apache builders. But it'd allow you to do about everything their builders can do, except for [styling][], which is an orthogonal feature.

My proposal is good in case you want to change what happens for some fields while leaving lombok do its usual job for the others. Leaving out an offending field has always been possible, but sometimes, that's too harsh. When you want to summarize it instead, my proposal helps. It tries to combine the flexibility of a hand-written method with Lombok's power of handling boilderplate for free.

[styling]: https://commons.apache.org/proper/commons-lang/javadocs/api-3.1/org/apache/commons/lang3/builder/ToStringStyle.html

liweys:
It will be good if these methods can be customized. I would like to use json presentation for toString().

rzwitserloot:
Working on it: Working on it: https://github.com/rzwitserloot/lombok/wiki/FEATURE-IDEA:-Annotation-based-Include-Exclude-for-ToString,-EqualsAHC,-etc

Issue #1163, opened 2016-07-27
Define excludes from @ToString and @EqualsAndHashCode on the field


srsp:
At the moment the only way to exclude fields from `@ToString` or `@EqualsAndHashCode` is: 

```
@ToString(exclude = "bar")
@EqualsAndHashCode(exclude = "bar")
public class Foo {

  private String bar;

}
```

I think something like this would be nice:

```
@ToString
@EqualsAndHashCode
public class Foo {

  @ToStringExclude
  @EqualsAndHashCodeExclude
  private String bar;

}
```

I would prefer this solution for several reasons:
- You don't have to type the field's name.
- Less hassle when renaming or deleting the field.  
- The information of whether the field is included / excluded in `@ToString` is bundled with field definition.

The annotation name is just an idea. You could also consider something in lines of `@ToString(exclude = true)` on the field.

What do you think?

rzwitserloot:
Working on it: https://github.com/rzwitserloot/lombok/wiki/FEATURE-IDEA:-Annotation-based-Include-Exclude-for-ToString,-EqualsAHC,-etc

Issue #123, opened 2015-07-14
Annotate fields for inclusion/exclusion in toString and EqualsHashCode


lombokissues:
_Migrated from Google Code (issue 50)_

rzwitserloot:
Working on it: https://github.com/rzwitserloot/lombok/wiki/FEATURE-IDEA:-Annotation-based-Include-Exclude-for-ToString,-EqualsAHC,-etc

Issue #512, opened 2015-07-14
Add (experimental) support of java.lang.Comparable


lombokissues:
_Migrated from Google Code (issue 439)_

rzwitserloot:
Proposal:

* We deprecate `@EqualsAndHashCode` (but it'll never be removed, really, or at least not for a long while).

* Using both `@EqAHC` and `@Ordered` at the same time is invalid.

* `@Data` and `@Value` switch over from implying `@EqAHC` to implying `@Ordered` (without comparable=true!)

* We create a new annotation, `@Ordered`. It broadly fills the same role as `@EqAHC`.

* `of` and `exclude` go away. `comparable = true/false` (default = depends on whether you added implements Comparable to your implements list) is added. `onParam` is renamed to `onEqualsParam`, and `onCompareToParam` is added (oCTP without comparable=(default) true is an error).

* the same batch of fields is used for all 4 methods (`canEqual`, `equals`, `hashCode`, and `compareTo`).

* For comparable=true, any field so included must either be a primitive (we know how to compare this intrinsically), or must implement `Comparable`. We don't check this; we just call `.compareTo(o)`, and if that's not a valid action the compiler will complain about generated code. This means we do not have to do resolution which is a good thing.

* Trying to compare arrays is a lombok-generated error; you must specify a comparison method/comparator class or exclude the field.

* If comparable=true, we don't just generate `compareTo`, we also add `implements Comparable<SelfType>` to the implements list if it isn't already there.

* To configure individual fields, add sub-annotations of ordered:

    * `@Ordered.Include`. If present, any use of `@Ordered.Exclude` anywhere in this class is invalid. All fields without an `@Ordered.Include` are excluded if any field is so annotated.
    * `@Ordered.Exclude`. Nuff said.
    * `@Ordered.NullOrder(NullOrder.NULL_IS_MAX -or- NullOrder.NULL_IS_MIN -or- NullOrder.NULL_IS_ERR)`. Should the default behaviour be NULL_IS_MIN, or NULL_IS_ERR?
    * `@Ordered.DoNotUseGetter`; overrides the doNotUseGetters of the @Ordered annotation.
[...]

PR #1492, opened 2017-10-12
Field-based @ToString and @EqualsAndHashCode of and exclude parameters

sim642:
This PR adds `Of` and `Exclude` subannotations for `@ToString` and `@EqualsAndHashCode` to allow specifying explicitly included or excluded fields directly at the field declaration, not as variable name strings in the annotation parameter on the class. Past issues requesting such feature: #1163 and #123. Subannotation syntax similar to #1393.


Maaartinus:
Cool!

I'd prefer `Include` to `Of` as it's more obviously an opposite. While `@ToString(of={"a", "b"})` reads better than `@ToString(include={"a", "b"})`, for the repeated `@ToString.Of`, it's IMHO no more true. Especially, when it gets arguments like
```
@ToString.Of(limit=80, omitIfNull=true)
String someString;
```
or something like I proposed in https://github.com/rzwitserloot/lombok/issues/512#issuecomment-171851841.

sim642:
The `Of` vs `Include` naming issue definitely occurred to me, especially because internally the handlers call them `includes` as well and I agree with your reasoning. At the moment I went with `Of` because it closely emphasizes the subannotation's relationship to the `of` annotation parameter and follows the existing naming of things. Would definitely be good to hear some more opinions about this and I'm willing to make the necessary changes if needed.

rzwitserloot:
Not sure we're gonna go with this PR, but, we're working on this. See: Working on it: https://github.com/rzwitserloot/lombok/wiki/FEATURE-IDEA:-Annotation-based-Include-Exclude-for-ToString,-EqualsAHC,-etc

rzwitserloot:
ToString take on this is now on the master branch as a trial: ad21a1573bab57c63ffd5b9867f8e19ac7f0c94b

We'll expand it to EqualsAndHashCode and probably a newly minted `@SomeArgsConstructor` too.

Issue #1440, opened 2017-07-13
Lombok doesn't work with error-prone


KroArtem:
As it's stated in ChangeLog for 1.16.18,
`BUGFIX: Lombok now works with Bazel and Error Prone. Issue #1290`
However it doesn't seem to work correctly.
Lombok 1.16.18 + error-prone (plexus-compiler-javac-errorprone 2.8.2, error_prone_core 2.0.21) as it's stated in [installation](http://errorprone.info/docs/installation), maven build.

When I try to run tests, it fails with the following message:

```
Information:java: Errors occurred while compiling module 'tests'
Information:error-prone compiler was used to compile java sources
Information:13.07.2017 15:30 - Compilation completed with 17 errors and 1 warning in 3s 554ms
D:\qa_***\tests\src\test\java\com\***\tests\***.java
    Warning:Warning:line (29)java: lombok.javac.apt.LombokProcessor could not be initialized. Lombok will not run during this compilation: java.lang.IllegalArgumentException: com.sun.tools.javac.api.ClientCodeWrapper$WrappedStandardJavaFileManager extends com.sun.tools.javac.api.ClientCodeWrapper$WrappedJavaFileManager implements javax.tools.StandardJavaFileManager

  	at lombok.javac.apt.LombokFileObjects.getCompiler(LombokFileObjects.java:130)

  	at lombok.javac.apt.InterceptingJavaFileManager.<init>(InterceptingJavaFileManager.java:40)

  	at lombok.javac.apt.LombokProcessor.placePostCompileAndDontMakeForceRoundDummiesHook(LombokProcessor.java:164)

  	at lombok.javac.apt.LombokProcessor.init(LombokProcessor.java:85)

  	at lombok.core.AnnotationProcessor$JavacDescriptor.want(AnnotationProcessor.java:87)

  	at lombok.core.AnnotationProcessor.init(AnnotationProcessor.java:140)

  	at lombok.launch.AnnotationProcessorHider$AnnotationProcessor.init(AnnotationProcessor.java:69)

  	at com.sun.tools.javac.processing.JavacProcessingEnvironment$ProcessorState.<init>(JavacProcessingEnvironment.java:539)

  	at com.sun.tools.javac.processing.JavacProcessingEnvironment$DiscoveredProcessors$ProcessorStateIterator.next(JavacProcessingEnvironment.java:636)

[...]

Issue #1472, opened 2017-09-24
Java 9 Annotation Processing Broken


gudenau:
I am using Lombok 1.16.18 on JDK 9 and IntelliJ Idea 2017.2.4. 

When using Lombok's annotation processor it causes the compiler to throw an exception.
`Warning:(8, 8) java: lombok.javac.apt.LombokProcessor could not be initialized. Lombok will not run during this compilation: java.lang.IllegalArgumentException: com.sun.tools.javac.api.ClientCodeWrapper$WrappedStandardJavaFileManager extends com.sun.tools.javac.api.ClientCodeWrapper$WrappedJavaFileManager implements javax.tools.StandardJavaFileManager
  	at lombok.javac.apt.LombokFileObjects.getCompiler(LombokFileObjects.java:130)
  	at lombok.javac.apt.InterceptingJavaFileManager.<init>(InterceptingJavaFileManager.java:40)
  	at lombok.javac.apt.LombokProcessor.placePostCompileAndDontMakeForceRoundDummiesHook(LombokProcessor.java:164)
  	at lombok.javac.apt.LombokProcessor.init(LombokProcessor.java:85)
  	at lombok.core.AnnotationProcessor$JavacDescriptor.want(AnnotationProcessor.java:87)
  	at lombok.core.AnnotationProcessor.init(AnnotationProcessor.java:140)
  	at lombok.launch.AnnotationProcessorHider$AnnotationProcessor.init(AnnotationProcessor.java:69)
  	at jdk.compiler/com.sun.tools.javac.processing.JavacProcessingEnvironment$ProcessorState.<init>(JavacProcessingEnvironment.java:675)
  	at jdk.compiler/com.sun.tools.javac.processing.JavacProcessingEnvironment$DiscoveredProcessors$ProcessorStateIterator.next(JavacProcessingEnvironment.java:774)
  	at jdk.compiler/com.sun.tools.javac.processing.JavacProcessingEnvironment.discoverAndRunProcs(JavacProcessingEnvironment.java:869)
  	at jdk.compiler/com.sun.tools.javac.processing.JavacProcessingEnvironment.access$2200(JavacProcessingEnvironment.java:108)
  	at jdk.compiler/com.sun.tools.javac.processing.JavacProcessingEnvironment$Round.run(JavacProcessingEnvironment.java:1206)
[...]

Issue #1435, opened 2017-07-06
1.16.18 does not work with Netbeans Dev + Java 9


md-5:
Relates to #985.

When using 1.16.16, Java 9+174 and a recent build of Netbeans (eg 201706210001) there are no issues. When upgrading to 1.16.18, Netbeans no longer recognises any lombok constructs (getters, constructors etc), leading to syntax errors.

The below error is likely the cause and of relevance:

```
lombok.javac.apt.LombokProcessor could not be initialized. Lombok will not run during this compilation: java.lang.IllegalArgumentException: org.netbeans.modules.java.source.parsing.ProxyFileManager extends java.lang.Object implements javax.tools.JavaFileManager
  	at lombok.javac.apt.LombokFileObjects.getCompiler(LombokFileObjects.java:130)
  	at lombok.javac.apt.InterceptingJavaFileManager.<init>(InterceptingJavaFileManager.java:40)
  	at lombok.javac.apt.LombokProcessor.placePostCompileAndDontMakeForceRoundDummiesHook(LombokProcessor.java:164)
  	at lombok.javac.apt.LombokProcessor.init(LombokProcessor.java:85)
  	at lombok.core.AnnotationProcessor$JavacDescriptor.want(AnnotationProcessor.java:87)
  	at lombok.core.AnnotationProcessor.init(AnnotationProcessor.java:140)
  	at lombok.launch.AnnotationProcessorHider$AnnotationProcessor.init(AnnotationProcessor.java:69)
  	at com.sun.tools.javac.processing.JavacProcessingEnvironment$ProcessorState.<init>(JavacProcessingEnvironment.java:684)
  	at com.sun.tools.javac.processing.JavacProcessingEnvironment$DiscoveredProcessors$ProcessorStateIterator.next(JavacProcessingEnvironment.java:786)
  	at com.sun.tools.javac.processing.JavacProcessingEnvironment.discoverAndRunProcs(JavacProcessingEnvironment.java:881)
  	at com.sun.tools.javac.processing.JavacProcessingEnvironment.access$2400(JavacProcessingEnvironment.java:110)
  	at com.sun.tools.javac.processing.JavacProcessingEnvironment$Round.run(JavacProcessingEnvironment.java:1240)
  	at com.sun.tools.javac.processing.JavacProcessingEnvironment.doProcessing(JavacProcessingEnvironment.java:1319)
[...]

PR #1495, opened 2017-10-15
fix issue #1440, #1472: wrap BasedFileManager for jdk9

tmurakam:
I've created a workaround to fix the issue #1440, #1472 for IntelliJ + Java9.
It works for me with IntelliJ 2017.3 EAP.

The wrappedManager is proxied with java.lang.reflect.Proxy, so it can't be cast to BaseFileManager.
So I created BaseFileManagerWrapper to wrap it.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -20,6 +20,7 @@ Robbert Jan Grootjans <grootjans@gmail.com>
 Roel Spilker <r.spilker@gmail.com>
 Sander Koning <askoning@gmail.com>
 Szymon Pacanowski <spacanowski@gmail.com>
+Takuya Murakami <tmurakam@tmurakam.org>
 Taiki Sugawara <buzz.taiki@gmail.com>
 Thomas Darimont <thomas.darimont@gmail.com>
 Yun Zhi Lin <yun@yunspace.com>
--- a/src/core/lombok/javac/apt/LombokFileObjects.java
+++ b/src/core/lombok/javac/apt/LombokFileObjects.java
@@ -22,13 +22,18 @@
 
 package lombok.javac.apt;
 
+import java.io.IOException;
 import java.lang.reflect.Field;
 import java.lang.reflect.Method;
 import java.net.URI;
+import java.nio.charset.StandardCharsets;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.util.Iterator;
+import java.util.Set;
 import java.util.concurrent.atomic.AtomicBoolean;
 
+import javax.tools.FileObject;
 import javax.tools.JavaFileManager;
 import javax.tools.JavaFileObject;
 import javax.tools.JavaFileObject.Kind;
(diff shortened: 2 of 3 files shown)


rspilker:
Thanks for the pull request.

Can you add your name the the AUTHORS file, so we are legally covered?

Also, there is still a TODO in the code. Did you have a look at how the charset is used?

tmurakam:
I've added my name to AUTHORS.

And I updated the TODO code to use default charset of JVM.
In BaseFileManager class, the charset is used to decide charset in decode() method.

rspilker:
I think we need to get the encoding passed to the compiler. If you can find out how to do that, great. Otherwise I will probably be able to figure it out.

tmurakam:
> I think we need to get the encoding passed to the compiler. If you can find out how to do that, great. 

I digged the implementation of IntelliJ...

The wrappedManager is created in [org.jetbrains.jps.javac.JavacMain](https://github.com/JetBrains/intellij-community/blob/master/jps/jps-builders-6/src/org/jetbrains/jps/javac/JavacMain.java).wrapWithCallDispatcher(), and it proxies JavaFileManager to [org.jetbrains.jps.javac.JavacFileManager](https://github.com/JetBrains/intellij-community/blob/master/jps/jps-builders-6/src/org/jetbrains/jps/javac/JavacFileManager.java).

One possible way to get encoding is get  [OutputFileObject](https://github.com/JetBrains/intellij-community/blob/master/jps/jps-builders-6/src/org/jetbrains/jps/javac/OutputFileObject.java) instance using JavaFileManager#getJavaFileForOutput(), and access myEncodingName private field.

tmurakam:
I successfully get the encoding. The code is 4298d97.
But it is very ugly, and depends intellij internal implementation...

wrprice:
Note that I get a Proxy-wrapped manager using Win10/Gradle and has nothing to do w/ IntelliJ.  So there needs to be a wrapper solution that _does not_ depend on IntelliJ internals.  You can pass `null` as the charset and it will pick the platform default.  I'm _not saying that's proper behavior_, but it is "legal" per the API.

rspilker:
As far as I know, both compilers use JavaC uder the hood, I'll see if it is possible to get to the encoding passed to the compiler.

rspilker:
Alternatively, and if everything else fails, we can use the default platform encoding unless a lombok specific property is set.

rspilker:
@wrprice Can you point me to the code that gradle uses to create the wrapper?

wrprice:
I don't know at this time, but I'll start looking.

Per my comments on #1472 I was coming at this from the other side. I know the end result is essentially the same, but not the cause or its location.

wrprice:
Well, that was quick.  A GitHub search for `StandardJavaFileManager` in the Gradle repository turns up one hit:

[org.gradle.api.internal.tasks.compile.JdkJavaCompiler.java:64](https://github.com/gradle/gradle/blob/573df4bfb335c6dc6e1dde6a65da9fbbfe99f99e/subprojects/language-java/src/main/java/org/gradle/api/internal/tasks/compile/JdkJavaCompiler.java#L64)

md-5:
What about #1435, all these issues seem related regressions in .18

PR #1510, opened 2017-11-04
fix JavaDoc of callSuper in the ToString annotation

pcornelissen:
It looks like there was a copy paste error in the javadoc it mentioned equals in the toString docs

### DIFF
--- a/src/core/lombok/ToString.java
+++ b/src/core/lombok/ToString.java
@@ -64,7 +64,7 @@ public @interface ToString {
 	 * Include the result of the superclass's implementation of {@code toString} in the output.
 	 * <strong>default: false</strong>
 	 * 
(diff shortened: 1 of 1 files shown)


src/core/lombok/ToString.java
```diff
@@ -64,7 +64,7 @@
 	 * Include the result of the superclass's implementation of {@code toString} in the output.
 	 * <strong>default: false</strong>
 	 * 
-	 * @return Whether to call the superclass's {@code equals} implementation as part of the generated equals algorithm.
+	 * @return Whether to call the superclass's {@code toString} implementation as part of the generated equals algorithm.
```
Tillerino: Looks like you missed the second "equals" there

PR #1512, opened 2017-11-07
Wrong documentation variables fix.

RuslanPopenko:
I've fixed documentation for Cleanup annotation. Doc provided before wrong variable information (you closed String instead of Closeable).


rspilker:
Thanks.

PR #1517, opened 2017-11-15
getName call it self result in stackoverflow

skyupward:
I think line 12 is incorrect, call `this.getName()` in it self will cause stackoverflow.Change to return `this.name`.


rspilker:
Thanks.

PR #1549, opened 2018-01-05
ConfigurationKeys.ADD_LOMBOK_GENERATED_ANNOTATIONS defaults to true

chonton:
Now that jacoco 0.8.0 pays attention to @lombok.Generated, for ease of use change default of ConfigurationKeys.ADD_LOMBOK_GENERATED_ANNOTATIONS to true.

### DIFF
--- a/doc/changelog.markdown
+++ b/doc/changelog.markdown
@@ -4,6 +4,7 @@ Lombok Changelog
 ### v1.16.19 "Edgy Guinea Pig"
 * v1.16.18 is the latest stable release of Project Lombok.
 * PLATFORM: Possible support for jdk9 in the new IntelliJ, Netbeans and for Gradle.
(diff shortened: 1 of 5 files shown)


rspilker:
We will not turn this on by default (yet) since it will result in introducing lombok dependency in delomboked code.

It is okay to have users opt-in for that using the config file.

rspilker:
Thanks for the suggestion and your effort.

chonton:
Just to make sure I understand; annotations in the .class file are self-describing, so there is no runtime dependency on lombok.jar to delombok the code.  There will, however, be a compile-time dependency on lombok.jar in order to compile the delombok .java file.

Is it because of this compile-time dependency that you are unwilling to make adding @lombok.Generated the default?

Will this never be the default? Or, are you willing for the default to change on a major version update?

PR #1566, opened 2018-01-23
remove "@see RequiredArgsConstructor" from @Data javadoc

montassarelbehi:


rzwitserloot:
`@Data` implies `@RequiredArgsConstructor`; not sure why you think it should be removed from the javadoc?

PR #1622, opened 2018-03-21
fixed ExtensionMethod example to return T (generic type)

danieltaub96:
seem to forgot or remove :)


rzwitserloot:
Whoops. Good catch!

Issue #1617, opened 2018-03-19
Lombok in Java 9 makes Netbeans 9.0 go crazy


victorwss:
I already reported this here sometime ago. [Edit: It was in #1435. It also relates to #985]

Trying to use lombok in Java 9 within a Netbeans 9.0 project makes the IDE go crazy spamming a lot of `IllegalArgumentException: URI is not absolute` exceptions making it unusable.

The cause of the issue is that Netbeans 9.0 uses `javax.tools.JavaFileObject` instances and calls `toUri().toURL()` on them. But this happens in code specific to handle modules, so it will only trigger in Java 9. To be precise, this happens in the `org.netbeans.modules.java.source.parsing.PatchModuleFileManager.java` class.

The `toURL()` method throws that exception when it is called in a URI that is not constructed as an absolute URI. This was never a problem until now, but now it is.

The `JavaFileObject` javadocs do not say anything about requiring or not that the URIs should be absolute. Jan Lahoda from the Netbeans project replied:

> I believe the code in NetBeans expects that it will get JFOs which NetBeans
has constructed. Which apparently is not the case here. This is not a
completely wrong assumption: I am not aware about any API that would allow
to inject external JFOs into javac. I guess a solution here might be to
ignore unknown JFOs, which is what the standard file manager in javac is
doing, for better or worse.

After hacking around lombok, I determined that the offending `JavaFileObject` is an instance of `lombok.javac.apt.EmptyLombokFileObject`. The name which produces the bad URI is `lombok.dummy.ForceNewRound0` and it is produced in the `forceNewRound(String, JavacFiler)` method of the `lombok.javac.apt.LombokProcessor` class.

A simple fix would be to rewrite the `toUri()` method of `EmptyLombokFileObject`. Instead of this:

```java
	@Override public URI toUri() {
		return URI.create(name);
	}
```

Let's use this:

```java
	@Override public URI toUri() {
		return URI.create("file://" + name);
	}
```

After that, I:
1. Recompiled lombok with my changes.
2. Closed Netbeans and obliterated its cache in the disk.
3. Replaced the lombok.jar in my project with the one that I compiled.
4. Restarted Netbeans and reopened my project.
5. Profit!

**And, it worked like a charm. A simple one-line fix!**

Ok, to be fair, I had to twist some other things in order to make it able to compile in my Java 9 environment, but this is for some other issue. I don't think that this is really important here, but I should say it anyway in the case of somebody else failing to reproduce my fix.

[...]

PR #1626, opened 2018-03-24
Fix for #1617.

victorwss:
Signed-off-by: Victor Williams Stafusa da Silva <victorwssilva@gmail.com>


rzwitserloot:
The simplest things, huh? Hey, we've spent about 40 hours once in total to find a bug that took about 6 characters to fix.

Can you add your name to the AUTHORS file? Then I can accept the pullreq :)

victorwss:
@rzwitserloot Done.

PR #1627, opened 2018-03-24
Partially fix for "A save participant caused problems" with Eclipse

nuessgens:
This pull request is related to #879

The fix provided by  @mateusz-matela did solve this issue for most of the Eclipse users.

The patch was tested bei several people with different Eclipse versions (reaching from Luna to Oxygen.2)

However some users are still experiencing errors in combination of "Format edited lines", Lombok and Eclipse. Those seem to be related to a bunch of very specific combinations of save-actions (not yet quite sure if it is the same issue or even a new one). 

But however: this patch resolves (or at least improves) the issue for the majority of the eclipse users.


rzwitserloot:
Hey, wow, thanks for the extensive testing!

Can you add another commit on top of this one where you add your name to the AUTHORS file in the root? Then I can accept this pullreq :)

nuessgens:
The main work was done by @mateusz-matela.
I just applied the patch he provided and did the "git/ant-tooling".

I don't wnat to get the credit for someone else's efforts ;)

So should I add @mateusz-matela name, only? Or both of our names?

rspilker:
Well, it's your name on the commit so we definitely need yours. Feel free to have @mateusz-matela also add the name (but it'd have to be a commit by that github user) if you like :)

mateusz-matela:
I'm glad I could give a helping hand but I don't care about formalities that much.
So I don't mind if the pull request goes in without my name, good work @nuessgens! :)

Maaartinus:
@mateusz-matela I'm afraid, that's necessary for legal reasons - read the bottom of the  [authors](https://github.com/rzwitserloot/lombok/blob/master/AUTHORS) file.

mateusz-matela:
OK then, I've created the commit: https://github.com/nuessgens/lombok/commit/86df2c359eb9d712636a9e24c4f5036d94814bcb
@nuessgens can you cherry-pick that, or should I create a pull-request to this pull-request? :)

nuessgens:
@mateusz-matela: In order to avoid pull-request-ceptions I cherry-picked it ;)

PR #1664, opened 2018-04-18
fix typo in EclipseProductLocationProvider.java

kdaemonv:
change NotAnIdeLocationException to CorruptedIdeLocationException in javadoc for private method create0(String path)


rspilker:
thanks

PR #1675, opened 2018-04-27
Preserve default values on fields when using @Builder.Default

### LINKED ISSUE #1347: Builder.Default causes fields to be null when instantiating with new

shssoichiro:
Particularly, this allows `@NoArgsConstructor` to continue using default
values even when `@Builder.Default` is present on a field.

Closes #1347 

I did spend some time trying to get the tests to pass, but they seemed to be fundamentally broken. Particularly, I found a comment stating that the priority of `@Builder` is such that it must run before `@Value` so that `@Builder` can error if a field has both annotations. However, the presently existing test includes `@Value` and `@Builder`, and this is the use case that produced this bug in the first place, because the default value must be removed with `@Value` because `@Value` transforms fields to be final. That being said, I think there's more cleanup that needs to happen in `@Builder`, but I feel it is outside the scope of this issue.

### DIFF
--- a/src/core/lombok/eclipse/handlers/HandleBuilder.java
+++ b/src/core/lombok/eclipse/handlers/HandleBuilder.java
@@ -651,7 +651,6 @@ public class HandleBuilder extends EclipseAnnotationHandler<Builder> {
 		FieldDeclaration fd = (FieldDeclaration) fieldNode.get();
 		out.returnType = copyType(fd.type, source);
 		out.statements = new Statement[] {new ReturnStatement(fd.initialization, pS, pE)};
-		fd.initialization = null;
 		
 		out.traverse(new SetGeneratedByVisitor(source), ((TypeDeclaration) fieldNode.up().get()).scope);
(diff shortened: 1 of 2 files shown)


rzwitserloot:
We wiped out the initializer on purpose. This pull request is not solving the problem in a workable way.

Issue #1347 is on @rspilker 's radar and he has an idea on how to address it. It's high on his todo list. Hopefully #1347 will be fixed soon :)

Issue #1580, opened 2018-02-15
Lombok disables incremental compilation in gradle


pbi-qfs:
There is an approach for Gradle 4.6 to allow incremental compilation when annotation processors are used: https://github.com/gradle/gradle/issues/1320

Is this approach sufficient for Lombok? Does it require a lot of work on Lombok side?

The current state prohibits the use of lombok in big gradle projects.

oehme:
Lombok is a bit of a special case, as it uses internal APIs to change existing types. This is not officially allowed by the Processor API. Given how incremental annotation processing in Gradle currently works (to be released in 4.7), Lombok would work fine. But there is no guarantee that this would keep working in the future if we decide to take more control of the round environment for instance.

I guess this is the trade-off that Lombok has always taken - Rely on compiler internals in order to provide the most concise code.

oehme:
@nsiniakevich  That change would also be great but is unrelated to incremental compilation. Can you please open a separate issue?

oehme:
I've taken a closer look at Lombok. Its usage of [`Trees.instance(procEnv)`](https://github.com/rzwitserloot/lombok/blob/master/src/core/lombok/javac/apt/LombokProcessor.java#L88) would fail if used with Gradle's incremental compiler, because Gradle decorates the processing environment. Of course Lombok could add a workaround to extract the original environment if the devs are ready to take on that additional workaround.

nsiniakevich:
@rzwitserloot what do you this about this workaround?

nsiniakevich:
@oehme I've just wanted to open issue, but it's already here https://github.com/rzwitserloot/lombok/issues/1611 ;)

rzwitserloot:
@oehme I'm gonna try to interpret this one:

Okay, so, Lombok should call Trees.instance(procEnv) as normal, but, before proceeding, call .getClass().toString() to check if it so happens to be gradle's wrapper around the Trees class. If it is, we use reflection to fetch the javac Trees impl, and we carry on with whatever we were doing, using that reflection-obtained original environment.

Is that all it's gonna take? Because – that sounds rather simple to do. Can you point me at the wrapper code, or, better yet, tell me the fully qualified name of the wrapper + the method/field we can use to extract the original environment?

oehme:
[...]

PR #1680, opened 2018-05-03
Allow gradle incremental compiling with lombok

pbi-qfs:
Fixes issue #1580

Any recommended way to set up test for this use case?

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -306,6 +306,9 @@ the common tasks and can be called on to run the main aspects of all the sub-scr
 		<echo file="build/lombok/META-INF/services/javax.annotation.processing.Processor">lombok.launch.AnnotationProcessorHider$AnnotationProcessor
 lombok.launch.AnnotationProcessorHider$ClaimingProcessor</echo>
 		<echo file="build/lombok/META-INF/services/org.mapstruct.ap.spi.AstModifyingAnnotationProcessor">lombok.launch.AnnotationProcessorHider$AstModificationNotifier</echo>
+		<mkdir dir="build/lombok/META-INF/gradle" />
+		<echo file="build/lombok/META-INF/gradle/incremental.annotation.processors">lombok.launch.AnnotationProcessorHider$AnnotationProcessor,isolating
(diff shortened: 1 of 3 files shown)


oehme:
@pbi-qfs I recommend you write a few end-to-end tests using Gradle's [tooling API](https://docs.gradle.org/current/userguide/embedding.html#sec:embedding_quickstart). The test would

- set up a small example project
- invoke Gradle's `javaCompile` task
- do a small incremental change
- invoke the compile task again
- assert that only the expected files got recompiled

oehme:
Please also note that the registration mechanism will change in Gradle 4.8 to be more flexible and will be based on processor options instead of a META-INF file. The META-INF file will still be supported till 5.0, but no longer work after. See [this](https://github.com/gradle/gradle/pull/5255) pull request in Gradle that I'm going to merge soon.

pbi-qfs:
@oehme: Thanks for the suggestions on testing, gradle incremental is one side of the medal, though. I'm more concerned with the lombok side, so to ensure that all transformations still run through, and I'm not familiar on how to set up tests with different compiling environment inside the test structure of the lombok project.

404hasFound:
.

rzwitserloot:
Hey @pbi-qfs can you send one more pullreq where you add your name to the AUTHORS file? Thanks!

Excellent work, thanks for the PR!

oehme:
See my comment above, the registration mechanism will change in 4.8 and Gradle 5.0 will ignore the meta-inf file. Can you please adjust Lombok to use the new mechanism instead?

pbi-qfs:
@oehme When my Processor is derived from javax.annotation.processing.AbstractProcessor, it should be sufficient to add @SupportedOptions("org.gradle.annotation.processing.isolating") to the processor, isn't it (as long as no specific getSupportedOptions() implementation exists)?

oehme:
Yes, that'll work as well.

pbi-qfs:
It is the easier approach with less lines of code :)

oehme:
@rzwitserloot Will there be a release with this soon? Just let me know so we can promote it :)

PR #1698, opened 2018-05-22
Make the null check exception a little bit more explicit

marcust:
Currently it is "NullPointerException: someFieldName" and that is not very
telling as to the reasons this exception is thrown. It should be easily identifiable that this is due to a Lombok null check.


gluefactory:
Nice. The first few times I used Lombok, this tripped me up

rzwitserloot:
For future reference, please put _a bit_ more effort into it. I finished it here, for future reference:
a0b901c3c04d0377f495a9da853196a3b0433eae

PR #1699, opened 2018-05-24
Missing Author addition

pbi-qfs:
As requested, I added Pascal Bihler to author list and restored alphabetica…l order


rspilker:
Thanks.

PR #1711, opened 2018-06-01
@SuperBuilder: support for fields from parent classes

janrieke:
Based on my initial work on a solution for #853 (and peichhorn/lombok-pg#78) in pull request #1337, I now propose an improved approach.
Core improvement is that the new approach does not require casting any more. To achieve this, it uses generics on the generated builder classes.
This new approach is implemented as a separate experimental annotation called `@SuperBuilder`.

PS: The pull request is based upon the branch `builderParentClass`. However, I could also fix the conflicts with the `master` branch and create another pull request directly to `master`, if that's more convenient for you.

Example usage:

```java
@SuperBuilder
public class Parent {
	@Getter
	private String parentString;
}

@SuperBuilder
public class Child extends Parent {
	@Getter
	private String childString;
}

public class TestBuilder {
		Child child = Child.builder()
				.childString("childStringValue")
				.parentString("parentStringValue")
				.childString("secondChildStringValue")
				.parentString("secondParentStringValue")
				.build();
		assert("secondChildStringValue".equals(child.getChildString()));
		assert("secondParentStringValue".equals(child.getParentString()));
	}
}
```

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -6,6 +6,7 @@ DaveLaw <project.lombok@apconsult.de>
 Dave Brosius <dbrosius@mebigfatguy.com>
 Dawid Rusin <dawidrusin90@gmail.com>
 Enrique da Costa Cambio <enrique.dacostacambio@gmail.com>
+Jan Rieke <it@janrieke.de>
 Jappe van der Hel <jappe.vanderhel@gmail.com>
 Kevin Chirls <kchirls@users.noreply.github.com>
 Liu DongMiao <liudongmiao@gmail.com>
--- a/src/core/lombok/ConfigurationKeys.java
+++ b/src/core/lombok/ConfigurationKeys.java
@@ -508,6 +508,15 @@ public class ConfigurationKeys {
 	 */
 	public static final ConfigurationKey<FlagUsageType> WITHER_FLAG_USAGE = new ConfigurationKey<FlagUsageType>("lombok.wither.flagUsage", "Emit a warning or error if @Wither is used.") {};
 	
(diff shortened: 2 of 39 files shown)


janrieke:
I forgot that there may be abstract classes in the hierarchy. Quickly fixed that. But this should complete it. :)

corningsun:
Thanks, this is exactly what I was looking for.

rzwitserloot:
I have no idea why github didn't pick up on this, but as of today it's merged into master, and available via the edge release. Awesome, awesome work. I just had to fix up the style and website text a tiny bit, and add a changelog.

janrieke:
Great! If there are any bug reports on the feature, feel free to let me know so that I can have a look at it.

timothyBrake:
Awesome thanks!

Cepr0:
> [@SuperBuilder was introduced as experimental feature in lombok v1.18.2](https://projectlombok.org/features/experimental/SuperBuilder).

Does anybody know when this release will be available?..

putneyj:
This would be a great feature to be able to start testing. Any update on when 1.18.1/1.18.2 will be released?

hereyougo:
Not sure if it was already addressed but in 1.18.2 @Builder.Default used on the field in superclass fails with warning and compilation error. It would be really useful to have this option available.

janrieke:
No, that bug is not known so far. Please create a new ticket for that.

janrieke:
BTW: Release 1.18.2 is there with this feature.

hereyougo:
Thanks for the reply, I will create an issue.

Issue #1359, opened 2017-04-11
Eclipse Javaagent and mapstruct AST implementation


pgaschuetz:
Hi,

there seems to be a classloader issue when using lomboks javaagent with Eclipse and mapstructs annotation processing.
There has been some discussion including observations and the relevant stacktrace in mapstruct/mapstruct#1159

Without fully understanding the custom integration between lombok and mapstruct, it seems, that lomboks javaagent references a wrong classloader when working with mapstruct.

Can you shed some light into the issue?

Many thanks
Philipp

rspilker:
I'm wondering what changed. From the discussion on https://github.com/mapstruct/mapstruct/issues/510 I think it used to work. Or did nobody ever use mapstruct from Eclipse?

kanouken:
I got  the same issues,

polle:
I'm having the same issue after upgrading to Mapstruct 1.2 Beta3 and Lombok 1.16.16. Yes, it used to work with Mapstruct (and Lombok) in Eclipse.

filiphr:
@polle did this work for you with `1.2.0.Beta1`? Maybe we need to use a different classloader for the Service Loader.

Currently MapStruct does this

```
ServiceLoader.load(AstModifyingAnnotationProcessor.class, AnnotationProcessorContext.class.getClassLoader())
```

`AnnotationProcessorContext` is the class that collects the `AstModifyingAnnotationProcessor`(s) within MapStruct.

filiphr:
We (@pgaschuetz) tried with using the Current thread class loader and this error comes

```
org.mapstruct.ap.spi.AstModifyingAnnotationProcessor: Provider lombok.launch.AnnotationProcessorHider$AstModificationNotifier not a subtype
```

Can there be some problem in the agent? Do you perhaps shade the `AstModifyingAnnotationProcessor` in the agent?

polle:
@filiphr I did not try with 1.2.0.Beta1. I can try it out if you want?

filiphr:
@polle you said that it used to work before. Which versions of Lombok and MapStruct were you using? Are you using the correct version of the jar?

polle:
I'm currently using the following versions, which works fine with Eclipse:

Project dependencies:

    org.mapstruct:mapstruct-jdk8:1.0.0.Final
    org.mapstruct:mapstruct-processor:1.0.0.Final
    org.projectlombok:lombok:1.16.2

Eclipse Lombok installation:

[...]

PR #1718, opened 2018-06-06
Fix #1716: Update for Gradle incremental compilation API change in Gradle 4.8 and above

### LINKED ISSUE #1716: ClassCastException for Lombok 1.6.22 and Gradle 4.8
To reproduce:

Run `gradle init --type java-library` and modify generated project:

build.gradle
```
plugins { id 'java-library' }

dependencies {
    compileOnly 'org.projectlombok:lombok:1.16.22'
    annotationProcessor 'org.projectlombok:lombok:1.16.22'
}

compileJava {
    options.incremental = true
}

repositories {
    jcenter()
}
```

Foo.java
```
import lombok.Data;

@Data
public class Foo {
    private final boolean historical;
}
```

Running `gradle build` causes error:
```
> Task :compileJava FAILED
src/main/java/Library.java:4: warning: lombok.javac.apt.LombokProcessor could not be initialized. Lombok will not run during this compilation: java.lang.ClassCastException: org.gradle.api.internal.tasks.compile.processing.IncrementalFiler cannot be cast to com.sun.tools.javac.processing.JavacFiler
public class Library {
       ^
        at lombok.javac.apt.LombokProcessor.getJavacFiler(LombokProcessor.java:447)
        at lombok.javac.apt.LombokProcessor.init(LombokProcessor.java:90)
        at lombok.core.AnnotationProcessor$JavacDescriptor.want(AnnotationProcessor.java:112)
        at lombok.core.AnnotationProcessor.init(AnnotationProcessor.java:165)
        at lombok.launch.AnnotationProcessorHider$AnnotationProcessor.init(AnnotationProcessor.java:73)
        at org.gradle.api.internal.tasks.compile.processing.DelegatingProcessor.init(DelegatingProcessor.java:57)
        at org.gradle.api.internal.tasks.compile.processing.IsolatingProcessor.init(IsolatingProcessor.java:44)
```

Note that the error appears in file not annotated with Lombok annotations (created by `gradle init`).

Workaround - either:
1. Downgrade Lombok to 1.16.20 (implicitly disables incremental compilation); or
2. Downgrade Gradle to 4.7

mszabo-wikia:
Looks like incremental compilation API has changed in 4.8, causing Lombok init to fail:
```
warning: lombok.javac.apt.LombokProcessor could not be initialized. Lombok will not run during this compilation:
java.lang.ClassCastException: org.gradle.api.internal.tasks.compile.processing.IncrementalFiler cannot be cast to jdk.compiler/com.sun.tools.javac.processing.JavacFiler
```

This change should get Lombok working again on Gradle 4.8 and above; fixes #1716 

CC: @pbi-qfs

### DIFF
--- a/src/core/lombok/core/AnnotationProcessor.java
+++ b/src/core/lombok/core/AnnotationProcessor.java
@@ -49,6 +49,7 @@ import lombok.patcher.ClassRootFinder;
 
 @SupportedAnnotationTypes("*")
 public class AnnotationProcessor extends AbstractProcessor {
+
 	private static String trace(Throwable t) {
 		StringWriter w = new StringWriter();
 		t.printStackTrace(new PrintWriter(w, true));
@@ -71,18 +72,33 @@ public class AnnotationProcessor extends AbstractProcessor {
 	 * the delegate ProcessingEnvironment of the gradle wrapper is returned.
 	 */
 	public static ProcessingEnvironment getJavacProcessingEnvironment(ProcessingEnvironment procEnv, List<String> delayedWarnings) {
-		final Class<? extends ProcessingEnvironment> procEnvClass = procEnv.getClass();
-		if (procEnvClass.getName().equals("org.gradle.api.internal.tasks.compile.processing.IncrementalProcessingEnvironment")) {
+		ProcessingEnvironment javacProcEnv = tryRecursivelyObtainJavacProcessingEnvironment(procEnv);
+
+		if (javacProcEnv == null) {
+			delayedWarnings.add("Can't get the delegate of the gradle IncrementalProcessingEnvironment.");
+		}
+
+		return javacProcEnv;
+	}
+
+	private static ProcessingEnvironment tryRecursivelyObtainJavacProcessingEnvironment(ProcessingEnvironment procEnv) {
+		if (procEnv.getClass().getName().equals("com.sun.tools.javac.processing.JavacProcessingEnvironment")) {
+			return procEnv;
+		}
+
+		for (Class<?> procEnvClass = procEnv.getClass(); procEnvClass != null; procEnvClass = procEnvClass.getSuperclass()) {
 			try {
 				Field field = procEnvClass.getDeclaredField("delegate");
 				field.setAccessible(true);
 				Object delegate = field.get(procEnv);
-				return (ProcessingEnvironment) delegate;
+
+				return tryRecursivelyObtainJavacProcessingEnvironment((ProcessingEnvironment) delegate);
 			} catch (final Exception e) {
-				delayedWarnings.add("Can't get the delegate of the gradle IncrementalProcessingEnvironment: " + trace(e));
+				// no valid delegate, try superclass
 			}
 		}
-		return procEnv;
+
+		return null;
 	}
 
 	
@@ -96,7 +112,7 @@ public class AnnotationProcessor extends AbstractProcessor {
 		@Override boolean want(ProcessingEnvironment procEnv, List<String> delayedWarnings) {
 			ProcessingEnvironment javacProcEnv = getJavacProcessingEnvironment(procEnv, delayedWarnings);
 
-			if (!javacProcEnv.getClass().getName().equals("com.sun.tools.javac.processing.JavacProcessingEnvironment")) return false;
+			if (javacProcEnv == null) return false;
 
 			try {
 				ClassLoader classLoader = findAndPatchClassLoader(javacProcEnv);
--- a/src/core/lombok/javac/apt/LombokProcessor.java
+++ b/src/core/lombok/javac/apt/LombokProcessor.java
@@ -70,6 +70,9 @@ import com.sun.tools.javac.util.Context;
  */
 @SupportedAnnotationTypes("*")
 public class LombokProcessor extends AbstractProcessor {
+	private static final String GRADLE_INCREMENTAL_FILER_CLASS =
+			"org.gradle.api.internal.tasks.compile.processing.IncrementalFiler";
+
 	private ProcessingEnvironment processingEnv;
 	private JavacProcessingEnvironment javacProcessingEnv;
 	private JavacFiler javacFiler;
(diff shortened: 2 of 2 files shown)


mszabo-wikia:
Oh yeah that's a way more robust solution, thanks! I'll update this PR :)

pbi-qfs:
Thanks, @mszabo-wikia!

There is a copy of ﻿﻿getJavacProcessingEnvironment in AnnotationProcessor with different error reporting, which we should adapt as well.

victorwss:
Hi, yesterday, I was trying to fix this same problem, unaware that you already had a PR for that. After I coded it, I noticed you PR.

Anyway, you can see it [here](https://github.com/victorwss/lombok/blob/824969c25728eb143d3258252051a744058c5f2a/src/core/lombok/javac/apt/LombokProcessor.java#L429-L460).

This is my code so far:

```java
	private static final String GRADLE_INCREMENTAL_FILER_CLASS_NAME = "org.gradle.api.internal.tasks.compile.processing.IncrementalFiler";


	/**
	 * This class casts the given filer to a JavacFiler. In case of
	 * gradle incremental compilation, the delegate Filer of the gradle wrapper is returned.
	 */
	public JavacFiler getJavacFiler(Filer filer) {
		final Class<?> filerClass = filer.getClass();
		final Class<?> filerSuperClass = filerClass.getSuperclass();
		final Class<?> gradleFilerClass =
				filerClass.getName().equals(GRADLE_INCREMENTAL_FILER_CLASS_NAME) ? filerClass
				: filerSuperClass.getName().equals(GRADLE_INCREMENTAL_FILER_CLASS_NAME) ? filerSuperClass
				: null;
		if (gradleFilerClass != null) {
			try {
				Field field = gradleFilerClass.getDeclaredField("delegate");
				field.setAccessible(true);
				Object delegate = field.get(filer);
				return (JavacFiler) delegate;
			} catch (final Exception e) {
				e.printStackTrace();
				processingEnv.getMessager().printMessage(Kind.WARNING,
						"Can't get the delegate of the gradle IncrementalFiler. Lombok won't work.");
			}
		}
		if (!(filer instanceof JavacFiler)) {
			String message = "Don't know how to handle unknown filer " + filerClass.getName() + ". Lombok won't work.";
			processingEnv.getMessager().printMessage(Kind.WARNING, message);
			throw new RuntimeException(message);
		}
		return (JavacFiler) filer;
	}
```

mszabo-wikia:
Thanks @pbi-qfs I've updated the logic there as well. I assume the two implementations need to remain separate, right?

mszabo-wikia:
@victorwss Sorry about that. I did update the linked issue once I was ready with the PR so something must've gone wrong in the communication.

pbi-qfs:
@mszabo-wikia Great, lokks fine. It's kind of difficult to unify the two getters since they slightly differ in what they return - if you have a good way to do, step forward. It would ease future maintainability, since you wouldn't forget one place when fixing another.

Maybe extract a common "resolveDelegate(object,targetClassName)" method which we call in all thrww places?

mszabo-wikia:
hmm, sounds like a good idea. however I think such refactor should be a separate change so as to avoid making too many changes as part of this PR

pbi-qfs:
Ok, then from my Point of view, the code is good to merge. @rzwitserloot, what do you think?

rzwitserloot:
Looks good! Merged. @mszabo-wikia I noticed your name is not yet in our AUTHORS file. If you file another PR where you add your name between Mart and Mateusz, we can make it official. Thanks!

mszabo-wikia:
@rzwitserloot sweet, thanks, I'll update that :)

oehme:
@rzwitserloot Could you push a release with this change? I'd like to add a smoke test to Gradle so we can proactively reach out if we ever break stuff again. Also I want to add a performance test.

rzwitserloot:
@oehme The current edge release includes this. Do you need a full release that's also on maven-central, or will an edge release do? These can only be downloaded from https://projectlombok.org/download-edge

dmaslakov:
@rzwitserloot I have faced this issue and it stops me from using Gradle 4.9. I would prefer the full release as easier to use. Thanks in advance.

bwolff:
@rzwitserloot I agree with @dmaslakov, it would be very helpful if there was a minor bugfix release that makes it compatible with Gradle 4.8/4.9, if it's not too much of a hassle. Thank you very much!

oehme:
@rzwitserloot I personally could work with the edge release, but I think our users would very much prefer a full one.

rspilker:
We will do a release soon, but are waiting for enough people to have tried the [edge release](https://projectlombok.org/download-edge). Especially the changes for #1359 might break something.

aplatypus:
After updating the Lombok jar to: **`1.18.2`** (latest), I found a further problem if even for a file that doesn't use Lombok compiled under a `groovy`  folder tree name with **`Gradle v4.9`**.  

*  [6120 -- Lombok / Gradle 4.9 problem with Groovy folder names(s) ... ](https://github.com/gradle/gradle/issues/6120)

Still works fine with Gradle 4.8.1 and earlier.

berngp:
I am running into the same issue that @aplatypus is expressing even after upgrading ti Lombok v1.18.2. The issue presents itself explicitly when running `./gradlew compileGroovyTest`. The project doesn't have a `src/main/groovy` but we do leverage Spock for tests and we have plenty of `src/test/groovy` files.

oehme:
Please include the full stacktrace where Lombok is failing.

aplatypus:
@berngp  ... there might be a workaround, or like me, stay on Gradle 4.8.  The main reason I need a solution is for up-coming Gradle 5.0.

Anyway, in my case I found that having `groovy` in my source path was the problem.  After using Lombok v1.18.2.  If that's the same you could try to keep Groovy  files under `java/` path.  

- [#6120 -- **Lombok / Gradle 4.9 problem with Groovy folder names(s)**](https://github.com/gradle/gradle/issues/6120)

@oehme ... The reproducible example is in the body of  [#6120](https://github.com/gradle/gradle/issues/6120).

In fact any Existing Lombok annotated project you have lying around will produce the problem.  Just rename:

```
mv  src/main/java  src/main/groovy
```

Build with Gradle 4.8 or 4.8.1 -- **_works_**

Fails to build using Gradle 4.9.

oehme:
I could reproduce the problem, though next time I'd be really grateful if you could just attach a full example project instead of some text snippets. It saves a lot of time on my side, which I can use for fixing :) 

The problem affected all forking compilers. The [fix](https://github.com/gradle/gradle/pull/6183) will be in Gradle 4.10.

berngp:
Thanks @aplatypus, will stay with Gradle 4.8 for now. Thank as well for providing the details and the description on how to replicate it.

@oehme I apologize for not providing a full example. I appreciate that you looked into it, were able to replicate it and arrive to a conclusion.

Thanks again @aplatypus & @oehme!

PR #1760, opened 2018-07-08
[Performance] Replaced StringBuilder by simple char array in "hotspot" method

rPraml:
This PR replaces a StringBuilder by simple char array which adds a little bit of performance

This method was identified as a hotspot method by jprofiler with 214k calls in my test project.
this change improves the run time for 214k calls from 1165ms to 109ms

### DIFF
--- a/src/core/lombok/bytecode/ClassFileMetaData.java
+++ b/src/core/lombok/bytecode/ClassFileMetaData.java
@@ -116,23 +116,24 @@ public class ClassFileMetaData {
 		int end = pos + size;
 		
 		// the resulting string might be smaller
-		StringBuilder result = new StringBuilder(size);
+		char[] result = new char[size];
+		int length = 0;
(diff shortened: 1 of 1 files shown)


rPraml:
This PR is NOT included in #1771 as it the trade off between readability and performance gain is not so good. So if you do not want that, just decline it.

hc-codersatlas:
What version of Java is this run against? In Java 9+ StringBuilder might be represented as a byte[] or char[] depending on content. There are also lots of intrinsics in StringBuilder.

rPraml:
I have done the tests with java 8

rspilker:
@rPraml Can you please add you name to the AUTHORS file in the appropriate location so we are legally covered?

rPraml:
See #1787

rspilker:
Thanks.

PR #1761, opened 2018-07-08
[Performance] optimized the to/fromQualifiedName method

rPraml:
This PR adds some optimization in methods that are called very often (3.7M times in test project)
Reduces the total time of all calls from 4973ms to 1718ms


rPraml:
Thsi PR is included in #1771 - so I'm closing this

PR #1762, opened 2018-07-08
[Perfromance] using map.get instead of searching for entry

rPraml:
@rzwitserloot Please check this.
I don't understand why iterating over the keys instead of using map.get (maybe due refactoring)
map.get should be faster. 
(before: 17999ms, after 1466ms - 3,7M invocations)


rPraml:
This PR is included in #1771 - so I'm closing this

PR #1763, opened 2018-07-08
[Performance] Do not use reflection to check if annotation is present

rPraml:
Cache the DeferUntilPostDiet flag instead of using reflection to check


rPraml:
This PR is included in #1771 - so I' closing this

PR #1765, opened 2018-07-08
[Performance] use declared ast Type instead of generic one 

rPraml:
There are many calls like `ast.traverseChildren` - due type erasue, ast was internally stored as object, which leads to a "checkcast" before each "invokevirtual" in the byte code.

This PR and #1763 reduces the total time spent in the `traverse` method (30M calls) from 52276ms to 51246ms.


rPraml:
This PR is NOT included in #1771 as it the trade off between readability and performance gain is not so good. So if you do not want that, just decline it.

PR #1766, opened 2018-07-08
experimental: skip lombok if no import found

rPraml:
This PR will try to skip lombok processing completely when `-Dlombok.checkImports` is set and no lombok import was found. This will break full qualified annotations, so this is experimental (but gives the most performance boost)


rPraml:
This change gives the best performance improvement on non-lombok projects but will break many things. So I'm closing this

Maaartinus:
@rPraml I guess, scanning the source for `"lombok."` is pretty fast (as you mentioned elsewhere) and very profitable for non-lombok projects, while the overhead for lombok projects is negligible.

Should the overhead for lombok projects matter, then you could either

- use a global boolean which gets set when the string gets found and switches to non-scanning
- or extract the information what lombok annotations get used, possibly eliminating parts of further processing (this idea is based on another PR of yours)

PR #1768, opened 2018-07-08
[Performance] Optimize the ShadowClassLoader when determining isPartOfShadow

rPraml:
This PR tries to restore the main performance drop on 1.18.1

**issue:**
I have a multi module maven project with 20 modules using ecj. 
With 1.18.1 the build time increased ~7 seconds compared to 1.18.0

The bottleneck was, that each compile task per maven module scans the complete rt.jar if "java/lang/Object" is extended. I also noticed, that every task creates a new ClassLoader (which is unneccessary IMHO)


rPraml:
This PR can be closed if #1771 will be merged

PR #1771, opened 2018-07-09
Reduced the AST traverse calls to improve ECJ performance

rPraml:
This PR brings several optimizations for the ECJ, especially when using it in a multi module maven build (as I do) - also the eclipse performance should be better, especially on modules that do not use lombok.

**So: some numbers:**
 1. Single maven project, not using lombok ~ 1200 class files:

|Type of build |Total maven build time | Est. compile time| Note |
|--------------|------------------------|------------------|------|
| No lombok   | ~8.5 sec                      |     5.0 sec  | this is the 'reference' build time, as this project can be built without lombok|
| 1.18.0 - original | ~12.5 sec | 9.0 sec | just adding lombok (as eclipse does by default for all projects) will nearly double the compile time, because lombok tries to enhance EVERY class. I tried ti scan for "interesting imports" which nearly restored the original performance, but the discussion showed, that this approach cannot detect FQNs. Maybe it is possible to add a switch to disable lombok for certain projects|
| 1.18.1 - original | ~ 13 sec | 9.5 sec | With 1.18.1 there was a change in the class loader that tries to detect if a type is part of the shadow. This scans every jar, also the RT jar which takes 300-500ms milliseconds. I tried to optimize this for "java" and "sun" packages to skip at least the rt.jar. In the multi module build, these 300ms were added to each module (20 Modules * 300ms = 6sec)
| 1.18.1 - modified | ~ 10 sec | 6.5 sec | So the patched version adds only 1.5 sec overhead instead of 4 sec to a project where no (or only a few classes) are using lombok annotations. This is done by tracking the next interesting priority. The old version performed 40M traverse calls, the new one has only 3,7M traverse calls on the same project. There are also some minor other changes like using char arrays instead of StringBuilders, which saves also some milliseconds|

2. Multi Module project 20 modules ~ 4000 class files

|Type of build |Total maven build time | Est. compile time| Note |
|--------------|------------------------|------------------|------|
| No lombok   |                                   |                            | Project requires lombok, so it is not testable |
| 1.18.0 - original | ~39 sec | 21 sec | This is the 'reference' |
| 1.18.1 - original | ~46 sec | 28 sec | the change in the class loader adds ~300ms to each module. See above |
| 1.18.1 - patched | ~46 sec | 17 sec | this is a noticeable performance improvement I would say. As most of the classes uses only @Getter and @Setter, most of the priorities can be skipped. The traverse calls were reduced from 31M to 4.5M. You can also see that lombok is no longer the time consumer in the profiler results |

**result on 1.18.1 - original**
grafik

**result on 1.18.1 - with this PR**
grafik

I will close the other PRs and I would be happy if this can be merged in

BTW: I had a problem to run the integration tests, because they do not run on java 1.8 or 10 (and java 9 is no longer officially downloadable)

### DIFF
--- a/src/core/lombok/core/AST.java
+++ b/src/core/lombok/core/AST.java
@@ -216,19 +216,19 @@ public abstract class AST<A extends AST<A, L, N>, L extends LombokNode<A, L, N>,
 		}
 	}
 	
-	private static final ConcurrentMap<Class<?>, Collection<FieldAccess>> fieldsOfASTClasses = new ConcurrentHashMap<Class<?>, Collection<FieldAccess>>();
+	private static final ConcurrentMap<Class<?>, FieldAccess[]> fieldsOfASTClasses = new ConcurrentHashMap<Class<?>, FieldAccess[]>();
 	
 	/** Returns FieldAccess objects for the stated class. Each field that contains objects of the kind returned by
 	 * {@link #getStatementTypes()}, either directly or inside of an array or java.util.collection (or array-of-arrays,
 	 * or collection-of-collections, et cetera), is returned.
 	 */
-	protected Collection<FieldAccess> fieldsOf(Class<?> c) {
-		Collection<FieldAccess> fields = fieldsOfASTClasses.get(c);
+	protected FieldAccess[] fieldsOf(Class<?> c) {
+		FieldAccess[] fields = fieldsOfASTClasses.get(c);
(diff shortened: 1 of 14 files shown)


rspilker:
Thanks for the research.

Can you please also add your name to the AUTHORS file in the appropriate location? That way we are legally covered.

rPraml:
See #1787

PR #1809, opened 2018-08-08
java.util.logging.Logger does not have an error method.

JoachimVandersmissen:
I replaced the `error` method invocation in the `@Log` examples with `severe`, because `java.util.logging.Logger` does not have an `error` method (as per https://docs.oracle.com/javase/10/docs/api/java/util/logging/Logger.html)


rspilker:
Thanks

PR #1818, opened 2018-08-16
SuperBuilder uses Builder's behavior for default

janrieke:
With this change, `@SuperBuilder` uses the same behavior as `@Builder` for fields annotated with `@Builder.Default`. In particular, it generates a `$default$field()` method that is called if there was no explicit set of the respective field. The initializing expression is moved from the field into this method.
See also issue #1347.


alturkovic:
Totally agree, I just thought you should know that `@Builder.Default` behavior is flawed in case someone reports a bug for `@SuperBuilder` since you said you rely on that behavior. Thanks for your input.

PR #1780, opened 2018-07-17
Allow jitpack to build lombok using JDK 9+

mmoayyed:
## Problem

Let lombok build on jitpack.io.

From:
https://jitpack.io/docs/BUILDING/#build-customization

> You can create a jitpack.yml file in the root of your repository and override the build commands:
>
> jdk:
>   -openjdk9

## Current

A sample failing build:
https://jitpack.io/com/github/rzwitserloot/lombok/37434ba0cb/build.log

With current failure:

```
-ensureJdk9:

BUILD FAILED
/home/jitpack/build/build.xml:176: To compile lombok, you need JDK9 or higher; lombok requires this version because it's rather difficult 
to produce lombok builds that are compatible on JDK9 without at least building with JDK9. Sorry about that.
```

## Why

This, once working, should allow access to SNAPSHOT builds of lombok whereby developers can quickly test changes and provided feedback. (Case in point: https://github.com/rzwitserloot/lombok/pull/1718)

## Caveats

I ended up switching the java version insufficiency check to use `java.version` rather than `ant.java.version`. This was done to get around jitpack's behavior:

```
-ensureJdk9:
     [echo] Java version: 1.8
     [echo] Java version: 9.0.4
     [echo] Java version insufficient: true
```

Using the above changes, I was able to successfully build lombok. [Here is a build log](https://jitpack.io/com/github/mmoayyed/lombok/-v1.18.0-gbff1237-99/build.log) that shows the build is successful.

# Usage

See: https://jitpack.io/#rzwitserloot/lombok/-SNAPSHOT

## Maven

```
<repositories>
		<repository>
		    <id>jitpack.io</id>
		    <url>https://jitpack.io</url>
		</repository>
</repositories>
<dependency>
	    <groupId>com.github.rzwitserloot</groupId>
	    <artifactId>lombok</artifactId>
	    <version>-SNAPSHOT</version>
</dependency>
```

## Gradle

```
allprojects {
		repositories {
			...
			maven { url 'https://jitpack.io' }
		}
}
dependencies {
	        implementation 'com.github.rzwitserloot:lombok:-SNAPSHOT'
}
```

PR #1827, opened 2018-08-22
Allow jitpack to use JDK 10 when building lombok

mmoayyed:
Subsequent to #1780 

Sample build outcome:
https://jitpack.io/com/github/mmoayyed/lombok/jitpack11-v1.18.0-gcf9f25a-152/build.log

### DIFF
--- a/jitpack.yml
+++ b/jitpack.yml
@@ -1,4 +1,6 @@
 jdk:
-- openjdk9
-before_install:
-- export JAVA_HOME=/usr/lib/jvm/java-9-oracle
+- openjdk10
+# before_install:
+# - export JAVA_HOME=/usr/lib/jvm/java-10-oracle
+# - echo "******Ant Version********"
+# - ant -version


rspilker:
The last line is `No build artifacts found`. But I expect that we can instruct jitpack to rin `ant install-maven` instead ant that that would solve the problem. Is that correct?

mmoayyed:
We surely could yes, but that bit is very strange. It looks to me as if, once the artifact is built, jitpack is trying to get it installed in _their own_ local repository and is looking for a pom.xml to do that. 

I'll tweak the build command to see if it can use the new ant command.

rspilker:
I merged it to get the jdk10 in.

rspilker:
FYI, jitpack is too unreliable and slow. In the meantime, and as a result of the lessons learned from trying to get jitpack to work, we have decided to host the snapshot releases on our own website.

Thanks for helping us get to that stage.

rspilker:
You may need a reload to get the latest css, fixing alignment problems of the maven/gradle examples.

mmoayyed:
That is excellent. Thank you for taking care of this.

PR #1828, opened 2018-08-22
Add ant task to install lombok jar into local maven repository

mmoayyed:


rspilker:
Shouldn't it contain the text `-SNAPSHOT`?

mmoayyed:
Good point. I have corrected that now.

PR #1870, opened 2018-09-24
toBuilder support for SuperBuilder

janrieke:
This PR adds `toBuilder` support for `@SuperBuilder`.
The approach is implemented as outlined in the [Google Groups thread](https://groups.google.com/forum/#!topic/project-lombok/VcgBsprEe5k), with the minor change that the `$fillValuesFromInto()` method has moved to the abstract builder class.


careylam:
I would recommend to generate a new builder method that takes in the object itself. e.g.

...
    public static DomainBuilder builder(DomainObject prototype) {
        return new Builder(prototype);
    }
...

PR #1898, opened 2018-10-12
toBuilder: wildcards and singular fix

### LINKED ISSUE #1885: Current edge 1.18.3 breaks toBuilder on Singular fields with wildcards
The following code compiles with 1.18.2, but not with the current 1.18.3 snapshot (downloaded today) today), using Java 8.

    @Builder(toBuilder = true)
    public class ExampleToBuilder {
        @Singular
        private List<?> objects;
    }

Eclipse says

    Wildcard is not allowed at this location

javac says:

    method emptyList in class java.util.Collections cannot be applied to given types;
    [ERROR]   required: no arguments
    [ERROR]   found: no arguments
    [ERROR]   reason: explicit type argument ? does not conform to declared bound(s) java.lang.Object"

I think 6a42bcc and af55bd0 broke it, but I have not checked that.

janrieke:
Fixes #1885 by replacing the wildcard with the lower bound type.
Furthermore, this fixes missing type arguments when using `@ObtainVia` with a static method.


janrieke:
Although the conflicts do not look too bad, I cannot promise anything at the moment.
I'd say: Don't let this delay a release.

Issue #1564, opened 2018-01-18
Custom @NotNull annotation makes Lombok do things


schuettec:
Consider the following code:
```
@Data
@AllArgsConstructor
@NoArgsConstructor
@Builder
public class Thing {
  @interface NonNull {}

  @NonNull
  private String a;

  public static void main(String[] args) {
    Thing.builder()
        .build();
  }
}
```

This ends up with a NPE:
```
Exception in thread "main" java.lang.NullPointerException: a
	at com.remondis.renew.validation.test.Thing.<init>(Thing.java:9)
	at com.remondis.renew.validation.test.Thing$ThingBuilder.build(Thing.java:11)
	at com.remondis.renew.validation.test.Thing.main(Thing.java:22)
```

I would not expect that, because the custom `NonNull` annotation is not part of lombok and should do very different things!

mithuns:
`@NonNull` is part of lombok [@NonNull](https://projectlombok.org/features/NonNull)
Are you saying, this NonNull , you are using is something else ? Can you post the full import statement to show which NonNull is this ?

schuettec:
@NonNull is implemented locally in the class to demonstrate that this
annonation is not part of lombok.

Am 19.01.2018 21:26 schrieb "Mithun" <notifications@github.com>:

@NonNull is part of lombok @NonNull
<https://projectlombok.org/features/NonNull>
Are you saying, this NonNull , you are using is something else ? Can you
post the full import statement to show which NonNull is this ?

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/rzwitserloot/lombok/issues/1564#issuecomment-359079795>,
or mute the thread

mithuns:
Right, sorry I think I misread somewhere and then realized, my question was unnecessary.

schuettec:
No problem ;-)

Am 19.01.2018 22:59 schrieb "Mithun" <notifications@github.com>:

vipulbansal91:
According to me, this is probably a design/language constraint. 

Related code links:
https://github.com/rzwitserloot/lombok/blob/master/src/core/lombok/core/handlers/HandlerUtil.java#L221
https://github.com/rzwitserloot/lombok/blob/master/src/core/lombok/javac/handlers/HandleConstructor.java#L287-L297

The same has been called out in the documentation as well.
https://projectlombok.org/features/NonNull reads :  
[...]

Issue #1585, opened 2018-02-19
Error with combination of custom @NonNull constraint and lazy getter


grv87:
This code:
```
import org.checkerframework.checker.nullness.qual.NonNull;
import lombok.Getter;

...

  @Getter(lazy = true)
  private final @NonNull Writable changeLog = generateChangeLog();

  private Writable generateChangeLog() { ... }
```
generates compilation error:
`scoping construct cannot be annotated with type-use annotation`

PR #1917, opened 2018-10-23
Added Spring NonNull and Nullable annotations

kkocel:
Added Spring NonNull and Nullable annotations

### DIFF
--- a/src/core/lombok/core/handlers/HandlerUtil.java
+++ b/src/core/lombok/core/handlers/HandlerUtil.java
@@ -85,6 +85,7 @@ public class HandlerUtil {
 			"org.jetbrains.annotations.NotNull",
 			"android.support.annotation.NonNull",
 			"org.eclipse.jdt.annotation.NonNull",
+			"org.springframework.lang.NonNull"
 		}));
 		BASE_COPYABLE_ANNOTATIONS = Collections.unmodifiableList(Arrays.asList(new String[] {
 			"lombok.NonNull",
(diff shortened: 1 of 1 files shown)


kkocel:
@jmax01 could you please elaborate more on this?  Those annotations need to be present in your java code in order to be copied. How it can break the serialization?

jmax01:
We have been doing two things:
* Retrofitting code with non-Lombok `NonNull/Nullable` annotations to support static analysis tooling
* Adding Lombok to legacy projects that already have `NonNull/Nullable` annotations 

We were unaware that Lombok was adding null checks to the code without explicitly using    `lombok.NonNull`.  

This is unacceptable to us as it changes the performance characteristics of the code and breaks code that could safely ignore the null safety checks such as during certain serialization scenarios.

In my mind an opt-in/opt-out mechanism is required.  If we could configure Lombok to ignore certain annotations this would be ideal. 

This opt-in/opt-out mechanism would also fix https://github.com/rzwitserloot/lombok/issues/1564 and possibly https://github.com/rzwitserloot/lombok/issues/1585.

kkocel:
I think you referencing separate concerns - my pull request was about adding support for spring annotations to be copied over to builder. Part of documentation about this feature:
> BREAKING CHANGE: Lombok will now always copy specific annotations around (from field to getter, from field to builder 'setter', etcetera): A specific curated list of known annotations where that is the right thing to do (generally, @NonNull style annotations from various libraries), as well as any annotations you explicitly list in the lombok.copyableAnnotations config key in your lombok.config file. Also, lombok is more consistent about copying these annotations. 

Whereas you are referencing to functionality of lombok that generates null checks for you if it finds any annotatnion named `NonNull`, described here: https://projectlombok.org/features/NonNull

I am not aware of any mechanism to disable null check generation ( full list of options is available at ConfigurationKeys class)  but you can always submit a PR with such functionality, as I did :)

PR #2004, opened 2019-01-03
Copy Javadoc to builder setters

### LINKED ISSUE #1033: @Builder and javadoc
Hi guys,

Maybe I missed it, but if I create a class with an @Builder annotation and I add javadoc to the fields, I find that this javadoc is not "transported" onto the builder methods. It would be very helpful if this were possible. 

If it is possible, I'm probably doing something wrong. How can I get it right?

My code looks like this:

``` java
@AllArgsConstructor(access = AccessLevel.PRIVATE)
@Builder
@ToString
public class SomeClass {
    /** The name of the item */
    private final String name;
}
```

Cheers!

### LINKED ISSUE #1445: Javadoc is not generated for the code generated by @Builder annotation
lombok1.6.8 
Java8
IDE:
Version: Neon.2 Release (4.6.2)
Build id: 20161208-0600

emlun:
Fixes #1033. Fixes #1445.

This causes Javadoc to be copied to `@Builder` setters using SETTER copy mode. See the added tests for an example.

Notes:
- The change is currently only made in the `makeSimpleSetterMethodForBuilder` method, so I don't think this will apply to "singular" setters or other special cases - I haven't quite been able to grok how the `else` branch in `makeSetterMethodsForBuilder` method works.
- It looked like the ECJ variant strips Javadoc altogether, so I didn't apply the same change there.
- The tests currently do not cover interactions such as between `@Builder` and `@Value`/`@Getter`. I'll try to add some for that.

### DIFF
--- a/src/core/lombok/javac/JavacAugments.java
+++ b/src/core/lombok/javac/JavacAugments.java
@@ -21,9 +21,11 @@
  */
 package lombok.javac;
 
+import java.util.Set;
 import lombok.core.FieldAugment;
 
 import com.sun.tools.javac.tree.JCTree;
+import lombok.javac.handlers.JavacHandlerUtil.CopyJavadoc;
 
 public final class JavacAugments {
 	private JavacAugments() {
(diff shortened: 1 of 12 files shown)


emlun:
Turns out this doesn't play nicely with `@Setter` - `@param` tags get moved to the builder setter and are omitted from the instance setter. I'll see if I can fix that...

emlun:
Forum discussion: https://groups.google.com/forum/#!topic/project-lombok/aZx5wJp891k

rzwitserloot:
In commit ccfab5ebffbd14df59689b23ff6f3df52890dca8 I've added a cleanup tasks system for the javac run of lombok. This should make it _a lot_ simpler to fix this problem properly. I've also fixed the setter vs. wither conflict and written a test to show the fix is working, as proof of concept.

Would it be okay if you merge/rewrite your PR to use it? I'm pretty sure it's a breeze with this update, just, swamped with work and it was a late evening. Muchos gracias and thanks for putting in the effort to fix this oldie :)

Also, can you add your name to the CONTRIBUTORS file in the root of the project?

emlun:
Cool, thanks! It indeed looks like this will be a lot simpler with that new infrastructure. I'll close this and open a new PR.

PR #2018, opened 2019-01-13
Reduce code duplication in singularizers

emlun:
As I was working on #2004, I noticed this part of the code looked like it could need some love. The main changes are

- Split the `generateMethods` methods up into parts so that identical parts can be reused between them.
- Condense the "use Guava instead" logic into a single check which replaces the whole singularizer instance once instead of sprinking `if (useGuavaInstead())` checks all over the place.

When I started, I expected the end result would be more substantial than a net -42 LOC change (-69 if ignoring whitespace), so I was a bit underwhelmed by the reduction volume and I'm not sure it's worth the rather extensive code changes. That said, the changes are organised into small commits such that it's easy to manually verify the correctness of each step, and I do believe that splitting the code into smaller functions makes it easier to work with.

Anyway, here it is. This was my attempt to make this code region easier to read and understand. You're welcome to close this or chery-pick individual changes as you like, and I welcome any critique on how this could be done better.

These changes are pure refactorizations, with the exception of commits a633e39 through 92fc11f which do change internal behaviour but should not change externally observable behaviour.

### DIFF
--- a/src/core/lombok/javac/handlers/HandleBuilder.java
+++ b/src/core/lombok/javac/handlers/HandleBuilder.java
@@ -788,7 +788,7 @@ public class HandleBuilder extends JavacAnnotationHandler<Builder> {
 			name = type.toString();
 			
 			String targetFqn = JavacSingularsRecipes.get().toQualified(name);
-			JavacSingularizer singularizer = JavacSingularsRecipes.get().getSingularizer(targetFqn);
+			JavacSingularizer singularizer = JavacSingularsRecipes.get().getSingularizer(targetFqn, node);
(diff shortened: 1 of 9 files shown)


rspilker:
Thanks, I haven't had the time to look into the list, allowing jdk9+ to compile a multi-module project had the higher priority. Hope to look into this soon (tm).

emlun:
No worries, no hurries. :slightly_smiling_face:

rzwitserloot:
Is there like an oscar for PRs? I vote for this one.

emlun:
Hahah, thanks! I'm glad you like it.

PR #2029, opened 2019-01-22
Augment NONNULL_ANNOTATIONS with more aliases for lombok.NonNull.

mernst:
Also augment BASE_COPYABLE_ANNOTATIONS with more type annotations from the
Checker Framework.


rspilker:
I've looked at the list. My only concern is `javax.validation.constraints.NotNull`

If I understand correctly, it is perfectly valid for a field annotated with `javax.validation.constraints.NotNull` to contain a null value UNTIL it is persisted. That was one of the reasons why it was missing from the list in the first place.

I don't have enough data to determine if this distinction would lead to problems in practice. Do our users actually create those object with null values and fill them in later. What do we expect from static analyzers here?

@mernst Do you have an opinion about this?

mernst:
Your understanding is probably better than mine.
I was going not by any personal experience but by the Javadoc:
https://docs.oracle.com/javaee/7/api/javax/validation/constraints/NotNull.html
https://javaee.github.io/javaee-spec/javadocs/javax/validation/constraints/NotNull.html
which does not mention that null is ever permitted, or facts about the lifecycle and persistence.

I have removed `javax.validation.constraints.NotNull` from the list, and added a comment so that other people know why it is not there.

Thanks for catching that!

Issue #1186, opened 2016-09-08
Compile error when static-importing a field from @UtilityClass


ghost:
How to reproduce:
1. Define a public constant in a @UtilityClass (AnnotatedUtilityClass.java in example)
2. Static import this constant in another class (StaticImportingClass.java in example)
3. Run 'mvn compile'

Attached minimal example:
[lombok-utility-class-issue.zip](https://github.com/rzwitserloot/lombok/files/461451/lombok-utility-class-issue.zip)

hobbypunk90:
I've the same problem with a static import of the whole utility class.

askoning:
This also occurs without Lombok. It's not possible to static import within the same package.
```$ cat Main.java   
import static Utility.*;

public class Main {
        Main() {
                String s = FOO;
        }
}```

```$ cat Utility.java 
public final class Utility {
        public static final String FOO = "Foo";
}```

```$ javac *.java
Main.java:1: error: cannot find symbol
import static Utility.*;
              ^
  symbol: class Utility
Main.java:5: error: cannot find symbol
                String s = FOO;
                           ^
  symbol:   variable FOO
  location: class Main
2 errors```

Static importing a field from a @UtilityClass in another package should work just fine - at least, I can't reproduce the issue with the latest release.

oliverboehme:
> It's not possible to static import within the same package.
That is new to me.

Also I'm quite sure that I first encountered this problem when importing from a different package.

mmlinford:
I don't have a lot of time to investigate right now, but I can confirm that this is still happening in lombok 1.18.0.

class MyClass is in package a.b.c
class MyUtilityClass is in package a.b.c.d

MyClass.java does `import static a.b.c.d.MyUtilityClass.someMethod`. When I try to compile, seemingly *all* lombok annotations failed to process. When I switch to just calling `MyUtilityClass.someMethod()` without the static import, things work fine.

However, if I go to MyUtilityClass and give the method a `static` modifier by hand, it compiles fine again.

I'll try to post a gist or something when I have time. Thanks.

rzwitserloot:
Duplicate of #2044

Issue #884, opened 2015-07-23
static import of method in @UtilityClass breaks compilation


almson:
The following Maven project demonstrates the problem on Netbeans 8.0.2 running on Oracle Java 8u51 x86_64 on Windows 8.1 (Ubuntu 14.04 x86_64 is seemingly not affected with the same Netbeans/Java platform): 

```
package com.mycompany.lombokutilityclassbug;

@lombok.experimental.UtilityClass
public class MyUtilityClass {

    void foo() { System.out.println ("hi"); }
}
```

.

```
package com.mycompany.lombokutilityclassbug;

import static com.mycompany.lombokutilityclassbug.MyUtilityClass.foo;

public class Main { }
```

.

```
<?xml version="1.0" encoding="UTF-8"?>
<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
    <modelVersion>4.0.0</modelVersion>
    <groupId>com.mycompany</groupId>
    <artifactId>LombokUtilityClassBug</artifactId>
    <version>1.0-SNAPSHOT</version>
    <packaging>jar</packaging>
    <dependencies>
        <dependency>
            <groupId>org.projectlombok</groupId>
            <artifactId>lombok</artifactId>
            <version>1.16.4</version>
        </dependency>
    </dependencies>
    <properties>
        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
        <maven.compiler.source>1.7</maven.compiler.source>
        <maven.compiler.target>1.7</maven.compiler.target>
    </properties>
</project>
```

kazuki-ma:
`static void foo()`?

almson:
@kazuki-ma `@UtilityClass` adds the static modifier

btiernay:
I hit this issue a lot and seems to break compilation in a way that is very hard to diagnose since it creates many noisy downstream errors.

mjgp2:
This is also the case for `@FieldDefaults(level=PUBLIC)` when trying to import a static field

piotrturski:
i've just hit this issue with lombok 1.16.6 and java 1.8
any workarounds or `import static` simply doesn't work?

Shredder121:
> any workarounds or import static simply doesn't work?

I have an interesting find.
[...]

PR #2039, opened 2019-02-15
Add documentation to UtilityClass to warn users away from static imports

jasonab:
RE: #1382, #1186, #884: add documentation to @UtilityClass to warn users not to use static imports with this annotation.


jasonab:
Now relevant to #2044

Issue #1889, opened 2018-10-09
Scaling issues on high DPI displays


peterstritzinger:
My Lenovo Yoga Pro has a high resolution of 3200x1800 with scaling at 300% => the Lombok UI useless.
The problem is that i can either resize nore maximize the window. See following screenshot.

![delme_lombok](https://user-images.githubusercontent.com/26869780/46671440-12484980-cbd5-11e8-939e-1dda52fbd0d0.PNG)

My Story: As a lombok UI user I would like to be able to maximize or resize the window.

rspilker:
Duplicate, solved by recent java versions.
[Why we closed this issue](https://github.com/rzwitserloot/lombok/wiki/ISSUES:-Why-we-close-issues-without-a-lot-of-explanation.)

peterstritzinger:
Hi,
sorry to say that but "solved by recent java versions" is in my eyes not a satisfying anwser.
Could you please provide more information.
By "duplicate" you probably mean https://github.com/rzwitserloot/lombok/issues/1355
I can not find a solution there either.

Thx

anandchakru:
Making it a fixed size window is the root cause.
Created a [PR#2057](https://github.com/rzwitserloot/lombok/pull/2057) for a fix

PR #2057, opened 2019-03-06
#1889 fix

anandchakru:
https://github.com/rzwitserloot/lombok/issues/1889


rspilker:
We'll need to test this manually before we're merging this one.

rzwitserloot:
I've tried to reproduce this on a hi-dpi mac using JDK1.7 and can't.

When I apply this fix, the window is resizable, which is mostly bad; the GUI layout does not scale up so it just looks stupid. Worse, if I make the window _smaller_, the GUI breaks completely.

If I then make the window larger again, the GUI stays broken, which to me strongly suggests this fix doesn't in fact fix the problem in the first place. I can't confirm this theory.. because I can't reproduce it.

PR #2109, opened 2019-04-24
improvements type inference of type intersections (affects ValLambda.java, ValWeirdTypes.java tests)

bulgakovalexander:


rzwitserloot:
Unfortunately, looks like I have to shoot this one down. I don't know how you do it, but you manage to make javac (with lombok with your patch) correctly untangle this marvellous line of code:

```java
lombok.val foo = (System.currentTimeMillis() > 0) ? (Runnable)()-> {} : System.out::println;
```

If I try to compile that with javac with lombok with your patch in it, it works. But if I try to compile the same thing with ecj, it won't (it can't target-type `System.out::println` into a functional interface). This compiles fine on both ecj and javac without lombok:

```java
Runnable foo = (System.currentTimeMillis() > 0) ? (Runnable)()-> {} : System.out::println;
```

and this won't compile on either:

```java
Object foo = (System.currentTimeMillis() > 0) ? (Runnable)()-> {} : System.out::println;
```

adding to the impressiveness of the patch :)

Nevertheless, if we can't make ecj and javac do the same thing here, I prefer to just call it a day and go with j.l.Object for this stuff. Not great either, but at least consistent.

rzwitserloot:
This is some seriously impressive voodoo magic, Alex! Will be in the next release :)

rzwitserloot:
See also 2eab72cc9ffbd840b951900818727ddb8d137b8e

Issue #2100, opened 2019-04-18
[FEATURE]  Value annotation should use already defined private constructor


DevQuanti:
**Describe the feature**
When I use the @Value annotation with staticConstructor = "of" and a private Constructor is already defined, Lombok should use the private Constructor to generate a static of Constructor.

Lomboked version

```
@Value(staticConstructor = "of")
public final class SampleDocument {

    private final String value;

    private SampleDocument(final String value) {
        assertThat(value).isNotEmpty();
        this.value = value;
    }

}
```

Java version
```
public final class SampleDocument {

    private final String value;

    private SampleDocument(final String value) {
        assertThat(value).isNotEmpty();
        this.value = value;
    }

    public static SampleDocument of(final String value) {
        return new SampleDocument(value);
    }

[...]
}
```

**Describe the target audience**
Every programer who want's to do DDD and validate its value objects in the constructor to have only valid domain objects.

**Additional context**
Currently when a private constructor is already defined no static of constructor will be generated at all.

rspilker:
I understand the request, and think it is a good idea to generate the static method if `staticConstructor` is provided, even if we don't generate the (private) all-args constructor.

rspilker:
If someone would create a pull request with implementations for ecj and javac and create test files in our test infrastructure, this gets higher priority.

daliclass:
Liked the idea and have have a pull request up for this here https://github.com/rzwitserloot/lombok/pull/2119

rspilker:
It's fixed. We just need to add it to the changelog.

PR #2119, opened 2019-05-05
[Feature] staticConstructor should use already defined private constr…

daliclass:
…uctor if available

to solve issue https://github.com/rzwitserloot/lombok/issues/2100

### DIFF
--- a/src/core/lombok/eclipse/handlers/HandleConstructor.java
+++ b/src/core/lombok/eclipse/handlers/HandleConstructor.java
@@ -242,8 +242,7 @@ public class HandleConstructor {
 			
 		ASTNode source = sourceNode.get();
 		boolean staticConstrRequired = staticName != null && !staticName.equals("");
-		
-		if (skipIfConstructorExists != SkipIfConstructorExists.NO && constructorExists(typeNode) != MemberExistsResult.NOT_EXISTS) return;
+
 		if (skipIfConstructorExists != SkipIfConstructorExists.NO) {
 			for (EclipseNode child : typeNode.down()) {
 				if (child.getKind() == Kind.ANNOTATION) {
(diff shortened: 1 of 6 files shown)


rspilker:
Thanks a lot!

To be legally covered, can you please add your name to the AUTHORS file and ajust the pull request?

I'm still reviewing the code, looks good so far.

Great that you included the tests for both delombok and ecj!

src/core/lombok/eclipse/handlers/HandleConstructor.java
```diff
@@ -273,12 +272,20 @@ public void generate(
 		
 		if (noArgs && noArgsConstructorExists(typeNode)) return;
 		
-		ConstructorDeclaration constr = createConstructor(
-			staticConstrRequired ? AccessLevel.PRIVATE : level, typeNode, fieldsToParam, forceDefaults,
-			sourceNode, onConstructor);
-		injectMethod(typeNode, constr);
+		if (skipIfConstructorExists != SkipIfConstructorExists.NO && constructorExists(typeNode) != MemberExistsResult.NOT_EXISTS) {
```
rspilker: Why 
```
if (condition) {
  // generateStatic()
} else {
  // createConstructor
  // generateStatic()
}
```
instead of
```
if (!condition) {
  // createConstructor
}
//generateStatic()

I think the latter is simpler.

rspilker:
I've added a comment in the code. This comment applies for both Eclipse and Javac.

src/core/lombok/eclipse/handlers/HandleConstructor.java
daliclass: Greate catch 👍

daliclass:
> Thanks a lot!
> 
> To be legally covered, can you please add your name to the AUTHORS file and adjust the pull request?
> 
> I'm still reviewing the code, looks good so far.
> 
> Great that you included the tests for both delombok and ecj!

Added to Authors, the tests are good was easy to figure out how to add new cases etc

PR #2155, opened 2019-06-15
Avoid Eclipse warnings about lombok.NonNull when NonNullByDefault is used

brychcy:
When annotations based null analysis is enabled in Eclipse and NonNullByDefault is in effect, @lombok.NonNull causes warning, e.g. in the following example:
```
import org.eclipse.jdt.annotation.NonNullByDefault;
import lombok.NonNull;

@NonNullByDefault
public class App {
	String f(@NonNull String s) {
		return s;
	}
}
```
two warnings are reported  at the `@NonNull` annotation:
- `Dead code`
- `Redundant null check: comparing '@NonNull String' against null`

(obviously, both warnings are only reported when enabled in the Compiler preferences)


rzwitserloot:
Wow, when I read this I was totally expecting a scan for the `@NonNullByDefault` annotation, but this is a much better way to go about it. Let me just review this first, but this is the right approach.

rzwitserloot:
@brychcy Can you send another PR where you add your name to the AUTHORS file? Thanks!

PR #2136, opened 2019-05-22
Fixes for CustomLog, AUTHORS entry, extended changlog

juriad:
I noticed the style and tried to keep it mostly; I missed some places though. Thanks for fixing them.

I have no complaints to the changes.
There is one improbable breaking change: setting `lombok.log.fieldName=private` used to be valid (and it compiled!); similarly `lombok.fieldNameConstants.innerTypeName=42`, `lombok.copyableAnnotations+=int.float.package.class`, and `lombok.copyableAnnotations+=a-b-c`.
I also found two bugs regarding `TypeName`.

PR #2161, opened 2019-06-24
Add lombok.builder.className configuration property

tagae:
Allows to configure the default builder class name that `@Builder` will use when the `builderClassName` option is not passed explicitly.


juriad:
@tagae In my PR https://github.com/rzwitserloot/lombok/pull/2136, I introduced IdentifierName (https://github.com/rzwitserloot/lombok/blob/2a98c4d1fd80ff0b3181da4fe8d5f3f32f168644/src/core/lombok/core/configuration/IdentifierName.java). You might want to update yours to use it.

rzwitserloot:
Good idea but I changed the implementation considerably; yours left a bunch of now useless code, and, crucially, made it impossible to use the type name in the buildername. I noticed your update to the gradle conf file; I also pushed that through.

The changes for supporting the config property are commit c1039553b088dc45d272b942d930414d3224b06c

tagae:
Nice! Thanks a lot for taking in the idea. No problem that my code didn't get merged: I just cared about having the feature, which is the case now :-)

PR #2207, opened 2019-08-19
Allow empty build method name

tagae:


rzwitserloot:
I don't get this. a builder without a build method is completely useless. Please explain the reasoning.

Maaartinus:
@rzwitserloot  That's in the change log diff:

> ENHANCEMENT: You can now suppress the generation of the build method by passing `buildMethodName=""` as parameter to `@Builder`.

I guess, it"s for forcing the use of `toBuilder` instead.

tagae:
@rzwitserloot I have a use case where I'd like to have a build method that returns an interface implemented by the class, instead of the class type itself. So I cannot just override the build method, because the return type is different. I am aiming at Lombok-izing an existing public API, so I don't have all the control I could wish for in refactoring the API.

Allowing for omission of the build method is arguable indeed. Another, maybe better approach would be to allow changing the visibility of the build method, so that one can declare it private and wrap it with a public custom method that adds logic before or after, and such custom method could even have a different signature like in my use case. What would you say of such a feature?

janrieke:
Why don't you simply implement the `build` method manually? Lombok should not generate the method if it already exists.

tagae:
As already mentioned: I cannot override it because the signature is different (the return type is different).

janrieke:
So you have your own builder class that extends a Lombok-generated builder? But still: Why don't you implement it manually with the different signature in the Lombok-generated builder?

tagae:
If you implement the build() method manually with a different return type, Lombok will correctly detect that the build() method is not present (because the signature is incompatible), and then add an additional build() method. So there will be a compilation error because of the two incompatibly overloaded build() methods.

I don't inherit from any Lombok-generated builder. I have a number of classes in a large public API containing manually-written builders, which I would like to Lombok-ize. Thing is, those existing builders have build() methods that return an interface type rather than the class type. So if I would add the `@Builder` annotation and remove the manually-written build() methods, I would be changing their signature, which is backwards incompatible.

But I am thinking it will be better not to include the feature implemented in this pull request. If anything, the alternative I proposed earlier would be better: allow changing the visibility of the build method, so that one can declare it private (and with a different name) and wrap it with a custom build() method that can add logic before or after, and such custom method could even have a different signature if one so wishes. Waiting for @rzwitserloot 's feedback on such a feature.

janrieke:
Well, then that should be fixed: if there is a method of the same name, Lombok should not generate another (regardless of the parameters or the t return type).

tagae:
Mind that the parameter types can be considered part of the method's "name". But when having a method with the same parameters and name, though different return type, Lombok could indeed refrain from generating the method.

I will delete this pull request because other alternative features seem more appropriate to have than this one.

janrieke:
I'm confused now. I just looked it up in the sourcecode and Lombok actually behaves how it should: It only checks for the name of the method, nothing else. So when you simply implement your own `build` method, everything should be working.
So either you did something wrong when customising the builder class, or you spotted a bug. So please file a bug report (with a minimal example) so that we can look into it.

PR #2220, opened 2019-08-27
Update gradle documentation

larsgrefer:
The linked gradle plugin doesn't work with Gradle 2.12 (which is way too old anyway) and doesn't support running the eclipse installer.


rspilker:
Thanks

PR #2245, opened 2019-09-23
Update With.html

jyterencekim:
## Summary
The current sentence fragment (`Lombok's own <a href="/features/constructor"><code>@AllArgsConstructor</code></a>`) does not fully clarify the author's intention, which I presume is to recommend using `@AllArgsConstructor`.


rzwitserloot:
Ah, yeah, your take is much better. Muchos gracias; integrated in the source, and because it's much better, I'm going to quickly go fix the website instead of waiting for the next release cycle.

Issue #2268, opened 2019-10-20
[BUG] problem with setter when extending a class with the same simple name 


mateidavid:
**Describe the bug**

Original report on the Google group:
https://groups.google.com/d/msg/project-lombok/YZ8CRFX8HFA/ZSZIa4c2BgAJ

When the Lombok annotation processor generates a chained setter for an inner class (e.g. `A.Inner`), it attempts to use as return type only the simple name of the inner class (`Inner`), as opposed to the partially qualified name that includes the enclosing class (`A.Inner`). In some contexts the simple name of the inner class might not be unique, leading to compilation error.

**To Reproduce**

```
interface A {
   int a();

   @FieldDefaults(level = AccessLevel.PRIVATE)
   @Getter
   @Setter
   @Accessors(fluent = true)
   class Mutable implements A {
      int a;
   }
}

interface B extends A {
   int b();

   @FieldDefaults(level = AccessLevel.PRIVATE)
   @Getter
   @Setter
   @Accessors(fluent = true)
   class Mutable extends A.Mutable implements B {
      int b;
   }
}
```

On build, this produces an error that in `B.Mutable`, the reference to `Mutable` is ambiguous. Most likely, that's because Lombok produces:

```
public Mutable b(int b) {
   this.b = b;
   return this;
}
```

... and in that context `Mutable` is ambiguous between `A.Mutable` and `B.Mutable`.

**Expected behavior**

Have the setter return type use the partially qualified name that includes the enclosing class, elimintaing the ambiguity:

```
public B.Mutable b(int b) {
   this.b = b;
   return this;
}
```

**Version info (please complete the following information):**

- Lombok 1.18.8
- JDK 1.8.0_221
- IntelliJ 2019.1.1

**Additional context**

- Note that IntelliJ does not show an error, suggesting the Lombok IntelliJ plugin somehow does the right thing.
- The workaround is to de-lombok the setters inside `B.Mutable`.

ghunteranderson:
This is so interesting! I spent some time digging into this and wanted to share what I found. The tldr is that everything confirms what @mateidavid suggested with one addition. All outer classes need to be specified, not just the most immediate.

## Original Issue
I debugged the `@Setter` handler and confirmed your assumption about the generated code is correct. It just returns type `Mutable`. I've been playing around with your example to get the minimal code that still creates the issue. Here's how far I got.
``` java
interface A {
    class Foo {}
}

interface B {
    @Setter
    @Accessors(fluent = true)
[...]

PR #2274, opened 2019-10-28
Using partially qualified name when cloning a type reference

ghunteranderson:
I starting working on a solution for #2268. 

Referring to a nested class by simple name can create ambiguous code or resolve to a type other than what was expected. This bug fix tries to address issues with the return types of fluent setters being ambiguous or resolving to a type other than `this`. The change attempts to include all parent types in the JCExpression or TypeReference creating a more specific reference.

Disclaimer: This is my first time working with Eclipse/Javac API for parse trees and documentation is hard to find. Feedback is more than welcome if there's a better way to do this.

### DIFF
--- a/src/core/lombok/eclipse/handlers/EclipseHandlerUtil.java
+++ b/src/core/lombok/eclipse/handlers/EclipseHandlerUtil.java
@@ -35,6 +35,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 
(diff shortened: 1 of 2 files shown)


rzwitserloot:
This was vastly more complicated; you really picked a doozy to work on!

I hope you don't mind that I kinda started over; lombok generates types (which therefore may need qualifying) _everywhere_, particularly Handle(Super)Builder.

The work is in 889c935ec9f0e45bba1e88b0f256e1f29a734f39 and I did credit you in the changelog, thanks for showing me the way :)

ghunteranderson:
Glad I could help. 😊 Im looking forward to learning from your commit.

PR #2283, opened 2019-11-04
Update the list of Checker Framework annotations for version 3.0.0.

smillst:


rzwitserloot:
Given that something like `org.checkerframework.checker.nullness.qual.NonRaw` is in 2.9, maybe we should leave all the types in; lombok doesn't mind if some of these types no (longer) exist in 3.0. Can you update the PR?

smillst:
@rzwitserloot I've updated the pull request and added a comment to only add annotations.

PR #2347, opened 2020-01-24
Change README to README.md and add hyperlinks

kraiss:
Spicing up the README file with markdown syntax and hyperlinks to tidelift subscription and files mentionned.

What do you think ?


rspilker:
I like the idea. To improve plain text reading, I prefer numbered reference links, like 
```
[LICENSE][1]
...
   [1]: https://github.com/rzwitserloot/lombok/blob/master/LICENSE

```

That would be rendered as:
---
[LICENSE][1]
...
   [1]: https://github.com/rzwitserloot/lombok/blob/master/LICENSE
---

I also think the build file needs to be modified to put it in the jar.

kraiss:
I made some changes according to your feedbacks. 

I used "text" reference links so it keeps a bit of context for someone reading/updating the links part only.

rzwitserloot:
We'll take a look this thursday. Thanks, @kraiss!

PR #2355, opened 2020-02-06
Avoid more name clashes for SuperBuilder

### LINKED ISSUE #2297: [BUG] @SuperBuilder does work on classes named B or C
**Describe the bug**
`@SuperBuilder` causes a compilation error when used on classes named `B` or `C`.

**To Reproduce**
Put `@SuperBuilder` on a class named `B` or `C`.

**Expected behavior**
Compiles.

**Version info:**
 - Lombok version 1.18.10
 - javac and eclipse

**Additional context**
Right now lombok only avoids name clashes between builder generics and the class's type parameters. When I implemented `@SuperBuilder`, I forgot to also check for clashes with the class name itself. :(

janrieke:
Previously, the names of the type parameters of `@SuperBuilder` are chosen that they do not collide with the type parameters of the annotated class.

With this PR, the classname of the annotated class and the type names of fields in that class are also considered in this collision check.

This fixes #2297.

### DIFF
--- a/src/core/lombok/eclipse/handlers/HandleSuperBuilder.java
+++ b/src/core/lombok/eclipse/handlers/HandleSuperBuilder.java
@@ -29,6 +29,7 @@ import java.lang.reflect.Modifier;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.List;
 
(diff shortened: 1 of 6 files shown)


rzwitserloot:
Not sure we really ought to cater to people naming classes `B`, but, hey. I'll take it, thanks @janrieke!

PR #2387, opened 2020-03-03
@Jacksonized: modify generated @(Super)Builders such that they will be used by Jackson's deserialization

janrieke:
This PR implements the proposal [discussed here](https://groups.google.com/forum/#!topic/project-lombok/L7OVH_noUAY).

Adding `@Jacksonized` to a context where there is also a `@Builder` or `@SuperBuilder` annotation does the following:
1. Configure Jackson to use the builder for deserialization using `@JsonDeserialize(builder=MyClass.MyClassBuilder[Impl].class))` on the class. (An error is emitted if such an annotation already exists.)
2. Copy Jackson-related configuration annotations (like `@JsonIgnoreProperties`) from the class to the builder class. This is necessary so that Jackson recognizes them when using the builder.
3. Insert `@JsonPOJOBuilder(withPrefix="")` on the generated builder class to override Jackson's default prefix "with". If you configured a different prefix in lombok using `setterPrefix`, this value is used. If you changed the name of the `build()` method using using `buildMethodName`, this is also made known to Jackson.
4. For `@SuperBuilder`, make the builder implementation class package-private.

There is one minor difference to the proposal (with regard to step 1.): If there is already a `@JsonDeserialize` annotation on the class, it will give an error. The reason is that all other parameters of `@JsonDeserialize` are only useful when _not_ using a builder; often they will even contradict the usage of a builder. So I though it is reasonable to error out then and let the user decide what to do.

I did not yet write documentation for the website, because I didn't where to place it (there is no "external" section in the site). I propose adding it as "experimental".

### DIFF
--- a/src/core/lombok/core/handlers/HandlerUtil.java
+++ b/src/core/lombok/core/handlers/HandlerUtil.java
@@ -76,7 +76,7 @@ public class HandlerUtil {
 		return 43;
 	}
 	
-	public static final List<String> NONNULL_ANNOTATIONS, BASE_COPYABLE_ANNOTATIONS, COPY_TO_SETTER_ANNOTATIONS;
+	public static final List<String> NONNULL_ANNOTATIONS, BASE_COPYABLE_ANNOTATIONS, COPY_TO_SETTER_ANNOTATIONS, JACKSON_COPY_TO_BUILDER_ANNOTATIONS;
 	static {
 		NONNULL_ANNOTATIONS = Collections.unmodifiableList(Arrays.asList(new String[] {
 			"androidx.annotation.NonNull",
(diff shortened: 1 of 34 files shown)


janrieke:
I pushed two new commits to the branch/PR:
* add flag usage configuration key (I just noticed I forgot this)
* add documentation and changelog entry

rzwitserloot:
This is amazing work, @janrieke – all I really had to do was think of a funny byline :)

src/core/lombok/core/handlers/HandlerUtil.java
```diff
@@ -315,6 +315,18 @@ public static int primeForNull() {
 			"com.fasterxml.jackson.annotation.JsonProperty",
 			"com.fasterxml.jackson.annotation.JsonSetter",
 		}));
+		JACKSON_COPY_TO_BUILDER_ANNOTATIONS = Collections.unmodifiableList(Arrays.asList(new String[] {
```
cowtowncoder, line 318: Looks like a reasonable list: the only one I was thinking might make sense to add would be `JsonAutoDetect`?

janrieke: So I suppose Jackson will just evaluate `setterVisibility` in the builder case, right?

cowtowncoder: Jackson will actually do full visibility evaluation for all accessors since sometimes visible getter is required to pull in non-visible setter. But for most cases I assume `setterVisibility` / `fieldVisibility` is what really matters.

janrieke: I created PR #2475 for this.

z0mt3c: What about com.fasterxml.jackson.annotation.JsonAnyGetter and com.fasterxml.jackson.annotation.JsonAnySetter?

janrieke: `@JsonAnyGetter` is only relevant for serialization, so there is no use in copying it to a builder.

A `@JsonAnySetter` method is called once for every attribute that is not directly mapped to a field. Considering lombok, it will work only for the singular method of a `@Singular Map<String, Object>` field, and this is exactly where the current edge release will copy it to.

Is there another use-case I don't see?

PR #2408, opened 2020-03-25
Endline Normalization and bugfixing

PabloNicolasDiaz:
I've fixed various bugs in Lombok :

1.  In classes with already defined factory methods and constructors, but a "staticConstructor" is also included, then lombok does not generate any method (see ValueStaticConstructorOfOverriden test cases).

2. After running Spotbugs over Lombok, I found the error VA_FORMAT_STRING_USES_NEWLINE in various calls, and I fixed it.

3. I normalized some files (https://help.github.com/en/github/using-git/configuring-git-to-handle-line-endings)

Please, take these PR in account.


rzwitserloot:
I can't accept this PR. The stylistic changes (specifically, normalizing the line endings) is not okay; the noise in the tracker is not worth it, and git already takes care of this. The VA_FORMAT_STRING_USES_NEWLINE spotbugs error is in our opinion incorrect; java will render just \n fine on windows terminals, it's \n on all other OSes worth talking about, and having 'stable' output is on its own useful as well.

Your [1] item is not entirely clear to me; the fact that your file update there also looks like every line changed means you're making it very hard for me to diff this. I'm guessing that's referring to a bug whereby asking lombok to generate a static constructor may cause issues if you already have it?

Thanks for the effort, but by bundling these things into one package I can't accept this :(

I'd prefer it if it went like this:

1. file a bug for that staticonstructor thing with a clear example of how current master branch lombok gets it wrong,

2. file a PR with a fix, where the diff lists only actually changed lines (so don't autoformat the entire file, or change the line ending on every line).

rzwitserloot:
I'm going to deny this PR; I've looked at `ValueStaticConstructorOfOverriden` and it seems like the bug report you're trying to fix here is that lombok will attempt to generate the static method even if it is already explicitly there.

That's not a bug, it's intentional: If you go out of your way to explicitly tell lombok to generate a method by some certain name, then lombok is going to do what you asked it to. Lombok only silently does nothing in specific cases where it makes sense (such as not generating an already-existing getter, but only if the lombok annotation that would generate it is also doing other things (such as Getter on an entire class, or `@Data`). Where convenient lombok will recognize the problem and generate a more specific and clear error message, but in this case that's not worth the maintenance burden.

PR #2416, opened 2020-04-06
Also copy @JacksonXmlProperty to setter

janrieke:
`@JacksonXmlProperty` is the XML counterpart to `@JsonProperty`. It allows XML-specific settings, e.g., setting the namespace.

This will allow using a builder to deserialize XML when there are `@JacksonXmlProperty`s present.
See also FasterXML/jackson-dataformat-xml#384.

PS: There are probably a few more of those annotation in the JAXB annotation package, but I never worked with those, so I'm not sure for which of these annotations copying would be always correct.


janrieke:
Maybe we should wait for the results of the discussion on #2419. There are probably some more annotations that should be in this list.

janrieke:
Closing for now. I'll create a new PR when I figured out how to deal with `@Singular`.

PR #2421, opened 2020-04-11
Encode partial URL. Fixes #1536

### LINKED ISSUE #1536: ClassNotFoundException: lombok.core.AnnotationProcessor
Hello,

I've created a simple project with 1 class inside:
```
@Getter
public class Test {
   private String val = "some value";
}
```

I put the Lombok jar inside a my-libs directory within the project.

When I try to compile this class using `javac -cp my-libs/lombok-1.16.18.jar src/main/java/Test.java`, everything works as expected.

When I rename the 'my-libs' directory to 'my+libs' and try to compile the class again using `javac -cp my+libs/lombok-1.16.18.jar src/main/java/Main.java`, i am getting this error:

`error: Bad service configuration file, or exception thrown while constructing Processor object: javax.annotation.processing.Processor: Provider lombok.launch.AnnotationProcessorHider$AnnotationProcessor could not be instantiated: java.lang.RuntimeException: java.lang.ClassNotFoundException: lombok.core.AnnotationProcessor`

The only difference between the 2 cases, is the presence of the '+' sign in the name of the directory.

Please find an example ([lombok-test.zip](https://github.com/rzwitserloot/lombok/files/1555648/lombok-test.zip)) that shows how to reproduce the problem.

It is not common to have a '+' in the name of the directory, but I can't just understand why it is not working.

Regards,

tkalmar:
url.toString is not encoded properly, so to not lose the "+" in the url one has to properly encode it before decoding other chars like " " are correctly encoded
According to https://docs.oracle.com/javase/7/docs/api/java/net/URLDecoder.html this behaviour is correct "+" is Encoded as " " which is not the file someone whants
This Fixes #1536

### DIFF
--- a/src/launch/lombok/launch/ShadowClassLoader.java
+++ b/src/launch/lombok/launch/ShadowClassLoader.java
@@ -321,8 +321,9 @@ class ShadowClassLoader extends ClassLoader {
 	}
 	
 	private static String urlDecode(String in) {
+		final String plusFixed = in.replaceAll("\\+", "%2B");
 		try {
-			return URLDecoder.decode(in, "UTF-8");
+			return URLDecoder.decode(plusFixed, "UTF-8");
 		} catch (UnsupportedEncodingException e) {
 			throw new InternalError("UTF-8 not supported");
 		}


rzwitserloot:
We'll put this one on the 'boy, who finds this stuff?' list. 👏 colour me impressed!

PR #2459, opened 2020-05-09
Handle generic supertypes

### LINKED ISSUE #2382: [BUG] @Delegate compilation issue
hello.
because an example is much better than a lot of words: 
here is a public repo where you can see the issue:
https://github.com/clembou590/lombokFails

two branches in this repo:
**"master" branch** works. (just type mvn clean install)
**"fails" branch** does not work (just type mvn clean install)

the 2 branches should succeed... (the only difference between the 2 branches is the use of @Delegate)

here is the error I get in the "fails" branch
Compilation failure
[ERROR]  cannot find symbol
[ERROR]   symbol:   method save(com.example.deleg.fail.Message)
[ERROR]   location: variable messageService of type com.example.deleg.fail.MessageService

for information 
1.18.12 is the lombok version that is brought in by  org.springframework.boot:spring-boot-starter-parent:2.2.5.RELEASE

Rawi01:
This PR resolves #2382. 
`supertype_field` and `interfaces_field` are empty if the parent uses generics. The method `types.directSupertypes` returns the parent and sets the generic type information properly.

### DIFF
--- a/src/core/lombok/javac/handlers/HandleDelegate.java
+++ b/src/core/lombok/javac/handlers/HandleDelegate.java
@@ -389,10 +389,11 @@ public class HandleDelegate extends JavacAnnotationHandler<Delegate> {
 			boolean isDeprecated = (member.flags() & DEPRECATED) != 0;
 			signatures.add(new MethodSig(member.name, methodType, isDeprecated, exElem));
 		}
-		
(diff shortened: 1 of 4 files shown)


rspilker:
Thanks

rspilker:
Including tests!

PR #2475, opened 2020-05-26
[Jacksonized] copy JsonAutoDetect to builder class

janrieke:
Follow-up for PR #2387, as per remark of Jackson maintainer @cowtowncoder.


rspilker:
Thanks!

Issue #1543, opened 2017-12-19
[Feature request] Reordering of comparison in @EqualsAndHashCode for better performance


stsypanov:
Consider simple class:
```java
@EqualsAndHashCode
class Smth {
  Long id;
  List<Long> props;
  String name;
  int count;
}
```

It generates this code:

```java
public boolean equals(Object o) {
  if (o == this) return true;
  if (!(o instanceof Smth)) return false;
  final Smth other = (Smth) o;
  if (!other.canEqual((Object) this)) return false;
  final Object this$id = this.id;
  final Object other$id = other.id;
  if (this$id == null ? other$id != null : !this$id.equals(other$id)) return false;
  final Object this$props = this.props;
  final Object other$props = other.props;
  if (this$props == null ? other$props != null : !this$props.equals(other$props)) return false;
  final Object this$name = this.name;
  final Object other$name = other.name;
  if (this$name == null ? other$name != null : !this$name.equals(other$name)) return false;
  if (this.count != other.count) return false;
  return true;
}

protected boolean canEqual(Object other) {
  return other instanceof Smth;
}
```

Then the same code generated by IDEA (for Java 7+):

```java
public boolean equals(Object o) {
    if (this == o) return true;
    if (o == null || getClass() != o.getClass()) return false;
    Smth smth = (Smth) o;
    return count == smth.count &&
            Objects.equals(id, smth.id) &&
            Objects.equals(props, smth.props) &&
            Objects.equals(name, smth.name);
}
```

IDEA-generated code reorders comparison (primitives and wrappers first) yielding better performance for the case when `count` or `id` fields are mutually different and we can immediately return `false` skipping more complicated comparison of strings and collections.

 It'd be nice if Lombok could do the same.

Maaartinus:
I'm not sure about the usefulness of such an optimization. Obviously, testing the string before the list is better if only the former differs. It's probably also better in the average real use case, but what if the string is megabytes long and the list differ already in their size?

I always order my fields accordingly; not as an optimization, but placing the simple things first makes sense to me.  There's a proposal for generating `compareTo`, where the fields can be ranked via a special annotation.

stsypanov:
My proposal is only about primitives and wrappers like `Integer` (their comparison is in fact comparison of underlying primitives) as you are never sure about other fields of type `Object`.

Maaartinus:
[...]

PR #2485, opened 2020-06-05
[Feature request] @EqualsAndHashCode Flexibility for rank properties for equal/hashcode method

samukce:
Given some special cases where we need to explicitly order the comparation fields for performance improvement, I have applied the same idea that we have for `@ToString.Include(rank = -1)`. ie.: `@EqualsAndHashCode.Include(rank = -1)`

Also, there are some discussions related to this here (#1543) about the Equals algorithm and performance. So, given this flexibility can help me with some special cases and I believe that can be helpful for more people too. Let me know what you think folks.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -34,6 +34,7 @@ Robbert Jan Grootjans <grootjans@gmail.com>
 Robert Wertman <robert.wertman@gmail.com>
 Roel Spilker <r.spilker@gmail.com>
 Roland Praml <pram@gmx.de>
+Samuel Pereira <samuel.p.araujo@gmail.com>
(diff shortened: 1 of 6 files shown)


samukce:
hello @rspilker , could you take a look at this PR? This improvement would help me given that I use to need to tunning java code and be able to handle those equals configuration through Lombok would be helpful and keep the code clean.

rspilker:
Great job!

I love that you've even added a test case, and added you name the the AUTHORS file.

rspilker:
This also opens the door to give primitives as default rank. We could document them to have a default of 1000 for instance.

rspilker:
I'm nearly done

PR #2487, opened 2020-06-10
[fixes #2481] Copy constructor javadoc to builder methods

### LINKED ISSUE #2481: [BUG] @Builder methods do not copy javadoc
**Describe the bug**
1. When using `@Builder` on a **class** while having a javadoc comment on a **field**, the resulting builder method does not show the javadoc of the field. (Delombok generates a javadoc comment in this case.)
2. Likewise, when using `@Builder` on a **constructor** while having a javadoc comment on a **constructor parameter**, the resulting builder method does not show the javadoc of the parameter. (Neither Delombok not the Eclipse agent generate a javadoc comment in this case.)

**To Reproduce**
1. `@Builder` on class:
```
@lombok.Builder
public class Test1 {
    /** Comment on Field */
    private int number;

    /** @param number Comment on Constructor */
    public Test1(final int number) {
        this.number = number;
    }

    private void bla() {
        Test1.builder()
            .number(1) // <--- no Javadoc displayed in tooltip without delombok, but delombok would generate Javadoc
            .build();
    }
}
```

2. `@Builder` on constructor:
```
public class Test2 {
    /** Comment on Field */
    private int number;

    /** @param number Comment on Constructor */
    @lombok.Builder
    public Test2(final int number) {
        this.number = number;
    }

    private void bla() {
        Test2.builder()
            .number(1) // <--- no Javadoc displayed in tooltip without delombok, and delombok won't generate Javadoc
            .build();
    }
}
```

**Expected behavior**
1. When using `@Builder` on a class while having a javadoc comment on a field, the resulting builder method should have a javadoc comment which contains the javadoc of the field. This should be consistent with the delombok behaviour (which generates a javadoc comment from the field's javadoc).
2. Likewise, when using `@Builder` on a constructor while having a javadoc comment on a constructor parameter, the resulting builder method should have a javadoc comment which contains the javadoc of the parameter. This seems to be completely missing - neither delombok nor the Eclipse agent generate javadoc in this case.
(It would also be OK if the javadoc comment of the identically-named field was copied, but that seems somewhat sloppy. Most of the time, there's a reason to use `@Builder` on the constructor, so the information should come from there.)

 - Lombok 1.18.12
 - Eclipse 2020-03 (4.15.0)

**Additional context**
I know that #2008 should have addressed this issue and was baffled when I learned that it doesn't work - at least for me. I also read through #1033 and #1445 but didn't find out how I may have failed, but I'm not a Lombok expert, so please tell me if I missed something!

Rawi01:
As discussed in #2481 `@Builder` now copies the `@param` part of the constructor javadoc to the generated builder method.

### DIFF
--- a/src/core/lombok/javac/handlers/HandleBuilder.java
+++ b/src/core/lombok/javac/handlers/HandleBuilder.java
@@ -27,6 +27,8 @@ import static lombok.javac.JavacTreeMaker.TypeTag.typeTag;
 import static lombok.javac.handlers.JavacHandlerUtil.*;
 
 import java.util.ArrayList;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import javax.lang.model.element.Modifier;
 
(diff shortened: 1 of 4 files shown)


rspilker:
Thanks

PR #2497, opened 2020-06-22
Fix missing parameter names, annotations in following annotation processors

dstepanov:
Fixes https://github.com/micronaut-projects/micronaut-core/issues/1566

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -9,6 +9,7 @@ Christian Sterzl <christian.sterzl@gmail.com>
 DaveLaw <project.lombok@apconsult.de>
 Dave Brosius <dbrosius@mebigfatguy.com>
 Dawid Rusin <dawidrusin90@gmail.com>
+Denis Stepanov <denis.stepanov@gmail.com>
 Emil Lundberg <emil@yubico.com>
 Enrique da Costa Cambio <enrique.dacostacambio@gmail.com>
 Jacob Middag <jacob@gaddim.nl>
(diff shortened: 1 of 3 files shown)


Rawi01:
- I think `savedParameterNames` is not required, `params()` and `getParameters()` returns the `params` field if it is `!= null` and will not read `savedParameterNames`
- `parameterNames` is never used
- To build a new `List`, `ListBuffer` can be used

dstepanov:
@Rawi01 Thanks for suggestions!

rspilker:
@dstepanov Can you please add you name to the AUTHORS file, so we are legally covered?

dstepanov:
@rspilker Added

dstepanov:
@rspilker Do you think this PR can make it to the next release?

rspilker:
Yes, Thursday I'll work on getting this in. We do not have a release date yet.

Rawi01:
@dstepanov I noticed a bunch of `NullPointerExcceptions` compiling with this fix, it seems like that `param.sym` can be `null` and the `ListBuffer` does not allow `null` values. I had no time to look into it and figure out why it is `null` sometimes.

dstepanov:
@Rawi01 What are you compiling?

Rawi01:
Should have mentioned that, I noticed it compiling this one: https://github.com/dmuatdbh/lombok (you have to run the ClassGenerator first) using Java 8

dstepanov:
Probably some annotation processor is generating incorrect model. It looks like it needs null check :-/

rspilker:
Last Thursday, I already added the null checks. So if any `param.sym` is null, we now fall back to the old behavior.

But since we don't have tests for this feature, I don't know when this occurs, nor if there is a scenario when it doesn't.

PR #2499, opened 2020-06-24
Skip type resolution in isDelegate

Rawi01:
During profiling I noticed that `PatchDelegate.isDelegate()` consumes some time during compilation. I noticed that the type resolution of the annotation is responsible for that. Even if I'm not sure if that this increases the compilation time I added an additional check that returns early.

This causes some interesting side effects as it broke some unrelated test cases. After checking out why, I think that the unsuccessful type resolution (due to the unknown ckeckerframework annotation) stops compilation to early and hides the underlying problem. It seems like the `this` parameter have to be set as `Receiver` instead of an `Argument` in eclipse. Otherwise it fails with `Only the first formal parameter may be declared explicitly as 'this'`.

I think I fixed that but couldn't test it in a real world scenario.


rspilker:
Nice job!

rspilker:
I think we should have seen that the constructor body was empty...

PR #2500, opened 2020-06-24
Fix UtilityClass test case

Rawi01:
The changes introduced in 94440a8bd73f176637a9890dc10df67d26615674 breaks the test case. This PR fixes that.


rspilker:
Thanks

PR #2506, opened 2020-06-30
Add config key to turn off @SuppressWarnings("all")

### LINKED ISSUE #2433: [FEATURE] config key to turn off @SuppressWarnings("all") on generated members
**Describe the feature**

AFAIK there's currently no way to tell Lombok via the config that the generated methods should NOT have `@SuppressWarnings("all")`. That would be useful in certain contexts.

I'm aware that `--format=suppressWarnings:skip` flag exists for `delombok`, but I don't think that's possible without using the Delombok feature (e.g. when Lombok is simply referenced in a Gradle project as `annotationProcessor` & `compileOnly`).

Developers using static code analyzers, like https://checkerframework.org/.

**Additional context**

Let's consider the following interface:

```java
interface IFoo {
  String getFoo(); // Checker Framework assumes the result is @NonNull by default
}
```

and an implementing class:

```java
import org.checkerframework.checker.nullness.qual.Nullable;

class Foo implements IFoo {
  @lombok.Getter
  @Nullable // overridden getter must NOT be nullable since it's declared non-null in IFoo
  private final String foo;
}
```

The Lombok-generated getter will be something like:

```java
  @org.checkerframework.checker.nullness.qual.Nullable
  @SuppressWarnings("all")
  @lombok.Generated
  public String getFoo() {
    return this.foo;
  }
```

**If it wasn't for `@SuppressWarnings("all")`, then [NullnessChecker](https://checkerframework.org/manual/#nullness-checker) would issue an error, which it definitely should in this case**: a method declared as returning non-null in the interface can't be implemented as possibly returning null.

This wouldn't be a problem if a config key to turn off automatic warning suppression existed.

Rawi01:
This PR fixes #2433

### DIFF
--- a/src/core/lombok/ConfigurationKeys.java
+++ b/src/core/lombok/ConfigurationKeys.java
@@ -90,6 +90,13 @@ public class ConfigurationKeys {
 	public static final ConfigurationKey<Boolean> ADD_FINDBUGS_SUPPRESSWARNINGS_ANNOTATIONS = new ConfigurationKey<Boolean>("lombok.extern.findbugs.addSuppressFBWarnings", "Generate @edu.umd.cs.findbugs.annotations.SuppressFBWarnings on all generated code (default: false).") {};
 	
 	/**
+	 * lombok configuration: {@code lombok.addSuppressWarnings} = {@code true} | {@code false}.
+	 * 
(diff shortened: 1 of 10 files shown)


rspilker:
Once again perfect. Including tests and documentation!

PR #2512, opened 2020-07-09
Reduce memory usage

Rawi01:
While analyzing heap dumps for #2240 I noticed that lombok keeps some unnecessary objects in memory. This PR changes that.

I also removed some old code that copies a list into a new one which actualy does not change anything. This snippet was part of an old experiment and even contains an if condition that can never be true.

### DIFF
--- a/src/core/lombok/javac/JavacTransformer.java
+++ b/src/core/lombok/javac/JavacTransformer.java
@@ -21,7 +21,6 @@
  */
 package lombok.javac;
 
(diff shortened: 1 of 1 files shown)


rspilker:
Yeah, that IF can never trigger, we don't need the copy, nor keep the AST's in a list.

leonardowada:
@rspilker do you know when will be a version with this commits?

PR #2514, opened 2020-07-11
Set generated flag for annotation names

### LINKED ISSUE #977: @RequiredArgsConstructor + @NonNull breaks eclipse highlighting
I simplified code that breaks highlighting in eclipse:

```
package highlight;

import lombok.NonNull;
import lombok.RequiredArgsConstructor;

@RequiredArgsConstructor
public enum LombokHighlight {
    TYPE1("type1");

    @NonNull
    private final String format;

    public boolean method1(String s1, String s2) {
        return s1.equals(s2);
    }

}
```

Open this file in Eclipse (initially highlight is fine) then click on `equals` method in `return s1.equals(s2);` clause. Important: click just after opening file. Don't click anywhere else.
All upper code will be highlighted as comment (gray in my color scheme).

After this clicking on classname fixing highlight and clicking `equals` again don't break it anymore. But if you click at random places in code then after some time highlight is breaks again. This is not so oftern in this simple file. But in real project with bigger file when this enum is nested in other class etc - this is became very very often.

lombok-highlight

Rawi01:
This PR fixes #977. 
Unfortunately, as with all these changes, it is not possible to create a test case.


rspilker:
Thanks again

PR #2533, opened 2020-08-03
changed arraylist to set in builder doc

lebeier:
Hi,

I have noticed a discrepancy in the documentation on the variable type of the `occupation` variable for the Builder documentation. Therefore, I have decided to make a pull request to correct the mistake. Feel free to close this if the change is incorrect.

Thanks!

### DIFF
--- a/website/usageExamples/BuilderExample_post.jpage
+++ b/website/usageExamples/BuilderExample_post.jpage
@@ -25,7 +25,7 @@ public class BuilderExample {
 		private boolean created$set;
 		private String name;
 		private int age;
-		private java.util.ArrayList<String> occupations;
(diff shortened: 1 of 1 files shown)


rspilker:
Sorry, I was a bit too fast in accepting the change.

`new Set<String>()` doesn't compile, since `Set` is an interface.

The other reason is that in Lombok we store the data in the Builder in an ArrayList. In the builder, we don't care about duplicates, and array lists are nice, small and fast.

lebeier:
The change was to make it consistent with the type that was declared at the top of the document which was declared as Set<String>. I do agree with you that I should have used HashSet<String> during assignment. I will probably create another MR to rectify this issue.

Issue #1745, opened 2018-06-27
Delombok javadoc output doesn't work correctly under Windows


grv87:
Given the source:
```
  /**
   * @param websiteUrl project website URL
   * @return project website URL
   */
  @Getter @Setter
  private String websiteUrl;

  /**
   * @param issuesUrl issues URL
   * @return issues URL
   */
  @Getter @Setter
  private String issuesUrl;
```
using delombok under Windows (CR+LF line endings) I got extra empty lines that are not parsed correctly by javadoc:
```
  /**
   *

   * @return issues URL
   */
  @java.lang.SuppressWarnings("all")
  public String getIssuesUrl() {
    return this.issuesUrl;
  }

  /**
   * @param issuesUrl issues URL

   */
  @java.lang.SuppressWarnings("all")
  public void setIssuesUrl(final String issuesUrl) {
    this.issuesUrl = issuesUrl;
  }
```

grv87:
Sorry for delay.
Lombok 1.18.2
JDK 1.8.0u172
Yes, I use Gradle.

Test case: https://github.com/grv87/delombok-windows

Additional info: the problem occurs only when both `@param` and `@return` are used.

grv87:
The problem arises under Windows only and with source files having CRLF line endings only.
Under Linux, even if CRLF line endings are used, no problem.

rosti-il:
Please consider my pull request of #2548
Change of the JavacHandlerUtil.java in that PR fixes this issue by the right way.
https://github.com/rzwitserloot/lombok/pull/2548/commits/87d8e0016f17472af46aba8cec76445eae142d0d

rzwitserloot:
PR 2548 fixes this.

PR #2536, opened 2020-08-04
Fix #1745 : Delombok javadoc output doesn't work correctly under Windows

04637:
### Solve the problem of extra blank lines in Windows (CR+LF line endings). 

 **_My test files:_**
Source file
```
package test.de;

import lombok.Data;

/**
 * todo
 *
 * @author: 04637@163.com
 * @date: 2020/8/4
 */
@Data
public class Car {
    // Yeah! This is my car
    private String name;
    private String owner;
}

```
After delombok
```
package test.de;

/**
 * todo
 *
 * @author: 04637@163.com
 * @date: 2020/8/4
 */
public class Car {
    // Yeah! This is my car
    private String name;
    private String owner;

    @SuppressWarnings("all")
    public Car() {
    }

    @SuppressWarnings("all")
    public String getName() {
        return this.name;
    }

    @SuppressWarnings("all")
    public String getOwner() {
        return this.owner;
    }

    @SuppressWarnings("all")
    public void setName(final String name) {
        this.name = name;
    }

    @SuppressWarnings("all")
    public void setOwner(final String owner) {
        this.owner = owner;
    }

    @Override
    @SuppressWarnings("all")
    public boolean equals(final Object o) {
        if (o == this) return true;
        if (!(o instanceof Car)) return false;
        final Car other = (Car) o;
        if (!other.canEqual((Object) this)) return false;
        final Object this$name = this.getName();
        final Object other$name = other.getName();
        if (this$name == null ? other$name != null : !this$name.equals(other$name)) return false;
        final Object this$owner = this.getOwner();
        final Object other$owner = other.getOwner();
        if (this$owner == null ? other$owner != null : !this$owner.equals(other$owner)) return false;
        return true;
    }

    @SuppressWarnings("all")
    protected boolean canEqual(final Object other) {
        return other instanceof Car;
    }

    @Override
    @SuppressWarnings("all")
    public int hashCode() {
        final int PRIME = 59;
        int result = 1;
        final Object $name = this.getName();
        result = result * PRIME + ($name == null ? 43 : $name.hashCode());
        final Object $owner = this.getOwner();
        result = result * PRIME + ($owner == null ? 43 : $owner.hashCode());
        return result;
    }

    @Override
    @SuppressWarnings("all")
    public String toString() {
        return "Car(name=" + this.getName() + ", owner=" + this.getOwner() + ")";
    }
}

```

PR #2548, opened 2020-08-15
 Fix tests under Windows, fix test.javac11 and test.javac14, fix issue #1745

rosti-il:
This also should fix the #1745 issue.

The change of the TestConfiguration.java is based on the fact that
Git for Windows is configured with 'core.autocrlf=true' by default.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -35,6 +35,7 @@ Robbert Jan Grootjans <grootjans@gmail.com>
 Robert Wertman <robert.wertman@gmail.com>
 Roel Spilker <r.spilker@gmail.com>
 Roland Praml <pram@gmx.de>
+Rostislav Krasny <45571812+rosti-il@users.noreply.github.com>
 Samuel Pereira <samuel.p.araujo@gmail.com>
 Sander Koning <askoning@gmail.com>
 Szymon Pacanowski <spacanowski@gmail.com>
--- a/buildScripts/tests.ant.xml
+++ b/buildScripts/tests.ant.xml
@@ -119,6 +119,7 @@ This buildfile is part of projectlombok.org. It takes care of compiling and runn
 				<classpath refid="cp.stripe" />
 				<classpath location="build/lombok" />
 				<classpath location="build/tests" />
+				<classpath location="build/teststubs" />
 				<test name="lombok.TestJavac" />
 			</junit>
 		</sequential>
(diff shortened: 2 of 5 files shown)


rosti-il:
If you use lombok.maben plugin you must rebuild it with the new version of Lombok.
Just bump its version defined in the lombok.version property:
https://github.com/awhitford/lombok.maven/blob/master/pom.xml

Rawi01:
Unfortunatly there already is a PR (#2536) that also fixes this issue. Seems to be important :smile: 
I am courious if both actually solve the problem completly. If I understand the problem correctly there is an additional `\r` at the end. This PR tries to avoid adding that by removing it early on, the other removes it during writing. Not adding them at all is IMHO the better soloution. Have you tried to compile a non CRLF file on windows? Does that work after applying this PR?

Please also add yourself to the AUTHORS file.

rosti-il:
> Unfortunatly there already is a PR (#2536) that also fixes this issue. Seems to be important 😄

Why unfortunately? The project owner should just chose a better solution.

> I am courious if both actually solve the problem completly.

My starting point was a failure of several unit tests on Windows. Now all the tests pass on Windows and no expected result was changed by this PR.

> If I understand the problem correctly there is an additional `\r` at the end.

Right. That additional '\r' character reminds after eliminating the '\n' one and this happens only with CRLF files.

> This PR tries to avoid adding that by removing it early on, the other removes it during writing. Not adding them at all is IMHO the better soloution.

`com.sun.tools.javac.main.JavaCompiler#parse(String fileName)` adds them.
I think it would be wrong converting all source code into LF format on all platforms. Each OS has its own newline encoding standard. This is why Git for Windows configures itself with `core.autocrlf=true` by default.

> Have you tried to compile a non CRLF file on windows? Does that work after applying this PR?

The first version of this PR didn't work correctly with such incorrectly formatted files. I've updated this PR with a better solution. Now it works correctly even with such files.

> Please also add yourself to the AUTHORS file.

Done

Rawi01:
In my opinion its always unfortunate when the limited time of contributors is used for the same thing :smile: 

I think my comment about not adding wrong line breaks was not clear enough. What I meant to say was that there are in general two soloutions for the problem, removing the wrong line breaks in the end or do not add them at all. If I remember correctly your first version tried to add only right line breaks and the other PR removd them in `PrettyPrint`. Your new version now uses both soloutions at the same time which should also be fine.

Have you considered using `dc.split("\r?\n")` instead of the buffered reader?

rosti-il:
> Have you considered using `dc.split("\r?\n")` instead of the buffered reader?

Indeed, this (with double backslashes) allows to make shorter and more readable code. Thank you. I remade my PR respectively.

It's better to be coded `dc.split("\\R")` but than it will not work on Java older than Java 8. Should Lombok support such old versions of Java?

randakar:
It still does, so yes.

There are organisations out there still using versions of java as old as
java 6. I still do not understand why anyone would want to develop software
that far behind the times but somehow it's reality.

rzwitserloot:
awesome work, thank you!

PR #2574, opened 2020-09-02
Use simple name constructor in inner classes

### LINKED ISSUE #2571: [BUG] @With generates non-compiling code under Eclipse
**Describe the bug**
Note that the entire issue is about compilation in Eclipse; I am not able to reproduce the issue via Maven.

@With cannot be used on an inner class where the outer class is generic (I think). An error message such as this is generated when compiled with a recent Java version (used to work on 8):

`Cannot allocate the member type TestCase<T>.Inner using a parameterized compound name; use its simple name and an enclosing instance of type TestCase<T>`

**To Reproduce**
The following code, compiled with Java 11, generates the above error:

```
import lombok.AllArgsConstructor;
import lombok.With;

public class TestCase<T> {
        @AllArgsConstructor
	public class Inner {
		@With
		Object object;
	}
}
```

**Expected behavior**
I expect this to compile under Eclipse, as it used to when running with JDK 8.

 - Lombok version: 1.18.12
 - Eclipse: Version: 2020-06 (4.16.0) / Build id: 20200615-1200 (I do not believe this is relevant, as it occurs with multiple Eclipse versions)

Rawi01:
This PR fixes #2571.

Using `@With` in inner classes generates a constructor using the name of the outer and the inner class. `ecj` does not like that if there also is a generic. To solve this lombok now generates constructor class using raw types: `new Outer<T>.Inner<U>()` -> `new Outer.Inner()`.

### DIFF
--- a/src/core/lombok/eclipse/handlers/EclipseHandlerUtil.java
+++ b/src/core/lombok/eclipse/handlers/EclipseHandlerUtil.java
@@ -921,6 +921,16 @@ public class EclipseHandlerUtil {
 		return result;
 	}
 	
+	public static TypeReference cloneSelfTypeForConstructor(EclipseNode context, ASTNode source) {
+		TypeReference result = cloneSelfType(context, source);
+		if (result instanceof ParameterizedQualifiedTypeReference) {
+			ParameterizedQualifiedTypeReference parameterizedQualifiedTypeReference = (ParameterizedQualifiedTypeReference) result;
+			result = new QualifiedTypeReference(parameterizedQualifiedTypeReference.tokens, parameterizedQualifiedTypeReference.sourcePositions);
+		}
+		if (result != null && source != null) setGeneratedBy(result, source);
+		return result;
+	}
+	
 	public static TypeReference generateParameterizedTypeReference(EclipseNode type, TypeReference[] typeParams, long p) {
 		TypeDeclaration td = (TypeDeclaration) type.get();
 		char[][] tn = getQualifiedInnerName(type.up(), td.name);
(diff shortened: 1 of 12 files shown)


Gaibhne:
I manually built the jar to test it, and both the test case as well as the project the issue originally occurred in compile fine with my Eclipse now. Thank you!

rspilker:
I'm not sure it solves the right problem. In general, having a non-static inner type where the outer type contains generics is asking for compiler problems.

Maybe @rzwitserloot can share his experiences with this problem?

Rawi01:
Is there a case that is not covered by test cases? 

Information about inner classes + inner constructors + generics are quite rare and it seems like that even compilers disagree about how that should be handled 🤷‍♂️

Gaibhne:
For what it's worth: the bug this is addressing is a *regression* - the code that triggers the compiler error without this PR used to compile just fine a while ago.

rzwitserloot:
The problem is, I recently changed lombok to go from generating `new Inner()` to `new Outer.Inner()` to work around _another_ bug. So, apparently, we're stuck between a rock and a hard place. Will investigate

rzwitserloot:
I added all this to fix issue #2268; this revert would mean 2268 is then broken again. We can debate two things:

* Is it even possible to fix both this issue as well as 2268? If not, is ecj just broken here and should we be filing bugs over there instead?
* Failing that, is undoing the fix for 2268 and thereby fixing 2571 the lesser of the two evils?

But.. not for the next release, it'll have to wait.

Rawi01:
@rzwitserloot Reworked the solution, it now uses raw types for the constructor call in eclipse/ecj. I also reverted the `javac` changes because I think that we should not worsen the delombok output to fix this weird ecj behaviour.

PR #2582, opened 2020-09-18
Add missing null check

Rawi01:
At some point I removed this necessary null check...

Obviously the javadoc map can be `null` if lombok does not add any comments. Returning `null` is save here, the original method does that too and lombok only tries to add the comment if that happens. Without this fix eclipse will write an error to the error log everytime someone tries to show the javadoc comment e.g. during code completion.


Bengo7918:
hi :)
can i help with this issue?

Rawi01:
@Bengo7918 This is a pull request that already fixes the issue, a review for this single line change seems a little bit to much so I have no idea how you can help here. If you are interested in supporting this project, I would recommend to start with an easy to reproduce bug and not with a new feature.

rspilker:
@rzwitserloot is/was working on the related code, and he is aware of this pull request.

Rawi01:
@rspilker @rzwitserloot Oh, sorry, I missed that `org.eclipse.jdt.internal.core.CompilationUnit` is not part of `ecj` and that the delegate and javadoc features break ecj compilation. Moving the patch stuff to an eclipse only path and the augments to a different class should at least fix the compilation issue but the test cases are still broken after that because the javadoc comments are not there anymore. Let me know if there is anything I can do to help fixing that.

PR #2591, opened 2020-10-02
Rename TryWithResourcesVarRef

Rawi01:
This test always failed because the `after` file starts with a lower case letter.


rspilker:
Hmm, probably we didn't notice this because Reinier and I  use Mac and Windows.

Issue #2412, opened 2020-03-31
[BUG] Upgrading to android tools gradle plugin 3.6.0+ breaks lombok


darekxan:
**Describe the bug**
it appears there are no property assessors generated.

**To Reproduce**
I've prepared minimal failing build:
https://github.com/darekxan/lomboktest/tree/broken_3.6.1
(sources, build output there)

**Expected behavior**
This one works, version 3.5.3
https://github.com/darekxan/lomboktest/tree/master
(sources, build output there)

**Version info:**
 - Lombok version **v1.18.12**
 - Android Studio 3.6.1 + javac 13.0.1 + kotlin 1.3.71 + kapt3

CharlyLafon37:
It's really annoying. We can't update android gradle plugin in any of our projects because of this. @rzwitserloot @rspilker can you look at this bug please ?

rspilker:
Can you at least point us to the source code of `org.jetbrains.kotlin.kapt3.base.incremental.IncrementalProcessingEnvironment`?

rspilker:
I don't know which is better, 3.5.3 or 3.6.1. I know that 3.5.3 doesn't fail the build. But there is a warning in the log that lombok is disabled. So probably the code doesn't do what you expect.

In your example, you've annotated `TestModel` with `@Data`, but you never call `getTest()`.

rspilker:
I've located the code: https://github.com/JetBrains/kotlin/blob/master/plugins/kapt3/kapt3-base/src/org/jetbrains/kotlin/kapt3/base/incremental/incrementalProcessors.kt

rspilker:
My knowledge of Kotlin is limited, but is I look at the IncrementalProcessinEnvironment and the lombok source code at 
https://github.com/rzwitserloot/lombok/blob/master/src/core/lombok/core/AnnotationProcessor.java#L86
a possible solution is to not only see if there is a field `delegate` but also look for `processingEnv`.

rspilker:
Can you see if the [edge release](https://projectlombok.org/download-edge) fixes this problem?

AngryGami:
Nope, it didn't help. I didn't manage to get plugin (net.ltgt.apt) to work, though I'm not sure it is strictly required. Without this plugin it still complaints about field access inside builder class e.g. like this:
```
@Builder
public class MyConfiguration {
    @Builder.Default
    public final Integer mWidth = 352;

    public static class MyConfigurationBuilder {
[...]

PR #2594, opened 2020-10-04
Find ProcessingEnvironment in InteliJ 2020.3 and Kotlin

### LINKED ISSUE #2592: [BUG] Lombok Does not work with IntelliJ EAP 2020.3 Build 203.4203.26
**Describe the bug**
Compiling projects with Lombok in IntelliJ 2020.3 EAP (Build 203.4203.26) leads to:

>java: You aren't using a compiler supported by lombok, so lombok will not work and has been disabled.
>  Your processor is: com.sun.proxy.$Proxy32
>  Lombok supports: OpenJDK javac, ECJ

**To Reproduce**
Any project compiled in IntelliJ is affected.

 - Lombok version: 1.18.12
 - IntelliJ 2020.3 EAP (Build 203.4203.26)

**Additional context**

IntelliJ is now wrapping the `ProcessingEnvironment` in a `Proxy`. This lead to a problem for us at MapStruct (https://github.com/mapstruct/mapstruct/issues/2215). While trying to fix that and testing with our examples I realized that our examples with Lombok are not compiling anymore.

For us at MapStruct it is quite easy to do this changes, don't know about you at Lombok

Thanks to @bjansen some more information about this:

https://github.com/JetBrains/intellij-community/blob/master/jps/jps-builders-6/src/org/jetbrains/jps/javac/APIWrappers.java

https://github.com/JetBrains/intellij-community/commit/d6f2625c0df5ea7b5be9bad75f5d5c6ecda452d2#commitcomment-42898346

https://youtrack.jetbrains.com/issue/IDEA-250718

Lombok related IntelliJ issue (https://youtrack.jetbrains.com/issue/IDEA-252069)

Rawi01:
This PR fixes #2592. 
I also completed the potential fix for #2412 by @rspilker (he missed to add the additional field to `LombokProcessor`). I don't know why the code to search the processing environment is there at two places, thats why I have not moved it to a different class.

### DIFF
--- a/src/core/lombok/core/AnnotationProcessor.java
+++ b/src/core/lombok/core/AnnotationProcessor.java
@@ -26,8 +26,9 @@ import static lombok.core.Augments.ClassLoader_lombokAlreadyAddedTo;
 import java.io.File;
 import java.io.PrintWriter;
 import java.io.StringWriter;
-import java.lang.reflect.Field;
+import java.lang.reflect.InvocationHandler;
 import java.lang.reflect.Method;
+import java.lang.reflect.Proxy;
 import java.net.URL;
 import java.util.ArrayList;
 import java.util.Arrays;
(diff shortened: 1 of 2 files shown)


src/core/lombok/core/AnnotationProcessor.java
```diff
@@ -100,6 +97,40 @@ private static ProcessingEnvironment tryRecursivelyObtainJavacProcessingEnvironm
...
+		}
+	}
+	
+	/**
+	 * InteliJ >= 2020.3
+	 */
+	private static Object tryGetProxyDelegateToField(Class<?> delegateClass, Object instance) {
+		try {
+			InvocationHandler handler = Proxy.getInvocationHandler(instance);
+			return Permit.getField(handler.getClass(), "val$delegateTo").get(handler);
```
eugenezh, line 128: I would suggest making this heuristic a bit more reliable and not depending on particular field name. Also all Proxy-related suff will remain hidden. 
Consider using the _org.jetbrains.jps.javac.APIWrappers.WrapperDelegateAccessor_ interface.

First, check if the wrapped object is instanceof _APIWrappers.WrapperDelegateAccessor_; 
then use the _APIWrappers.WrapperDelegateAccessor.getWrapperDelegate()_ to access the wrapped object.
I have made the _WrapperDelegateAccessor_ public for this purpose.

eugenezh: Similar logic should work for accessing wrapped ProcessingEnvironment too.

Rawi01: Thanks for the feedback, I tried to do it that way but it does not work. For compilation I created a stub for `WrapperDelegateAccessor`. At build time the class should be already there but lombok seems to be unable to find it and is unable to perform the cast and the method call.

eugenezh: Ah, indeed, this is because JPS classes are deliberately isolated in a separate classloader from annotation processors. This is needed to avoid possible conflicts between libraries that use annotation processors and libraries used by JPS.
Then we can stick with the same idea, but use reflection to access the delegate like this:

Class<?> accessorInterface = Class.forName("org.jetbrains.jps.javac.APIWrappers$WrapperDelegateAccessor", true, proxy.getClass().getClassLoader());
Method accessor = accessorInterface.getMethod("getWrapperDelegate");
Object delegate = accessorInterface.isInstance(proxy)? accessor.invoke(proxy) : null;

The code assumes the "proxy" reference points to the proxy wrapper

PR #2595, opened 2020-10-04
The Object Construction Checker has been renamed the Called Methods Checker

mernst:
This pull request updates the names of annotations used by the Checker Framework.
The old versions continue to be supported by the Checker Framework, but it's better to use the current version.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -24,6 +24,7 @@ Maarten Mulders <mthmulders@users.noreply.github.com>
 Mark Haynes <markhaynes.work@gmail.com>
 Mart Hagenaars <marthagenaars@gmail.com>
 Mateusz Matela <mateusz.matela@gmail.com>
+Michael Ernst <mernst@alum.mit.edu>
 Michiel Verheul <cheelio@gmail.com>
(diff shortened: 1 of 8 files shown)


Rawi01:
I think that @rzwitserloot fixed the same thing 20 hours ago together with my messed up eclipse patches. I have not verified that it is exactly the same, can you do that?

mernst:
It is not the same.  I started my pull request from head, after @rzwitserloot's changes (which I didn't notice until I started to make my changes).

Rawi01:
Ah, fascinating, he changed that one recently from `org.checkerframework.checker.builder` to `org.checkerframework.checker.objectconstruction` but I think that you are right and `org.checkerframework.checker.calledmethods` is the right annotation. At least I found [some documentation](https://checkerframework.org/api/org/checkerframework/checker/calledmethods/qual/CalledMethods.html) online. 

Even if you already contributed some code you are not listed in the AUTHORS file, can you add yourself?

mernst:
> Even if you already contributed some code you are not listed in the AUTHORS file, can you add yourself?

I have done so.

PR #2602, opened 2020-10-09
Fix inconsistency in lombok.toString.callSuper doc

lars-sh:
As of line 275 the default value for `lombok.toString.callSuper` is `skip`, while line 177 states the default to be `warn`. As far as I read the code `skip` is the correct default here.


rspilker:
Thanks. Since this is only a change in documentation, You're not required to add your name to the AUTHORS file.

PR #2609, opened 2020-10-14
ToString output of field encryption support

narusas:
Hi

I need to remove security text of log using Lombok `@Data`. 

So I make an extension  `@ToString.Secured` on field. 

Encryption mechanism is variable on site, project. So I make encryption decide by lombok.config or annotation. 

please review. If I need modify, let me know. 

Thank your Lombok for year!


rspilker:
We've discussed this in a feature request in the past, and don't see much value for this over the current features of `@ToString`.

We don't expect a lot of users having a lot of fields that they want to have encrypted in the `toString` result.

The current alternative:
```java
@ToString
class User {
    String name;
    String secret;

    @ToString.Include(name="secret") // Override the field
    private hashedSecret() {
        MyHasher.hash(secret);
    }
}
```

Or 

```java
@ToString
class User {
    String name;
    @ToString.Exclude
    String secret;

    @ToString.Include(name="redactedSecret")
    private hashedSecret() {
        MyHasher.hash(secret);
    }
}
```

PR #2634, opened 2020-10-29
#2633 Split PatchDelegate into 2 classes, 1 without jdt.core deps.

### LINKED ISSUE #2633: [BUG] Regression: #2006 breaks Delegate in Tycho
**Describe the bug**
Inclusion of the new method PatchDelegate.getChildren(IJavaElement[], SourceTypeElementInfo) from #2006 and according implementation with imports from org.eclipse.jdt.internal.core break compatibility with Tycho (which uses ECJ, didn't test with plain ECJ) because the packages are not available there. 
Loading from lombok.eclipse.agent.PatchDelegatePortal.Reflection throws ClassNotFoundExceptions (with -Dlombok.debug.reflection=true).
Thus @Delegate doesn't work at all with the latest Lombok on Tycho.

**To Reproduce**
Reproduction would need my changes for Tycho, but should be reproducable with ECJ only as well.

**Expected behavior**
Delegates should be working on Tycho/ECJ as well.

 - 1.18.16
 - Tycho/ECJ >=1.5.0

**Additional context**
@Rawi01 I think it should work if you can put the additions to a new class and load it seperately in PatchDelegatePortal.

Wouldn't it be good to show the Exceptions when Delegate is actually used. Swallowing makes sense when it's not used, but in this case it took me a long time to figure out what was wrong because I didn't see the exception without the lombok.debug.reflection switch.

rgra:
Fixes #2633 by splitting up PatchDelegate into PatchDelegateChildren and delegating to it via an interface

### DIFF
--- a/src/eclipseAgent/lombok/eclipse/agent/IPatchDelegateChildren.java
+++ b/src/eclipseAgent/lombok/eclipse/agent/IPatchDelegateChildren.java
@@ -1,0 +1,43 @@
+/*
+ * Copyright (C) 2010-2020 The Project Lombok Authors.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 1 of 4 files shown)


Rawi01 (commented):
- There should be a formatter profile if you import lombok in eclipse, can you use that one? Your current profile removed a lot of whitespace.
- New files should use only 2020 as copyright year.
- `findAlreadyImplementedMethods` is not an eclipse only method, disabling this in ecj mode will lead to inconsistent behaviour.
- I don't think that it is required to split up `Reflection` into two classes as long as `getChildren` is the second method in the static block.
- I don't think that it is required to split everything up into two classes, you can use the non existing class as long as you do not create an instance and do not looup the methods using reflection. It might be easier to move all eclipse only methods into a nested class e.g. `EclipseOnlyMethods` that detects if eclipse is available and if not all methods simply return `null`. Example:
```java
private static class EclipseOnlyMethods {
	private static boolean eclipseAvailable = true;
	static {
		try {
			CompilationUnit.class.getName();
		} catch (NoClassDefFoundError e) {
			eclipseAvailable = false;
		}
	}
	private static List<SourceMethod> getDelegateMethods(SourceType sourceType) {
		if (!eclipseAvailable) return null;
		...
	}
	...
}
```

rzwitserloot:
@Rawi01 Given that you wrote in already-merged PR #2742 that that one 'fixes' this one, can I close this without merging it?

Rawi01:
@rzwitserloot Yes, this PR can be closed. I missed to do that, sorry for wasting your time.

PR #2637, opened 2020-11-02
Write classes only once

Rawi01:
This PR fixes https://github.com/mplushnikov/lombok-intellij-plugin/issues/969.

In
https://github.com/eclipse/eclipse.jdt.core/blob/b3cc25e26a9b327ef6e79fbec3cde4d4fbeb60fc/org.eclipse.jdt.compiler.tool/src/org/eclipse/jdt/internal/compiler/tool/EclipseCompilerImpl.java#L444-L448
the stream gets closed twice. By wrapping the `ByteArrayOutputStream` in a `FilterOutputStream` the second close no longer perform any action.

@rzwitserloot After merging your latest Tycho changes I get a `java.lang.IllegalArgumentException: argument type mismatch` error. InteliJ does not print a full stacktrace so I do not know what exactly is broken.

### DIFF
--- a/src/core/lombok/core/PostCompiler.java
+++ b/src/core/lombok/core/PostCompiler.java
@@ -22,6 +22,7 @@
 package lombok.core;
 
 import java.io.ByteArrayOutputStream;
+import java.io.FilterOutputStream;
 import java.io.IOException;
 import java.io.OutputStream;
 import java.io.PrintWriter;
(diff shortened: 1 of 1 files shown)


mplushnikov:
@rspilker @rzwitserloot Can you please merge this pr?

rzwitserloot:
As I understand it, you're wrapping the BAOS in a `FilterOutputStream` because the impl of FOS has a volatile boolean-based system to ensure a close() call is only propagated to the wrapped OutputStream exactly once.

Unfortunately, a quick check of the JDK sources shows that this is not how FOS worked in java7. The volatile boolean was introduced somewhere between 7 and 14.

Given that we are still targeting some fairly old versions, I think we need to handroll this boolean. Shouldn't be a problem, it's still only like 5 lines of code. I'll set it up tonight, just informing you that the thing this PR intends to fix will be fixed, but not via this PR.

Anything I missed?

Rawi01:
@rzwitserloot That's exactly why I wrapped it in a `FilterOutputStream`. I have not checked the older versions because the current comment described that behaviour and I thought that it behaves like that since Java 1...

mkurz:
Any chance you update the edge release?

rzwitserloot:
@mkurz just released a new edge, with this fix.

mkurz:
Thanks!

PR #2638, opened 2020-11-03
@Builder annotation in a class with defined constructor.

caiocruz:
This change makes builder processor to find an all argument matching constructor before skip constructor generation.

So it opens the possibility of having a `@Builder` along with `@RequiredArgsConstructor` or `@NoArgsConstructor` annotations, as well as user defined constructor methods.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -4,6 +4,7 @@ Adam Juraszek <juriad@gmail.com>
 Aleksandr Zhelezniak <lekan1992@gmail.com>
 Andre Brait <andrebrait@gmail.com>
 Bulgakov Alexander <buls@yandex.ru>
+Caio Cruz <acruz.caio@gmail.com>
 Caleb Brinkman <floralvikings@gmail.com>
 Christian Nüssgens <christian@nuessgens.com>
 Christian Schlichtherle <christian-schlichtherle@users.noreply.github.com>
--- a/src/core/lombok/eclipse/handlers/EclipseHandlerUtil.java
+++ b/src/core/lombok/eclipse/handlers/EclipseHandlerUtil.java
@@ -1903,21 +1903,27 @@ public class EclipseHandlerUtil {
 	 *            The parameter list to match the constructor
 	 */
 	public static MemberExistsResult constructorExists(EclipseNode node, List<EclipseNode> fieldParams) {
+
 		List<ConstructorDeclaration> constructors = getConstructors(node);
+		int params = fieldParams.size();
 		for (ConstructorDeclaration constructor : constructors) {
-			int constructorArgumentsNumber = 0;
-			if (constructor.arguments != null) {
-				constructorArgumentsNumber = constructor.arguments.length;
-			}
-			if (constructorArgumentsNumber == fieldParams.size()) {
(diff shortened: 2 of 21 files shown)


rzwitserloot:
Still reviewing but, @caiocruz can you add your name to the AUTHORS file? Note the legal notice at the end of that file :)

rzwitserloot:
Hmm, I'm not sure this is going to fly. You can't compare types like this. Lombok has no resolution, thus, there is no way to know for sure that any given constructor could fit. Here are some examples where lombok stands no chance, or it does but it gets complicated quickly (and your code doesn't get it right):

```
class Example1 {
    String a, b, c, d;

   Example1(String... args) {}
}

class Example2 {
    String a;

    Example2(Object a) {}
}

class Example3<T extends Number, V extends T> {
    V a;

    Example3(Number b) {}
}

class Example4<T extends Number, V extends T> {
    V a;

    Example4(T b) {}
}

// This is where it gets extremely problematic
class Example5<T extends Number, V extends T> {
    List<? extends V> a;

    Example3(List<Integer> b) {}
}
```

For example5, if lombok generates an all-args, then you get a conflict constructor issue (2 constructors that are signature compatible, thus, illegal), and yet the constructor that does exist isn't compatible. Optimally lombok should give you an error message that explicitly explains the problem, but even if you want to go there, trying to figure out if a given parameter's type is compatible with the field, once you involve generics, is a giant can of worms.

I think this PR needs to be abandoned :( it is vastly more complicated than this.

But there is one way out: We already use a halfway between 'just by name' and 'by full signature type match' concept elsewhere: aritycount.

The arity of a method is the # of arguments that you can feasibly pass to it. For example, some arities:

```
void foo(String a, String b); // This is foo:2
void bar(String... c); // this is baz: all arities.
void baz(Object a, Object b, String... c); // this is baz, arities 2-*.
```

We can do some arity counting and if arity counting concludes that no constructor exists that covers the required arity, then lombok generates, otherwise it won't. This still lets you stick a `@NoArgsC` on things without stopping builder. It covers 99% of all cases.

This arity counting is already used in `methodExists` (see the `int params` argument of that method); constructorExists can be trivially expanded to function similarly.

Now that I think of it, this code needs a fix. This is legal these days:

```
class Example {
    void foo(Example this) {}
}
```

the arity of the `foo` method is, in fact, 0. That's a 'fake' argument you can use if you want to annotate the receiver, but otherwise it has no meaning.

rzwitserloot:
Hmm, about that `Example this` thing, the ASTs don't add it to the param list, so nevermind. No changes needed there.

What do you think about the arity counting idea?

caiocruz:
@rzwitserloot Thanks for the feedback! I'll try out arity counting idea.

rzwitserloot:
@caiocruz Hey Caio, are you still working on this, or would you prefer if we abandon it, or is the idea of arity-counting as a compromise acceptable, but you're not going to get around to it anytime soon?

I think we can look into adding it to our todo list, but it'd be a waste if you're also planning on working on it :)

caiocruz:
Hey @rzwitserloot, very sorry for huge delay.
Yes I plan to work on it in next weeks.

PR #2658, opened 2020-11-25
Reset inference context for extension methods

### LINKED ISSUE #2648: [BUG] ExtensionMethod for generic types does not work with generic parameters in ecj
**Describe the bug**
Adding extension methods for generic types seems to be broken if there also is a generic parameter. I think I broke this while working a bunch of other bugs. I will try to fix this myself...

**To Reproduce**
```java
import java.util.Arrays;
import java.util.List;

import lombok.experimental.ExtensionMethod;

@ExtensionMethod(ExtensionMethodGeneric.Extensions.class)
class ExtensionMethodGeneric {
	public void test() {
		List<String> stringList;
		List<Number> numberList;
		
		stringList.test(numberList);
	}
	
	static class Extensions {
		public static <T> void test(List<String> obj, List<T> list) {
			
		}
	}
}
```
```
12 ERROR The parameterized method <String>test(List<String>, List<String>) of type ExtensionMethodGeneric.Extensions is not applicable for the arguments (List<String>, List<Number>)
```

**Expected behavior**
It compiles without any error as it does if you call the extension method by hand.

**Version info:**
 - Lombok version: **1.18.16**
 - Platform: **Eclipse 2020-09**

**Additional context**
This is a simplified example, there might be more problems if you use more complex generic stuff.

Rawi01:
This PR fixes #2648.

In Java >= 8 ecj uses the inference context to resolve the generic information. This one is already set before lombok tries rewrite the method call. Simply copying the information does not cover all the different cases but reseting the inference contexts and running type inference again does.

### DIFF
--- a/src/eclipseAgent/lombok/eclipse/agent/PatchExtensionMethod.java
+++ b/src/eclipseAgent/lombok/eclipse/agent/PatchExtensionMethod.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2012-2014 The Project Lombok Authors.
+ * Copyright (C) 2012-2020 The Project Lombok Authors.
  * 
  * Permission is hereby granted, free of charge, to any person obtaining a copy
  * of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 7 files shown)


rzwitserloot:
Ooh, that's a much better way to solve this problem than me writing tons of code to try to reproduce all the various TypeBinding objects. This feels like one of those 'saves you an entire coding week's worth of effort' PRs. All PRs are very welcome but some PRs are _very_ welcome indeed.

Feels significant enough for a changelog entry, I'll add a commit in a bit for it.

Issue #2645, opened 2020-11-13
[BUG] lombok.config files aren't read in when compiling with sbt 1.4


rzwitserloot:
**Describe the bug**
Since about sbt 1.4, there's an incremental compilation thing happening with sbt which is causing the `sourcefile` field of javac's CompilationUnits to report a virtualized URI, which lombok cannot turn into a file path, thus causing lombok to just give up on trying to find the directory to start the chain of looking for `lombok.config` files.

**To Reproduce**
Compile 'Hello, World!' in a java file with sbt while you have a `lombok.config` file present; lombok will completely ignore that config file.

**Additional context**
The URI reported by sbt is something like `vf://tmp/$%7BBASE%7D/src/main/java/example/Main.java`

See also [Pull Request 2643](https://github.com/rzwitserloot/lombok/pull/2643)

raphaelNguyen:
Thanks for fixing this up. Can I ask if there is a release scheduled that would contain this change some time soon?

mkurz:
@rzwitserloot You [fix](https://github.com/rzwitserloot/lombok/commit/e1f82ac4d132769cfc272dccfc916aeba7181718) for this issues is not complete: We ran into problems when using sbt subprojects. I provide a patch here: #2668 Please have a look (and merge :wink) before the next release, thanks!

domdorn:
Any estimate when a new release will be published? This is currently blocking me from upgrading sbt 1.3.12 to 1.4.5. Thanks!

rzwitserloot:
@domdorn are you looking for an edge release or lombok v1.18.18?

mkurz:
The edge release does not yet contain my fix for sbt subprojects (#2668). But it does work already for projects where Lombok is NOT used in a sub-project.

mkurz:
@rzwitserloot Would be nice if you could update the edge release to latest master :wink:

rzwitserloot:
Tonight probably.

PR #2668, opened 2020-12-05
Fixing reading files in sbt 1.4+ subprojects

mkurz:
I did some testing with the current edge release, and found a bug:
[The fix](https://github.com/rzwitserloot/lombok/commit/e1f82ac4d132769cfc272dccfc916aeba7181718) for #2645 is not working when using [sbt subprojects](https://www.scala-sbt.org/1.x/docs/Multi-Project.html).
In my case I have a subproject, which pulls in and uses lombok, and a parent (root) project (which depends on the subproject).
Now when running sbt in the root project and start compiling that root project, which in turn also compiles the subproject, the `lombok.config` (located in the subproject) is, again, ignored, resulting in compiling errors in the subproject.

After some debugging it turns out that the URI reported by sbt for the subproject has the format:

```
vf://tmp//home/mkurz/myproject/src/main/java/example/Main.java
```

instead of the format

```
vf://tmp/$%7BBASE%7D/src/main/java/example/Main.java
```

when using a normal, single root project with lombok.

That makes sense, because the files of the subproject are obviously not located within the root project (=`BASE`, where relative paths like `${BASE}/src/main/java/example/Main.java` are used), therefore absolut paths are used (`/home/mkurz/myproject/src/main/java/example/Main.java`).

The fix is quite easy: When the file path doesn't start with `${` (when looking for `${BASE}`), we can assume the path is very likely an absolute one, so we just use it.

I can confirm this fixes our compiler errors and the app runs smoothly.

### DIFF
--- a/src/core/lombok/javac/JavacAST.java
+++ b/src/core/lombok/javac/JavacAST.java
@@ -174,7 +174,14 @@ public class JavacAST extends AST<JavacAST, JavacNode, JCTree> {
 		if (sbtMappedVirtualFileRootsField == null) return null;
 		
 		String encodedPath = (String) sbtMappedVirtualFilePathField.get(mappedVirtualFile);
-		if (!encodedPath.startsWith("${")) return null;
+		if (!encodedPath.startsWith("${")) {
+			File maybeAbsoluteFile = new File(encodedPath);
+			if (maybeAbsoluteFile.exists()) {
+				return maybeAbsoluteFile.toURI();
+			} else {
+				return null;
+			}
+		}
 		int idx = encodedPath.indexOf('}');
 		if (idx == -1) return null;
 		String base = encodedPath.substring(2, idx);


src/core/lombok/javac/JavacAST.java
```diff
@@ -174,7 +174,14 @@ private static URI tryGetSbtFile_(JavaFileObject sourcefile) throws Exception {
 		if (sbtMappedVirtualFileRootsField == null) return null;
 		
 		String encodedPath = (String) sbtMappedVirtualFilePathField.get(mappedVirtualFile);
-		if (!encodedPath.startsWith("${")) return null;
+		if (!encodedPath.startsWith("${")) {
+			File maybeAbsoluteFile = new File(encodedPath);
+			if (maybeAbsoluteFile.exists()) {
```
mkurz, line 179: Not sure if the check for existence is really necessary, or if we just should assume it's the absolute path and just `return new File(encodedPath);`

mkurz:
@rzwitserloot Can you please have a look at this pull request before the next release? The [sbt fix you provided](https://github.com/rzwitserloot/lombok/commit/e1f82ac4d132769cfc272dccfc916aeba7181718) unfortunately does not handle sbt subprojects yet. Thanks!

PR #2684, opened 2020-12-21
Fix NullPointerException during Javadoc copy

### LINKED ISSUE #2682: [BUG] NullPointerException during Javadoc copy
**Describe the bug**
Lombok does not check if the source javadoc is `null`. This leads to a bunch of NPE as soon as lombok tries to create a regex matcher. The exception gets catched in one of the calling methods and does not lead to a compilation error.

**To Reproduce**
- Add a breakpoint for thrown NullPointerExceptions
- Run tests

**Expected behavior**
Obviously there should be no exception. I also noticed that there are javadoc copy methods in `HandlerUtil` that only add new javadoc lines (e.g. `addReturnsThisIfNeeded`). Right now these methods get never called if the input is empty because of the exception. Is this intended? If not, lombok could also create the javadoc, I added an example how the result could look like. 

**Input**
```java
@lombok.experimental.Accessors(chain = true, fluent = true)
@lombok.Setter
class Example {
	private int fieldName;
}
```

**Current output**
```java
class Example {
	private int fieldName;

	public Example fieldName(final int fieldName) {
		this.fieldName = fieldName;
		return this;
	}
}
```

**Possible output**
```java
class Example {
	private int fieldName;

	/**
	 * @return {@code this}.
	 */
	public Example fieldName(final int fieldName) {
		this.fieldName = fieldName;
		return this;
	}
}
```

 - Lombok version: **latest**
 - Platform: **all**

**Additional context**
I can provide a PR if you let me know how lombok should handle these cases.

Rawi01:
This PR fixes #2682.

It is splitted up into two commits:
- The first one adds back support for javadoc in ecj tests. Instead of adding a `FieldAugment` for the concrete implementation I added a new `InterfaceFieldAugment` that supports interfaces. I also replaced the unavailable dependecies `Signature` and `IJavaElement`.
- The second one adds a some null checks and now always adds the `@return` statements as discussed in the referenced issue.


rzwitserloot:
That hits way more tests than I thought it would, but, yup. Great work!

Issue #375, opened 2015-07-14
@InheritConstructors to inherit constructors (especially useful for Exception classes)


lombokissues:
_Migrated from Google Code (issue 302)_

almson:
+1 I hate the boilerplate of rewriting constructors. It would be a great, simple addition.

yuankui:
+1

elexx:
+1

o-nix:
+1

tmarsteel:
+1

BrainStone:
What's the current state on this?

Quantum64:
+1 😁

Emaxoso:
+1

typingduck:
👍

naitonium:
+1

Tharsalbird:
+1

pufferbatterie:
+1

debanshuk:
+1

symphony-thibault:
+1

lusoalex:
+1

fowlie:
+1

sbnarra:
+1 - for the most part my exceptions look like so....

`public class MyException extends Exception {
    public MyException(String msg, Exception e) {
        super(msg, e);
    }

    public MyException(String msg) {
        super(msg);
    }
}`

developerrajtomar:
+1

pbhasker:
+1

buiatte:
+1

sorgloomer:
A low hanging fruit might be an @ExceptionConstructors annotation that would add the five constructors that Exceptions usually have. I think that would cover 95% of peoples needs (also mine) and doesn't require any sophisticated introspection.

janrieke:
There's also a discussion on this topic in the lombok forums: https://groups.google.com/forum/#!topic/project-lombok/x1DdJvG-PKc

My impression is that the lombok maintainers like the idea, but do not have it very high on their priority list. So if someone wants to take it, you should post in that thread and go for it.

erekose255:
+1

YanzheL:
+1

diepet:
+1

zhanghuabin:
+1

gregoavg:
+1

Kamil-Krynicki:
+1

nickmafra:
+1

daniloarcidiacono:
+1

FoKas90:
+1

[...]

PR #2702, opened 2021-01-06
@StandardException annotation for generating exception constructors

### LINKED ISSUE #1559: How to use lombok to create custom exception faster.
I need to create some custom exception which extends RuntimeException like:

    @NoArgsConstructor
    public class DuplicateException extends RuntimeException {
        private static final long serialVersionUID = 2610477518423283258L;

        public DuplicateException(String message) {
            super(message);
        }

        public DuplicateException(String message, Throwable cause) {
            super(message, cause);
        }

        public DuplicateException(Throwable cause) {
            super(cause);
        }
    }

They are basically the same, but some is subclass of this.

How to create them faster using lombok?

ttzn:
Fixes #1559, related to #375 and discussed in [this thread](https://groups.google.com/u/2/g/project-lombok/c/x1DdJvG-PKc/m/zIbn1aLaCAAJ).

This is a pretty advanced attempt at implementing the proposed `@StandardException` feature (naming to be discussed). It turns the following:

```java
@StandardException
public class ExampleException extends Exception {
}
```

... into:

```java
public class ExampleException extends Exception {
    public ExampleException() {
    }
    public ExampleException(String message) {
        super(message);
    }
    public ExampleException(Throwable cause) {
        super(cause);
    }
    public ExampleException(String message, Throwable cause) {
        super(message, cause);
    }
}
```

If one of those 4 constructors is already declared, the handlers skip it. Naturally, compilation fails if any of the parent constructors is missing, as this is really meant to be used on subclasses of JDK throwables. Right now there is no configuration option to control the behaviour.

Before going further with cleanup and documentation, I'd like this to be thoroughly discussed. Some words on the implementation:
* a huge portion of the code is copy-pasted from the `HandleConstructors` classes, with lots of cuts and modifications; some degree of factorization may be achieved by moving stuff into utility classes, which I plan to do if this feature is validated (along with a more comprehensive test suite);
* I had to add `stringType` and `throwableType` in the javac `Symtab` stub, as I found no way to reference these types using available APIs.

I hope this proves useful. Thanks to the maintainers for all the hard work by the way.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -2,6 +2,7 @@ Lombok contributors in alphabetical order:
 
 Adam Juraszek <juriad@gmail.com>
 Aleksandr Zhelezniak <lekan1992@gmail.com>
+Amine Touzani <ttzn.dev@gmail.com>
 Andre Brait <andrebrait@gmail.com>
 Bulgakov Alexander <buls@yandex.ru>
 Caleb Brinkman <floralvikings@gmail.com>
--- a/src/core/lombok/ConfigurationKeys.java
+++ b/src/core/lombok/ConfigurationKeys.java
@@ -695,4 +695,10 @@ public class ConfigurationKeys {
 	 */
 	public static final ConfigurationKey<CheckerFrameworkVersion> CHECKER_FRAMEWORK = new ConfigurationKey<CheckerFrameworkVersion>("checkerframework", "If set with the version of checkerframework.org (in major.minor, or just 'true' for the latest supported version), create relevant checkerframework.org annotations for code lombok generates (default: false).") {};
 
+	/**
+	 * lombok configuration: {@code lombok.standardException.flagUsage} = {@code WARNING} | {@code ERROR}.
+	 *
+	 * If set, <em>any</em> usage of {@code @StandardException} results in a warning / error.
+	 */
+	public static final ConfigurationKey<FlagUsageType> STANDARD_EXCEPTION_FLAG_USAGE = new ConfigurationKey<FlagUsageType>("lombok.standardException.flagUsage", "Emit a warning or error if @StandardException is used.") {};
 }
--- a/src/core/lombok/StandardException.java
+++ b/src/core/lombok/StandardException.java
@@ -1,0 +1,34 @@
+/*
+ * Copyright (C) 2010-2017 The Project Lombok Authors.
+ * 
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ * 
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ * 
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package lombok;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ */
+@Target(ElementType.TYPE)
+@Retention(RetentionPolicy.SOURCE)
+public @interface StandardException {
+}
--- a/src/core/lombok/eclipse/handlers/HandleStandardException.java
+++ b/src/core/lombok/eclipse/handlers/HandleStandardException.java
@@ -23,16 +23,16 @@ package lombok.eclipse.handlers;
 
 import lombok.AccessLevel;
 import lombok.ConfigurationKeys;
-import lombok.StandardException;
+import lombok.experimental.StandardException;
 import lombok.core.AST.Kind;
 import lombok.core.AnnotationValues;
 import lombok.eclipse.Eclipse;
 import lombok.eclipse.EclipseAnnotationHandler;
 import lombok.eclipse.EclipseNode;
 import lombok.eclipse.handlers.EclipseHandlerUtil.*;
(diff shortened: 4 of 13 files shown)


esend7881:
I thought this would have been part of the new 1.18.18 version but don't see it in there.... When will `@StandardException` get introduced into lombok, I'd like to use it.

rzwitserloot:
I like it, but we're on the hot seat to push 1.18.20 out the door. Only real hold up is a thorough review. @rspilker - what are you thoughts on the feature itself?

TODO:
* @ttzn did you add your name to AUTHORS?
* docs - this needs a website/template/StandardException.html page with examples and a tag line.
* We need to consider if this should go into experimental instead, to give it some time: That gives us a window to break how it works later on (whereas if we release it in the lombok package, that door is mostly closed).

ttzn:
@rzwitserloot treating this as experimental seems fair, so I moved it. I also rebased, added my name and started documentation.

rzwitserloot:
Reviewing the basic feature idea itself, I see some problems.

Some exceptions simply do not have the 'set of 4', so implementing all 4 constructors as, respectively:

* `super();`
* `super(msg);`
* `super(cause);`
* `super(msg, cause);`

will fail for these. Admittedly a lot of exceptions are growing this 'standard set', but there are still plenty that just do not have them. For example, [`java.lang.NumberFormatException`](https://github.com/openjdk/jdk16/blob/master/src/java.base/share/classes/java/lang/NumberFormatException.java) still does not have them, even in JDK16.

I'm not quite sure why you'd want to extend that class, but you can (it is not `final`), and it would be surprising if `@StandardException public class MyNumberProblemException extends NumberFormatException {}` results in weird 'superconstructor not found' compiler errors. Lombok cannot resolve these types so cannot check if the expected 'set of 4' isn't there. Lombok can't even tell if the parent type is some sort of throwable in the first place, unfortunately. I'm willing to overlook that one, though.

A second issue is that you may want to do stuff in addition to the standard super call.  By implementing the set-of-4 with a `super` call we make this impossible, and make the `@StandardException` feature unavailable for such cases.

We can fix both of these problems in one fell swoop but that brings in new problems.

Proposal:

The set-of-4 is implemented as follows (so, this is the delomboked version of `@StandardException class Foo extends Throwable {}`:

```
public class Foo extends Throwable {
    public Foo() {
        this(null, null);
    }
    public Foo(String msg) {
        this(msg, null);
    }
    public Foo(Throwable cause) {
        // This is duplicating the default behaviour of Throwable itself.
        this(cause == null ? null : cause.getMessage(), null);
    }
    public Foo(String msg, Throwable cause) {
        super(msg);
        if (cause != null) super.initCause(cause);
    }
}
```


This has considerable advantages:

* If you want to perform some extra operations in your exception class, you can do so - by only manually inserting the full-args constructor. Lombok will notice it and simply not generate that one. The other 3 will then invoke the one you wrote. Convenient.
* By default this works for virtually all extant Throwable subtypes, in java core and many third party libraries. They virtually always have the `(String msg)` constructor, and `initCause` is from Throwable itself, so that'll be there.

It also has disadvantages:
* It is slightly less standard, even if more convenient.
* __It is functionally slightly different!__ `super(msg, null)` and `super(msg)` are __not the same__: The second call means you can later call `initCause`, whereas the first one (passing `null` for cause) locks away the ability to set the cause later with `initCause`, forever. Therefore the unfortunate upshot of the above example is that our code is very slightly different in the following manner: If you invoke e.g `new IOException(msg, null)`, then you have locked the ability to `initCause` permanently. However, in our code, invoking `new MyCustomExMadeWithLombok(msg, null)`, then `initCause` can still be invoked.


I think in the grand scheme of things this is the best available option, but I could be convinced there are better alternatives.

Thoughts?

rzwitserloot:
Also, there is room for an expansion to the feature (perhaps better left as a second milestone and not for a first release):

What if we make it possible to annotate a _constructor_ in a class with `@StandardException`. This then means: There must be both a `String message` and `AnyType cause` parameter, with those exact names; if either one is not there it's a compile-time lombok error. If they are there, lombok will generate the 3 missing constructors, derived by removing first the `message` param, then the `cause` param, then both, and they are implemented in the exact same fashion: a `this` call to the annotated constructor, with the caveat that the message-less, cause-containing variant uses `cause == null ? null : cause.getMessage()`.

This would let you make constructors that take additional arguments.

Admittedly vastly more exotic than the usual 'just gimme the standard batch of 4 please' strategy.

rzwitserloot:
I've reviewed and implemented the above changes (but not the 'annotate an existing constructor' suggestion). @rspilker still needs to review the feature, though. I'll try my best to make the case for this feature :)

rzwitserloot:
@ttzn I had to refactor quite a bit but please don't take that as an indicator of quality of the PR - your PR was a welcome sight and helped a ton! Thank you!

rzwitserloot:
One more update I'd like to put in: What should lombok do if the annotated class extends nothing at all? E.g:

```
@StandardException
public class MyException {}
```

The two obvious actions are:
* Do not change anything, and put an error message on the annotation stating that you must extend some sort of Throwable type.
* Inject `extends Exception`, for the highest levels of convenience.

Ordinarily I always prefer 'do not presume to know what the developer wants if there are multiple plausible answers, and instead generate an error that explains precisely what needs to be done', but there is a case to be made that 'just extend Exception' is convenient, and it does mean you can just write `@StandardException public class Whatever{}` which is such a pretty one-liner.

Thoughts?

esend7881:
> One more update I'd like to put in: What should lombok do if the annotated class extends nothing at all? E.g:
> 
> ```
> @StandardException
> public class MyException {}
> ```
> 
> The two obvious actions are:
> 
> * Do not change anything, and put an error message on the annotation stating that you must extend some sort of Throwable type.
> * Inject `extends Exception`, for the highest levels of convenience.
> 
> Ordinarily I always prefer 'do not presume to know what the developer wants if there are multiple plausible answers, and instead generate an error that explains precisely what needs to be done', but there is a case to be made that 'just extend Exception' is convenient, and it does mean you can just write `@StandardException public class Whatever{}` which is such a pretty one-liner.
> 
> Thoughts?

To make this logic work, perhaps making a nearly equivelent `@StandardRuntimeException` which is similar to `@StandardException` but instead extends RuntimeException.

I haven't tried to implement this, but I'm sure there is a way to do this with minimal repeated code. I suppose we could have

- @StandardThrowable 
- @StandardRuntimeException 
- @StandardException 

Where @StandardThrowable contains most of the logic and @StandardException or @StandardRuntimeException contain the specifics related to extending Exception or RuntimeException. 

With this in mind, a user can either extend Exception/RuntimeException or not extend Exception/RuntimeException, it'd be a redundant statement.

rspilker:
@esend7881 If I would choose between having the different annotations, or having the author type an extends clause, the latter would always win.

esend7881:
> @esend7881 If I would choose between having the different annotations, or having the author type an extends clause, the latter would always win.

I actually agree, so the user should manually specify the exception superclass. That actually makes more sense because a user could want to extend a specific exception such as IOException or IllegalArgumentException, etc.

martianch:
Can you please finally build and release Lombok with this feature?

rzwitserloot:
@martiench release is imminent.

NSnietol:
Hi,
 is this feature working nowadays? I'm using the version **1.18.22**

I have this code
```
@StandardException
public class MyOwnException extends RuntimeException {

}

public class TestClass{

 public static void main(String ars[]) {
       throw new MyOwnException();
    }
}
```
At `main method ` level I can't use any of those constructors you mentioned we can get by using the annotation 
CC: @ttzn , @rzwitserloot

PR #2718, opened 2021-01-21
Use Enter and MemberEnter to update type mirrors

### LINKED ISSUE #1538: Builder not available to other Annotation Processors
I've been playing around with adding support for Builder in MapStruct and I have noticed that the public static method with no arguments created by Lombok and the Builder itself is not available during the annotation processing phase.

I don't know the code that much. However, I suppose that it needs something similar to what was done for getters and setter in https://github.com/rzwitserloot/lombok/commit/b867f81b8a251a8a32e42f53c2be34d520938bd7.

Would it be possible to have the builder creation method, the Builder and the builder method types to be available to other annotation processors?

Rawi01:
This PR fixes #1538.

As discussed in #1538 the inject methods now use the `javac` enter code to add the new types/methods/fields. As this stuff is a little bit difficult to test I added a hardcoded annotation processor to `Delombok.java#L721` that I used for debugging.

I tested these changes using Java 8 and 14 and also checked that it works in Netbeans.

### DIFF
--- a/src/core/lombok/javac/handlers/HandleConstructor.java
+++ b/src/core/lombok/javac/handlers/HandleConstructor.java
@@ -28,8 +28,6 @@ import static lombok.javac.handlers.JavacHandlerUtil.*;
 import org.mangosdk.spi.ProviderFor;
 
 import com.sun.tools.javac.code.Flags;
-import com.sun.tools.javac.code.Symbol.ClassSymbol;
-import com.sun.tools.javac.code.Type;
 import com.sun.tools.javac.tree.JCTree;
 import com.sun.tools.javac.tree.JCTree.JCAnnotation;
 import com.sun.tools.javac.tree.JCTree.JCBlock;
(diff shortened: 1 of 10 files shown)


dan-lind:
@Rawi01 could you please fix the conflicts?

Rawi01:
@dan-lind I planned to add an additional check to the tests anyway so I did both at the same time. I also uploaded a [pre-build version](https://github.com/rzwitserloot/lombok/files/5909654/lombok.zip) for you to make it easier for you to test if it works as expected together with mapstruct.

dan-lind:
Hi, I tested, but it looks like at least my issue with mapstruct still remains (https://github.com/mapstruct/mapstruct/issues/2348) . 

Basically I just did
`mvn install:install-file -Dfile=lombok.jar -DgroupId=org.projectlombok -DartifactId=lombok -Dversion=snapshot -Dpackaging=jar`

and changed my build.gradle to
```
implementation "org.mapstruct:mapstruct:${mapstructVersion}", "org.projectlombok:lombok:snapshot"
annotationProcessor "org.mapstruct:mapstruct-processor:${mapstructVersion}", "org.projectlombok:lombok:snapshot", "org.projectlombok:lombok-mapstruct-binding:${lombokMapstructBindingVersion}"    
```

Running with
```
springBootVersion = '2.4.2'
mapstructVersion = "1.4.2.Final"
lombokMapstructBindingVersion = "0.2.0"
```

Tried switching the import order as well since I've read that this can have some effect, i.e. importing lombok & bindings before mapstruct, but that didn't change anything.

Anything else you want me to try?

rzwitserloot:
Allright, merged. 1.18.20 (or possibly 1.20, depends on whether meta-annotations make it I think) feels like the right version for something that may have effects we aren't currently foreseeing.

Issue #2714, opened 2021-01-18
[BUG] Consider making 'You aren't using a compiler supported by lombok' diagnostic an error (instead of a warning)


cushon:
**Describe the bug**

When lombok runs in a compiler it doesn't support, it logs the following `WARNING` level diagnostic:

https://github.com/rzwitserloot/lombok/blob/04f17f4dcfbb0e7881d605708467498e3ec56521/src/core/lombok/core/AnnotationProcessor.java#L222

Because the diagnostic is only an error, the compilation can still succeed. In practice, the build will usually failure later when someone tries to reference the methods that lombok would have generated. (https://github.com/bazelbuild/bazel/issues/12837 is an example where the downstream error was seen, but not the warning about the underlying problem.)

I wonder if it that diagnostic should be an error? In cases where the build is going to fail anyways, I think it would be more helpful to fail early with the diagnostic about the unsupported compiler. (Perhaps there are situations where users are relying on this diagnostic being just a warning?)

**To Reproduce**

The repro is just to run lombok on a compiler it doesn't support, e.g. that isn't javac or ecj.

**Version info (please complete the following information):**

* lombok `1.18.16`
* openjdk version `11.0.6+10-LTS`
* annotation processing implementation: https://github.com/google/turbine

PR #2722, opened 2021-01-21
Make diagnostic about using unsupported compilers an error

cushon:
Fixes https://github.com/rzwitserloot/lombok/issues/2714


rzwitserloot:
There's a reason we generate a warning. There are a bunch of environments for help files and other bizarre scenarios where 'emit a warning but otherwise don't prevent this build from succeeding' is the better move.

So, instead I added some code to explicitly detect Turbine and fail with an error in that case. Just pushed, but I haven't tested it though. We'll roll out an edge release with this soon.

PR #2726, opened 2021-01-27
Add end positions to all generated nodes

### LINKED ISSUE #2691: Feature request [internal]: Universally fix up the Javac.storeEnd table
Lombok appears to not always properly add generated nodes to Javac's `storeEnd` table. This doesn't do much (positions of generated nodes are perpetually in a state of limbo, as so many tools make assumptions that do not hold, such as that no 2 exclusive nodes can share a start position, or that nodes have non-zero length), but it should help a little bit sometimes. It is hard to know what the impact is (it'll impact compiler plugins of all stripes, most likely) for example:

* Linters like errorprone (see issue #2605)
* source navigation tools (I can foresee language server-based editor setups such as vi+a java plugin to be affected by this)
* code coverage tools like jacoco
* Editors in general (such as netbeans, see Pull Request #471)
* Other lombok-like APs, though I'm not aware of any
* No doubt I'm forgetting other tools that end up reading that table one way or another

I'm pretty sure it's worth doing this update and then deal with issues as they are raised if doing this causes other tooling to do bizarre things. Most likely it causes other tools to shift from exceptions to undesired reporting on generated code, but we can then start filing requests with those libraries to add facilities for ignoring generated code, and add documentation to the site to explain how to set it up (if we can't convince the project maintainers of things like errorprone to recognize lombok itself and bake into their code that lombok means: generated, so, skip).

Rawi01:
This PR fixes #2691

`setGeneratedBy` and `recursiveSetGeneratedBy` now accept a `JavacNode` instead of `JCTree` and `Context`. Lombok can use this node to find the compilation unit and its end position table to add the generated node there. 

To validate that all nodes have positions I added an annotation processor that scans the ast and performs that check. I also tested this successful in NetBeans and with error-prone.

### DIFF
--- a/src/core/lombok/javac/handlers/HandleBuilder.java
+++ b/src/core/lombok/javac/handlers/HandleBuilder.java
@@ -109,7 +109,6 @@ public class HandleBuilder extends JavacAnnotationHandler<Builder> {
 		boolean isStatic;
 		List<JCTypeParameter> typeParams;
 		List<JCTypeParameter> builderTypeParams;
-		JCTree source;
 		JavacNode sourceNode;
 		java.util.List<BuilderFieldData> builderFields;
 		AccessLevel accessInners, accessOuters;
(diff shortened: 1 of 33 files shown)


rspilker:
Super!

PR #2779, opened 2021-03-19
[fixes #2768] Making the Builder constructor's visibility configurable. 

### LINKED ISSUE #2768: [BUG] Builder Constructors are protected when should be public in 1.18.18
**Describe the bug**
The constructor method to obtain a builder does not have public accessibility even though this is default. It does not respect the accessibility parameter either. The Class.builder() method works as expected.

**To Reproduce**

```
package a;
import lombok.Builder;
@Builder
public class Data {
}

package b;
import a.Data;
public class Usage {
    public void use() {
        Data.builder().build(); // OK
        new Data.DataBuilder().build(); // Doesn't compile
    }
}

```

**Expected behavior**
new Data.DataBuilder() should be public. Explicitly specifying the access as public doesn't have an effect.

 - 1.18.16, 1.18.18
 - Multiple JDKs, command line and IntelliJ

SebestyenBartha:
This makes the Builder constructor's visibility configurable. Also fixes #2768, however, I just stumbled upon that bug while trying to fix our own problem. 
The problem we had in our project is that we need to use Jackson and configure it so that it doesn't override access modifiers (set MapperFeature.CAN_OVERRIDE_ACCESS_MODIFIERS to false) because in our runtime environment (embedded) there is a SecurityManager which doesn't allow that and we have no control over that. In this configuration Jackson cannot work unless the Builder's constructor is public. Making this configurable while keeping the default visibility PACKAGE, existing code is unaffected, but we can add the parameter to our Builder annotations which is more elegant than having to explicitly write the Builder classes.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -41,6 +41,7 @@ Roland Praml <pram@gmx.de>
 Rostislav Krasny <45571812+rosti-il@users.noreply.github.com>
 Samuel Pereira <samuel.p.araujo@gmail.com>
 Sander Koning <askoning@gmail.com>
+Sebestyen Bartha <sebestyen.bartha@gmail.com>
 Szymon Pacanowski <spacanowski@gmail.com>
 Taiki Sugawara <buzz.taiki@gmail.com>
 Takuya Murakami <tmurakam@tmurakam.org>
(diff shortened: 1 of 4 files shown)


rzwitserloot:
The way it works now was intentional. Reasoning stated in, and awaiting feedback at, linked issue #2768 before proceeding on this PR.

If this is discouraging, that sucks :( - good job on the PR 👍 though; proper styling applied, added name to authors file.

PR #2781, opened 2021-03-20
Add GitHub Actions Build+Test

md-5:
Checks that lombok still builds and functions across a wide range of JVMs.
Also fixes a failiure of the Java 12 tests.

### DIFF
--- a/.github/workflows/ant.yml
+++ b/.github/workflows/ant.yml
@@ -1,0 +1,22 @@
+name: Ant Build
+
+on: [push, pull_request]
+
+jobs:
+  build:
+    runs-on: ubuntu-20.04
+
+    strategy:
+      fail-fast: false
+      matrix:
+        java: [11, 12, 13, 14, 15, 16]
+
+    name: Java ${{ matrix.java }}
+
+    steps:
+    - uses: actions/checkout@v2
+    - uses: actions/setup-java@v1
+      with:
+        java-version: ${{ matrix.java }}
+    - run: java -version && ant -version
+    - run: ant clean && ant dist && ant test test.javac6 test.javac8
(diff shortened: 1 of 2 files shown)


Rawi01:
I did something similar [here](https://github.com/Rawi01/lombok/tree/ci) and added a [post in the google group](https://groups.google.com/g/project-lombok/c/vhdEExQnpiw) some time ago but still wait for feedback.

Rawi01:
I rebased my version and slightly modified it to pass again, the main differences is that your version runs `test.javacCurrent`, `test.eclipse-202006`, `test.javac6` and `test.javac8` on JDK 11-16 and my version only runs a single goal . I think thats the better approach because the name of the failed job matches the failing test, WDYT? 
If that would be okay for you I would add my commits to this PR so everything can be merged together.

md-5:
@Rawi01 the javac6 and java8 tests are special in that they use class files from those versions but on the **current VM**, which is why I included them in all jobs to test eg java 15 vm + java 8 class files

Rawi01:
That's right but in a real world scenario no one will compile code like this 😂

Rawi01:
I have also added some integration tests using the provided docker images

rspilker:
Oh, and using all those docker images looks great @Rawi01!

Rawi01:
@rspilker I think it does not block anything unless you enable it (https://docs.github.com/en/github/administering-a-repository/defining-the-mergeability-of-pull-requests). If you want to simply try it out you can also fork the repo as I did.

Rawi01:
@rspilker I just noticed that you added some code quality actions, do you think that we can also manage to merge this/some of my changes/both?

bxb100:
> Hi,
> 
> I'm trying to figure out how this works. Is there a way to test this, without blocking commits or pull requests?

@rspilker 
IMHO, you can fork this repo and checkout this PR, then add `workflow_dispatch`[^1] event trigger to manually run

[^1]: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_dispatch

md-5:
This was added

PR #2783, opened 2021-03-21
Use diamond operator in val post example

Illutax:
This feature exists since java 7 and the extendes support for java 6 ended three years ago.

### DIFF
--- a/website/usageExamples/valExample_post.jpage
+++ b/website/usageExamples/valExample_post.jpage
@@ -4,17 +4,17 @@ import java.util.Map;
 
 public class ValExample {
 	public String example() {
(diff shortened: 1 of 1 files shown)


rzwitserloot:
... and `var` exists since java 10. Lombok still supports java6, because supporting a thing is better than not supporting a thing, and so far it hasn't been all that much effort to keep java6 in the fold.

The examples also serve the function of literally showing what lombok is exactly doing under the hood, though this is a little bit of an art more than a science (because we don't show the various mostly irrelevant annotations we generate to ensure linter tools do the right thing). Still, lombok does generate the stuff-in-the-diamond, even on java7+, and in this case I feel it is better to show that in the example.

PR #2793, opened 2021-03-25
Add licenses of embedded dependencies

Rawi01:
@mirabilos [noticed](https://groups.google.com/g/project-lombok/c/z2OAsdkX1VQ) that there are no license information of embedded dependencies. I used the format that is also used by [gradle](https://github.com/gradle/gradle/blob/master/LICENSE) so I think this should be fine.


mirabilos (approved):
Looks good to me.

Issue #1297, opened 2017-02-13
@ToString formatting 'language'.


rzwitserloot:
This is the umbrella issue for any requests to improve or modify the way `@ToString` prints in some fashion.

The general notion would be to have a regex like pattern language to specify exactly what you want. Some examples, given:

    package foo;
    public class Test {
        String name;
        int age;
        int[] values;

        public String getSummary() { return "hello"; }")
    }
   1:  new Test("Joe", 50, new int[] {5, 12, 14})
   2: new Test(null, 50, new int[0]);
   3: new Test(null, 50, null);

    @ToString("<class>{<#name>: <#getSummary()>}")
    1: Test{Joe: hello}
    2,3: Test{null: hello}

    @ToString("<fqn>{<#values: [a, b]>: <fieldNames: n1=v1 n2=v2>}")
    1: foo.Test{[5, 12, 14]: name=Joe age=50}
    2: foo.Test{[]: name=Joe age=50}
    3: foo.Test{null: name=Joe age=50}

etcetera.

Concerns:

* Print classname, fully qualified classname, or nothing.
* Print all fields, vs all fields not explicitly called out
* Print the result of a no-args method.
* Specify how to print listy things (prefix/suffix/infix, and/or how to deal with an empty one)
* Specify alternate content in case of `null`.
* Print content indented, including properly indenting the content if it, itself, has newlines in it, to allow structured printing.

This issue supercedes #876 #1216 #1276 #918 

Not sure if we actually want this; there's a lot to cover, toString() is a debugging aid (if it's something else, you're doing it wrong), and it feels like reinventing an entire programming language when you can just.. program your toString() method in java by.. just typing java code in the editor you're already in.

askoning:
This issue also supercedes #256 #529 #1211

gmathias:
I don't think hideNulls (#256 #529 #1211) should be covered (only) by pattern. The pattern is for formatting while hideNulls is for filtering content. Also it would be over-complicated to write a pattern just to say "hide nulls". Please keep the boolean flag separated from the pattern.

jefferyyuan:
a separate boolean flag for hideNulls is very useful.
- This is the most wanted feature about ToString annotation
- It's much easier to use for client
- It's much easier to implement

wheredevel:
Is there a way of weaving methods in Lombok? (_Traits...?_) (_Mixins...?_)
Why do I ask under this issue:
Because I have a method called toJson().
[...]

PR #2807, opened 2021-04-05
add @ToString.Format(foramtter = <? extends ToStringFormatter>.class)

RezaGholamzad:
Add **@ToString.Format** to Configure the behavior of how this member is rendered in the toString method.
You should implement the **ToStringFormatter** interface then override the format method.
**This solution is compile time and does not depend on runtime, because we give the qualified name of the class to the formatter.**
for example : 
```
public class PasswordFormatter implements ToStringFormatter {
    @Override
    public <T> String format(T t) {
        return "formatted field : " + t;
    }
}
```
```
@Data
public class User {
    @ToString.Format(formatter = PasswordFormatter.class) // give the formatter the qualified name of the class.
    public long password;
}
```
thanks.

### DIFF
--- a/src/core/lombok/ToString.java
+++ b/src/core/lombok/ToString.java
@@ -90,7 +90,7 @@ public @interface ToString {
 	
 	/**
 	 * Configure the behavior of how this member is rendered in the
-	 * {@code toString}; you should implement the {@code ToStringFormatter};
+	 * {@code toString}; You should implement the {@code ToStringFormatter};
 	 * interface then override the format method.
 	 */
 	@Target(ElementType.FIELD)
--- a/src/core/lombok/ToString.java
+++ b/src/core/lombok/ToString.java
@@ -89,6 +89,17 @@ public @interface ToString {
 	boolean onlyExplicitlyIncluded() default false;
 	
 	/**
+	 * Configure the behavior of how this member is rendered in the
+	 * {@code toString}; you should implement the {@code ToStringFormatter};
+	 * interface then override the format method.
+	 */
+	@Target(ElementType.FIELD)
+	@Retention(RetentionPolicy.SOURCE)
+	public @interface Format {
+		Class<? extends ToStringFormatter> formatter();
+	}
+	
+	/**
 	 * If present, do not include this field in the generated {@code toString}.
 	 */
 	@Target(ElementType.FIELD)
--- a/src/core/lombok/ToStringFormatter.java
+++ b/src/core/lombok/ToStringFormatter.java
@@ -1,0 +1,5 @@
+package lombok;
+
+public interface ToStringFormatter {
+	<T> String format(T field);
+}
--- a/src/core/lombok/eclipse/handlers/EclipseHandlerUtil.java
+++ b/src/core/lombok/eclipse/handlers/EclipseHandlerUtil.java
@@ -1545,6 +1545,42 @@ public class EclipseHandlerUtil {
 		return call;
 	}
 	
+	/**
+	 * Creates an expression that calls a method on an external object whose
+	 * argument reads the field(Will either be {@code this.field} or
+	 * {@code this.getField()} depending on whether or not there's a getter).
+	 */
+	static Expression createFieldAccessor(EclipseNode field, FieldAccess fieldAccess, ASTNode source, String className, char[] methodName) {
+		
+		int pS = source == null ? 0 : source.sourceStart, pE = source == null ? 0 : source.sourceEnd;
+		long p = (long) pS << 32 | pE;
+		
+		className = className.substring(0, className.length() - 6);
+		String[] classNames = className.split("\\.");
+		long[] poss = {p};
+		char[][] tokens = new char[classNames.length][];
+		for (int i = 0; i < classNames.length; i++) {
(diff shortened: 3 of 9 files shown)


janrieke:
Admitted that I only took a brief look at your solutions (mainly the test cases), I think there is a severe problem with your approach: Lombok is a compile-time-only dependency, but your solution requires it to be on the (runtime) classiest. That's an immediate showstopper.

There is a reason why this hasn't been implemented, yet: Noone has yet come up with a sufficiently good proposal, not to mention implementing it.

janrieke:
The umbrella issue for discussions on this topic is #1297.

RezaGholamzad:
> Admitted that I only took a brief look at your solution (mainly the test cases), I think there is a severe problem with your approach: Lombok is a compile-time-only dependency, but your solution requires it to be on the (runtime) classpath. That's an immediate showstopper.
> 
> There is a reason why this hasn't been implemented yet: Noone has yet come up with a sufficiently good proposal, not to mention implementing it.
> 
> Disclaimer: I'm not a Lombok team member, just a contributer.

Give the formatter the qualified name of the class.

janrieke:
But don't you require the annotation parameter `formatter()` to be of type `ToStringFormatter`? And looking at the 
_after_ test files, the `implements ToStringFormatter` is still present after Lombok has finished. That means it remains in the class file after compilation, i.e. a runtime dependency.

RezaGholamzad:
> But don't you require the annotation parameter `formatter()` to be of type `ToStringFormatter`? And looking at the
> _after_ test files, the `implements ToStringFormatter` is still present after Lombok has finished. That means it remains in the class file after compilation, i.e. a runtime dependency.

The code is generated at compile time and placed inside the toString method at compile time.
Execution of this code has an effect on the owner's code, which surely the owner of the code knows what he is doing.
Therefore, the code is the responsibility of the code owner and has no effect on Lombok code at all, so there is no need to worry.
Lombok only created the code at compile time that the owner of the code wanted.

Rawi01:
Yes, the code inside the `toString` method does not require lombok at runtime but the actual formatter does as already pointed out by @janrieke.

RezaGholamzad:
> Yes, the code inside the `toString` method does not require lombok at runtime but the actual formatter does as already pointed out by @janrieke.

Yes, there will be a dependency between the two classes, but this is a choice we give the user.
The user knows the advantages and disadvantages of using this feature, so the user decides whether to use it or not depending on the conditions of his project.

janrieke:
> The user knows the advantages and disadvantages of using this feature, so the user decides whether to use it or not depending on the conditions of his project.

I really do not understand what you are proposing here.
It's probably not this: "If you want this feature, modify `pom.xml` or `build.gradle` to include `lombok.jar` in your runtime." Because that's just too weird. Not to mention that Lombok was never designed to be a runtime dependency, which would probably introduce a bunch of other problems.

So what's the solution? A separate `lombok-runtime.jar`? Let lombok copy the `ToStringFormatter` interface into the annotated class?

RezaGholamzad:
> > The user knows the advantages and disadvantages of using this feature, so the user decides whether to use it or not depending on the conditions of his project.
> 
> I really do not understand what you are proposing here.
> It's probably not this: "If you want this feature, modify `pom.xml` or `build.gradle` to include `lombok.jar` in your runtime." Because that's just too weird. Not to mention that Lombok was never designed to be a runtime dependency, so putting `lombok.jar` to runtime would probably introduce a bunch of other problems.
> 
> So what's the solution? A separate `lombok-runtime.jar`? Let Lombok copy the `ToStringFormatter` interface into the annotated class?

Yes, thank you very much for your explanation.
The two ways you mentioned can be options, although they do not seem to be acceptable.
In order to look for a better way, we need to know that the team members want to upgrade the Java version, for example Java 8, so that we can use the Lambda expressions feature.
Another way is to use one of the interfaces inside the Java library that is similar to the ToStringFormatter interface.
These are just ideas that may be the prelude to a better idea.
thanks.

rzwitserloot:
This will have to be shot down, I'm afraid.

* There is a very long history on the desire to make the to string code more flexible, but all proposals fall way short of what people imagine it could do. This, too, falls short. The discussion linked earlier explains in more detail.
* What this is particularly good at, is already available! You don't need this change.

Here is how you would program your use-case, which has worked for a few years' worth of lombok versions:

```
@Data
public class User {
    public long password;

    @ToString.Include(name = "password") String renderPassword() {
        return "formatted field: " + password;
}
```

This will include the result of invoking `renderPassword` in your toString, using the name `password`, which then hides the password field.

rzwitserloot:
NB: As a side note, I'm having a hard time imagining any code base that results in having objects that contain passwords in String form no less as fields, which then get themselves logged to boot. That just sounds like a bad code base and I don't like it if lombok is enabling bad code. However, programming is complex and I won't go so far as to say that _all_ possible cases of 'whoops, the default toString would result in passwords showing up in logfiles' are bad. Just.. I bet most of them are, and the fix is to e.g. not have that field at all, to clean up your JPA states, to stop treating objects representing requests-in-flight as casual things that you can just toss about with abandon, etc.

Issue #1921, opened 2018-10-24
Proposal: @XXXArgsConstructor - handle automatically OR have an attribute `throws` for specifying Exception classes to declare as thrown by the generated constructor.


wheredevel:
```
class A {
  A() throws Exception {}
}

//Vanilla
class B {
  B() throws Exception {
    super(); // called implicitly, and it throws..! 
  }

}

//Lombok
@NoArgsConstructor(throws={Exception.class})
class B {
}
```

It might be handled automatically WITHOUT specifying such an attribute, as it's obvious: if B would not declare throws - it wouldn't compile as Vanilla case shows.  

Currently, if you put ```@NoArgsConstructor``` on B in it would give an error: "Unhandled exception of type Exception" - and you go and boiler-plate all the siblings.

rzwitserloot:
Well, 'throws' is a keyword so it can't be (throws=...).

Sooo.. hurls=? pitches, lobs, tosses, propels, chucks?

Maybe throwables=? or exceptions=?

I don't like any of those.

If no good ideas forthcoming by 2018-12-01, close issue.

wheredevel:
Take `throwing={Exception.class}`

rspilker:
I know I marked this issue as `accepted`, but now I've changed my mind. 

1. Currently, all constructors we generate cannot pass parameters to a super constructor. So only the no-args super constructor will ever be invoked.

2. The most important reason to add `throws` clauses would be because the super constructor would throw a checked exception.

3. We think it is very unlikely that a no-args super constructor would throw a checked exception

4. We never generate checked exceptions in a generated constructor

So the feature does not seem to carry it's weight. Too much maintenance costs for too little benefits.

rspilker:
If we ever get resolution to work reliably, we expect to propagate constructor parameters and `throws` clauses automatically.

PR #2834, opened 2021-05-06
Add XXXArgsConstructor exceptions thrown declaration (#1921)

dembanakh:
Implements #1921:
Added throwing property to XXXArgsConstructor, to specify the exceptions to declare on the constructor.
For now, one must manually specify the list of exceptions. As far as I understand, inferring the exceptions from the base class constructor may be problematic due to the resolving happening after lombok. But I haven't looked into this inferring part yet.


rspilker:
Thanks @dembanakh for spending the time and effort to create this pull request.

I have to apologize for not rejecting the feature request sooner. As you might have seen in #1921 we think that the added maintenance costs are no match for the expected benefits.

We do invite you to keep helping us with developing Lombok.

Before creating a pull request, please check with @rzwitserloot or me to see if it still makes sense.

PR #2840, opened 2021-05-12
Also copy vartype.type in TreeMirrorMaker

### LINKED ISSUE #2420: [BUG] NPE when using two "val"s together with anonymous class instantiation in nested scope
**Describe the bug**
When in the same nested scope there are two `val`s followed by the instantiation of an anonymous class, compiling with javac, a NullPointerException is raised:

```
Exception while resolving: NODE LOCAL (class com.sun.tools.javac.tree.JCTree$JCVariableDecl) @val()
final ___Lombok_VAL_Attrib__ k = 2(RegularFileObject[C:\Dev\lombok-crash-test-case\src\TestCase.java])
java.lang.NullPointerException
        at com.sun.tools.javac.comp.Check$Validator.visitTypeIdent(Check.java:1340)
        at com.sun.tools.javac.tree.JCTree$JCPrimitiveTypeTree.accept(JCTree.java:2084)
        at com.sun.tools.javac.comp.Check$Validator.validateTree(Check.java:1359)
        ....
```

**To Reproduce**
The following Java source can reproduce the problem when compiled with `javac -cp lombok.jar TestCase.java`:
```
import lombok.val;

class TestCase {
	public void func() {
		for (int i = 0; i < 1; ++i) {
			val j = 1;
			val k = 2;

			new TestCase() { };
		}
	}
}
```

**Expected behavior**
It should compile.

 - Lombok version: behavior observed both with 1.6.22 and 1.18.12
 - Platform: javac 1.8.0_192

**Additional context**
No problem at all when compiling in Eclipse.

Compilation succeeds even with javac, if any of the following happens:
- one of the `val`s is removed (j or k, it doesn't matter)
- the curly braces are removed after `new TestCase()` thus instantiating a named class instead of an anonymous class
- the for cycle is removed thus removing the nested scope and having the two `val`s and the anonymous class instantiation in the root function scope

Rawi01:
This PR fixes #2420

The `TreeMirrorMaker` has some special code to copy symbols and types for variables in some cases. It seems like that it is also required to copy the `vartype` field if we copy the other stuff. It also works if the whole symbol/type copy block gets removed but that most likely reintroduce the original issue even if I was unable to reproduce it.

**Edit**
Added a visitor to also copy reference types

### DIFF
--- a/src/utils/lombok/javac/TreeMirrorMaker.java
+++ b/src/utils/lombok/javac/TreeMirrorMaker.java
@@ -32,8 +32,10 @@ import lombok.javac.JavacTreeMaker.TypeTag;
 import com.sun.source.tree.LabeledStatementTree;
 import com.sun.source.tree.VariableTree;
 import com.sun.tools.javac.tree.JCTree;
+import com.sun.tools.javac.tree.JCTree.JCFieldAccess;
 import com.sun.tools.javac.tree.JCTree.JCVariableDecl;
 import com.sun.tools.javac.tree.TreeCopier;
+import com.sun.tools.javac.tree.TreeScanner;
 import com.sun.tools.javac.util.Context;
 import com.sun.tools.javac.util.List;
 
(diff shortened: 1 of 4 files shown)


Rawi01:
Doesn't work in all cases...

PR #2859, opened 2021-05-31
Add Locked annotations

### LINKED ISSUE #610: Synchronized with Reentrantlock
_Migrated from Google Code (issue 575)_

PimvanderLoos:
This PR closes #610 

This an implementation of the feature specification as defined in the associated issue.

Due to their similarities, this code was mostly adapted from `@Synchronized`.

I purposefully deviated from the specification a little regarding the field type of the generated lock objects. The specification states that the type of the generated field for `@Locked` should be `j.u.c.l.Lock` and `j.u.c.l.ReadWriteLock` for the Read/Write variants. However, I used `j.u.c.l.ReentrantLock` and `j.u.c.l.ReentrantReadWriteLock` respectively.
My defense for this decision is that the [Lock](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/locks/Lock.html) and even more so the [ReadWriteLock](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/locks/ReadWriteLock.html) interfaces are quite meager compared to the [ReentrantLock](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/locks/ReentrantLock.html) and the [ReentrantReadWriteLock](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/locks/ReentrantReadWriteLock.html). Because the specification states that the generated names should not start with a `$` to indicate that users can interact with the lock objects (as there wouldn't be much of a point in `@locked` otherwise), I thought that using the type that actually offers a bunch of useful methods would be nicer than requiring the user to cast to it every time they want to actually interact with it. However, if you disagree with this, it's easy enough to store them as the interfaces as originally intended instead.

Also, I haven't given this annotation a tagline for the website yet.


PimvanderLoos:
@rzwitserloot Could you review this, please?

rzwitserloot:
Looks good. Some nits:

* ~~The javadoc of `@Locked` refers to `/features/Locked` - shouldn't that be `/features/experimental/Locked`?`~~ - you caught this one in the last commit already.
* The `handleFlagUsage` call is missing the call specifically to handle flagging any experimental usage. See e.g. HandleFieldNameConstants; instead of `handleFlagUsage(annotationNode, ConfigurationKeys.LOCKED_FLAG_USAGE, annotationname);` replace it with `handleExperimentalFlagUsage`, which will also take care of flagging if the 'warn/err on experimental lombok stuff' setting is enabled.
* In `HandleLockedUtil`: a `boolean` field is wrapped into a 1-len `boolean[]`, which ordinarily is done as a stand-in for a mutable primitive. It looks like you did this solely to allow the eclipse-world's `preHandle` to change it; in the javac-world side it looks like it could have just been a `boolean` method. This is fine, but for code readability purposes, I'd prefer it if this was an `AtomicBoolean` instead of a `boolean[]`. I think it yells 'I am a mutable single value!' much better, it has better toString etc behaviours, and the locking overhead is infinitesemal when not needed on modern JVMs (and, of course, quite desired if needed).
* `src/core/lombok/eclipse/handlers/HandleLockedUtil` has a `typeReferenceNameEquals` method. It feels like this should either [A] be using an already available method in `src/core/lombok/eclipse/handlers/EclipseHandlerUtil`, [B] This method should be added to EclipseHandlerUtil, not to HandleLockedUtil, or most likely [C] just use `Arrays.deepEquals`, which can deal with comparing 2 `char[][]` objects as a one-liner.

Bang up job, We're (@rspilker is looking over my shoulder) impressed and feeling a little guilty for not getting to it earlier.

PimvanderLoos:
Hey, thanks for the feedback!
I've applied it to the PR.

I noticed it still says `@Locked was introduced as an experimental feature in lombok v1.18.21.` in `website/templates/features/experimental/Locked.html `, which was somewhat overly optimistic in hindsight. I didn't update this, but just thought I should mention that that should probably be updated when/if this gets merged (in addition to the tagline).

PimvanderLoos:
@rzwitserloot @rspilker Could you take another look at the updated code?

SimSonic:
@rzwitserloot @rspilker hi!

I'm here just to mension that with the release of Loom's virtual threads this functionality could be more important. Loom currently has bad support for syncronization on objects but fully supports java.util.concurrent package.

rzwitserloot:
Hey Pim - fantastic work. I only changed fairly superficial things and some style nitpicking, nothing to bother reviewing. The things that have some impact:

* I'll discuss it with @rspilker but with the more pressing needs of JDK20's virtual threads, I've shifted it straight to main package (skip `experimental`).
* You always generated the fields if they weren't there. I've reverted that behaviour back to how `@Synchronized` works: the fields are auto-genned if they don't exist _only_ if you use the parameterless version. If you explicitly pick a name, you have to explicitly write the field.
* The names are, like `@Synchronized`, dollarized if lombok generates them. (`$lock` / `$LOCK`).

Issue #2871, opened 2021-06-08
[BUG] FileOutputStream resource is not closed correctly


vn17:
**Describe the bug**
The resource would not be closed in case `out.write()` threw an Exception on [this line](https://github.com/projectlombok/lombok/blob/5120abe4741c78d19d7e65404f407cfe57074a47/src/core/lombok/core/debug/AssertionLogger.java#L73).

**Expected behavior**
The resource is closed even if an Exception is thrown in the try block

**Version info (please complete the following information):**
 - Lombok version - All
 - Platform (javac or eclipse, and if so, what is the output of `javac -version` / the version listed in the _about..._ dialog of eclipse. - All

**Additional context**
Change implementation to try-with-resources. 
Reference: https://www.oracle.com/technical-resources/articles/java/trywithresources.html

rspilker:
Since we are still compatible with java6, we cannot use try with resources.

I think some static code analysis detected this. Is that correct, or did you actually encounter a problem.

At the time of writing, we thought that closing the resource in a finally block would not be worth the code. We just opened an output stream and only wrote a few bytes. And only if we are debugging. In practice, this is never going to occur, and never in production.

PR #2872, opened 2021-06-08
Resource not closed corrrectly

vn17:
The resource would not be closed in case `out.write()` threw an Exception. Changing implementation to try-with-resources.
Issue: https://github.com/projectlombok/lombok/issues/2871
Reference: https://www.oracle.com/technical-resources/articles/java/trywithresources.html


rspilker:
Rejected, see #2871 for explanation.

PR #2878, opened 2021-06-16
Make @Builder work with both Default and throwing constructor

emlun:
Currently, if the constructor throws a checked exception then `@Builder` must be on the constructor in order for the checked exception to propagate to the throws clause of the `builder()` method. However, `@Builder.Default` requires that the `@Builder` annotation be set on the class. It is possible to annotate both the class and the constructor, but then the constructor one gets skipped. This means it is currently impossible to use `@Builder.Default` simultaneously with a throwing constructor (except perhaps by manually implementing the `builder()` method), as demonstrated by the tests added in commit d905be5.

This attempts to fix that. If you set `@Builder` on both the class and the constructor, the constructor's checked exceptions will now be added to the `builder()` method initially generated by the class annotation. I have also verified that this works in the real-world context where I encountered the issue.

### DIFF
--- a/src/core/lombok/javac/handlers/HandleBuilder.java
+++ b/src/core/lombok/javac/handlers/HandleBuilder.java
@@ -489,7 +489,12 @@ public class HandleBuilder extends JavacAnnotationHandler<Builder> {
 		
 		{
 			MemberExistsResult methodExists = methodExists(job.buildMethodName, job.builderType, -1);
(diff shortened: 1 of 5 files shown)


emlun:
Oh, and I'm basing the branch on `v1.18.20` because I couldn't get the tests to pass on any later commits. I hope that doesn't cause any trouble, I'm happy to fix the merge conflicts once that becomes relevant.

Issue #2778, opened 2021-03-19
[BUG] Override @JsonIgnore in the subclass not work properly as before.


ijustyce:
**Describe the bug**
Before spring boot v2.1.18.RELEASE I can override @JsonIgnore in the subclass. But from spring boot v2.1.18.RELEASE I can't. And I find spring boot v2.1.18.RELEASE update lombok to 1.18.16.

I have three class and named UserVO, InsertUserVO and UpdateUserVO I want to ignore the ID when create a user and don't ignore while update. Just like the below.

UserVO.java
```java
@Getter
@Setter
public class UserVO {

    private Long id;

    private String name;

    private Integer age;
}
```  

InsertUserVO.java
```java
@Getter
@Setter
public class InsertUserVO extends UserVO {

    @JsonIgnore
    private Long id;
}
```  

UpdateUserVO.java
```java
@Getter
@Setter
public class UpdateUserVO extends InsertUserVO {

    private Long id;
}
```

Before spring boot v2.1.18.RELEASE lombok will generate code like below

![right](https://user-images.githubusercontent.com/6041166/111744459-3dffab80-88c6-11eb-8703-1f70183419b3.png)

But when update to spring boot v2.1.18.RELEASE or latest lombok will generate code like below.

![error](https://user-images.githubusercontent.com/6041166/111744583-69829600-88c6-11eb-8ca0-2a3912c75f5f.png)

I hope to generate code like before. What should I do?

Rawi01:
The only difference between these two classes is the additional annotation on the setter and I think that Jackson ignores the field in both cases. Can you please explain why the annotation is a problem?

ijustyce:
I can't explain why, but Jackson not ignore the field if code generate as before. I have tested it.

arbing:
We've got the same problem

janrieke:
You aren't bound to the Lombok version that spring uses, as Lombok is not a runtime dependency.
Could you switch to the previous Lombok version to verify that this is indeed a Lombok issue?

janrieke:
However, I very much doubt that copying those Jackson annotations automatically is necessary at all. Jackson considers the setter, getter, and field as a single virtual property, and any annotation on any of those should work for all. See https://github.com/FasterXML/jackson-databind for an example.

I can only see two scenarios where copying helps:
1. Builders
2. when getter/setter name generation in Lombok does not match with Jackson (crappy BeanSpec).

ijustyce:
[...]

Issue #2769, opened 2021-03-12
[BUG] Field-based Jackson mixins break when using @Setter


mjustin:
## Describe the bug
Jackson allows creating [mixins](https://github.com/FasterXML/jackson-docs/wiki/JacksonMixInAnnotations) which allow defining Jackson annotation-based configuration on a class mirroring the class to be (de)serialized.  If the annotations are on the fields but not the getters/setters, then the "mixin" class must likewise be annotated on the fields.

For instance, take the following configuration:

```java
public class Example {
    @JsonProperty(access = JsonProperty.Access.READ_ONLY)
    private String value;

    private String other;
}

public class ExampleMixin {
    @JsonProperty(access = JsonProperty.Access.READ_WRITE)
    private String value;
}
```

In this case, deserializing the JSON value `{"value": "Some Value", "other": "Other Value"}` normally will produce an `Example` object with `value` set to null and `other` set to "Other Value", since the `@JsonProperty` annotation tells Jackson to skip the field during deserialization.  Deserializing it with the mixin added, on the other hand, will produce an `Example` object with `value` set to "Some Value" and `other` set to "Other Value", as the mixin tells Jackson to deserialize it.

Annotating `value` in class `Example` with `@Setter` will cause the field-based mixin to effectively be ignored, wheras manually adding a setter will not.

```java
// Deserializing the example JSON with ExampleMixin will result in a null value field.
public class Example {
    @Setter
    @JsonProperty(access = JsonProperty.Access.READ_ONLY)
    private String value;
}
```

```java
// Deserializing the example JSON with ExampleMixin will result in a value field set to "Some Value".
public class Example {
    @JsonProperty(access = JsonProperty.Access.READ_ONLY)
    private String value;
    
    public void setValue(String value) {
        this.value = value;
    }
}
```

The reason for this discrepancy is that Lombok automatically copies a number of Jackson annotations from fields to their generated setters per `@Setter`:

https://github.com/rzwitserloot/lombok/blob/2e5ea518cc4041f2ef79d5d9cc0f970d805b6e64/src/core/lombok/core/handlers/HandlerUtil.java#L316-L329

With `@Setter`, the generated setter has the `@JsonProperty(access = JsonProperty.Access.READ_ONLY)`.  Since Jackson getter/setter annotations have priority over field annotations, the generated setter annotation wins out over both the annotated field on the class and on the mixin.  Therefore, the field is not deserialized.

[...]

PR #2915, opened 2021-08-02
Copy JsonProperty annotation to getter

XenoAmess:
I strongly suggest copy JsonProperty annotation onto getter functions.
tests applied.

### DIFF
--- a/src/core/lombok/core/handlers/HandlerUtil.java
+++ b/src/core/lombok/core/handlers/HandlerUtil.java
@@ -312,6 +312,7 @@ public class HandlerUtil {
 			"org.springframework.lang.Nullable",
 			"org.netbeans.api.annotations.common.NonNull",
 			"org.netbeans.api.annotations.common.NullAllowed",
+			"com.fasterxml.jackson.annotation.JsonProperty"
 		}));
 		COPY_TO_SETTER_ANNOTATIONS = Collections.unmodifiableList(Arrays.asList(new String[] {
 			"com.fasterxml.jackson.annotation.JacksonInject",
--- a/test/jackson/CopyJsonPropertyAnnotationToGetterTest.java
+++ b/test/jackson/CopyJsonPropertyAnnotationToGetterTest.java
@@ -1,0 +1,58 @@
+package jackson;
+
+import java.util.Arrays;
+import java.util.List;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import org.junit.Test;
+
+import static org.junit.Assert.assertEquals;
+
+@lombok.Data
+@lombok.AllArgsConstructor
+class TestDto1 {
+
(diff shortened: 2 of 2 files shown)


janrieke:
I don't understand why copying is necessary at all. IIRC Jackson considers field, getter, and setter as a unit; an annotation on one of them holds for all.

XenoAmess:
> I don't understand why copying is necessary at all. IIRC Jackson considers field, getter, and setter as a unit; an annotation on one of them holds for all.
> Shouldn't Lombok simply stop copying those annotations at all (except for builders)?

I provided the test classes. You can just run it to see.

XenoAmess:
Hi. is there any news?

janrieke:
Jackson's behavior appears strange to me. Why should _property annotation merging_ not occur when there is no setter?
I did not find a hint in the documentation that explains this. Thus, I consider this a Jackson bug. Could you file an issue at [Jackson](https://github.com/FasterXML/jackson-databind)?

XenoAmess:
> Jackson's behavior appears strange to me. Why should _property annotation merging_ not occur when there is no setter? I did not find a hint in the documentation that explains this. Thus, I consider this a Jackson bug. Could you file an issue at [Jackson](https://github.com/FasterXML/jackson-databind)?

issue created https://github.com/FasterXML/jackson-core/issues/735

janrieke:
So my conclusion is that this is (again) a result of the crappy bean-spec. Question is: Should lombok once again try fixing this by copying the annotation?

When lombok introduced/extended the copy-to-setter mechanism, that already had unintended side-effects: #2778, #2769
Right now, once lombok copies an annotation, there is no way for users to stop lombok from doing so. So users would have to remove the lombok annotation and implement the generated methods by themselves. That's a workaround I personally do not like.
In contrast, the workaround for this case is quite easy: `@Getter(onMethod_=@JsonProperty("xYz"))`

However, you may argue that if lombok copies an annotation to a setter, then it should also be put on a getter. I agree with that. Right now, I don't see a reason why this would cause further issues with Jackson, but I did not investigate further. However, I still vote for lombok to stop copying completely (except for builders). But that's a decision for the maintainers.

If the maintainers decide that lombok should also copy Jackson annotations to the getter, this PR is most likely not sufficient. There are several other Jackson annotations that we have to look at, and we need the make sure that this really does not break any more use-cases.

XenoAmess:
> However, you may argue that if lombok copies an annotation to a setter, then it should also be put on a getter. I agree with that. 

Yes, I really do.

I really hate inconsistency like this... 

I can't give very much detail as that is somehow company secret, but it has really already bring me some trouble.

rzwitserloot:
If @janrieke is not willing to call that the right move is unambiguously to copy it to the getter, well, my (and @rspilker's) thorough lack of experience with jackson means we can't call it either. No go.

PR #2920, opened 2021-08-06
Use Java 7+ syntax to make example more concise

mwisnicki:
It's almost as good as `@Cleanup` :)

### DIFF
--- a/website/usageExamples/CleanupExample_post.jpage
+++ b/website/usageExamples/CleanupExample_post.jpage
@@ -2,24 +2,15 @@ import java.io.*;
 
 public class CleanupExample {
 	public static void main(String[] args) throws IOException {
-		InputStream in = new FileInputStream(args[0]);
-		try {
+		try (
+			InputStream in = new FileInputStream(args[0]);
 			OutputStream out = new FileOutputStream(args[1]);
-			try {
-				byte[] b = new byte[10000];
-				while (true) {
-					int r = in.read(b);
-					if (r == -1) break;
-					out.write(b, 0, r);
-				}
-			} finally {
-				if (out != null) {
-					out.close();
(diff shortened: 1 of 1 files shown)


rzwitserloot:
I get the problem but it's really difficult. Specifically:

* The pre- and post- snippets are dancing to two tunes: They are _both_ attempting to highlight what lombok _actually does_, as well as attempting to explain what one would write if one wasn't using lombok.
* Ordinarily these 2 tunes are entirely aligned and it's no big deal. However, here, they are not: Lombok generates explicit try/finally code, because why not - if you're on java6, the try-with-resources doesn't work, or if you have a `close()` method but aren't extending `@AutoClosable`, or you're using `@Cleanup("release")` to specify a different name, then it __does not work__ with the ARM (try-with-resources) construct, but `@Cleanup` works fine. Thus, I see no particular reason to attempt to make lombok's handling of `@Cleanup` attempt to use try-with-resources: It's backwards incompatible and would reduce lombok's ability to be compatible with admittedly long-EOLled versions of java. Thus, the _actually represent what lombok does_ tune is obliterated by this PR.
* Nevertheless, the 'this is what lombok replaces' part is much better represented by your PR.

I came down on it thusly: Whilst `@Cleanup` has benefits even over try-with-resources, it's a tossup, and from the view of 'when in rome, be like the romans' (i.e: Write idiomatic java code unless you have excellent reasons not to do so), `@Cleanup` should probably be considered an obsolete lombok feature. As such I think we should lean towards catering to the crowd that cares particularly about backwards compatibility. Meaning: We should NOT 'update' to try-with-resources in how it is implemented, and should be clear that we aren't using that in what your `@Cleanup` code gets desugared to.

Issue #2652, opened 2020-11-20
[BUG] slow compilation by using val in a local block for the same variable name


mschaaf:
**Describe the bug**
We have a gradle, spring, java build and use lombok in our test and main code.
We run into a compilation performance issue with using `val`.
We introduced the following test method and the compilation takes now 10 seconds longer
Slow to compile:
```java
    @SneakyThrows
    @Test
    void testImport() {
        {
            val user = ValueStore.user();
            user.setEmail("");
            user.setActive(true);
            userRepository.save(user);
        }

        try (val pkg = OPCPackage.open(TestClass.class.getResourceAsStream("/file.xlsx"))) {
            try (val wb = new XSSFWorkbook(pkg)) {
                val sheet = wb.getSheetAt(1);

                val connection = DriverManager.getConnection(url, username, password);
                val jdbcConnection = new JdbcConnection(connection);
                val mysqlDatabase = new MySQLDatabase();
                mysqlDatabase.setConnection(jdbcConnection);

                val consumer = new Consumer(mysqlDatabase);
                consumer.accept(sheet.getRow(0));
                consumer.accept(sheet.getRow(1));
                consumer.accept(sheet.getRow(2));

                mysqlDatabase.commit();
                mysqlDatabase.close();

                Assertions.assertThat(consumer.getUsersCreated().intValue()).isEqualTo(1);
                Assertions.assertThat(consumer.getUsersUpdated().intValue()).isEqualTo(1);
            }
        }

        {
            val user = userRepository.findByEmail("a");
            Assertions.assertThat(user.getFirstname()).isEqualTo("A");
            Assertions.assertThat(user.getLastname()).isEqualTo("C");
            Assertions.assertThat(user.getEmail()).isEqualTo("a");
            Assertions.assertThat(user.getEmail2()).isNull();
            Assertions.assertThat(user.getEmail3()).isNull();
            Assertions.assertThat(user.getActive()).isFalse();
            Assertions.assertThat(userRepository.getEmployeeId("a")).isEqualTo(0);
        }

        {
            val user = userRepository.findByEmail("b");
            Assertions.assertThat(user.getFirstname()).isNotEqualTo("A");
            Assertions.assertThat(user.getLastname()).isNotEqualTo("G");
            Assertions.assertThat(user.getEmail()).isEqualTo("b");
            Assertions.assertThat(user.getEmail2()).isNull();
            Assertions.assertThat(user.getEmail3()).isNull();
[...]

PR #2928, opened 2021-08-13
Replace val with native final var in Java >= 10

Rawi01:
This PR fixes the performance problems descried in #2652. It replaces `val` with `final var` if the build source version supports it.


NicklasWallgren:
Any estimate to when this will be released? Thanks.

rspilker:
Super, this was already high on our list.

NicklasWallgren:
@rspilker 
When can we expect a new release, or edge build?
Thanks

The latest edge build is version: 1.18.21 (2021-07-23 04:27:04 UTC)

PR #2961, opened 2021-09-15
Do not escape @ character on Linux Eclipse installer javaagent path. Fixes #2960

### LINKED ISSUE #2960: [BUG] lombok Eclipse Installer with @ char in Path on Linux
**Describe the bug**
The lombok Eclipse installer escapes @ characters in the javaagent path to lombok.jar in eclipse.ini, then Eclipse fails to start with an alert window containing "JVM terminated. Exit code=1" followed by the content of eclipse.ini.

**To Reproduce**

- Make a new Eclipse installation with a path containing @ character
- Make a lombok installation on it.
- Start Eclipse

**Expected behavior**
Eclipse starts.

**Version info:**
 - Lombok version: 1.18.20
 - Eclipse 2021-03 (4.19)
 - Ubuntu 20.04
 - OpenJDK 16 as Eclipse runtime (installed through Ubuntu apt repositories)

Stratehm:
Fixes #2960

### DIFF
--- a/src/installer/lombok/installer/IdeLocation.java
+++ b/src/installer/lombok/installer/IdeLocation.java
@@ -65,7 +65,7 @@ public abstract class IdeLocation {
 		}
 	}
 	
-	private static final String LEGAL_PATH_CHARS = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789.-_/";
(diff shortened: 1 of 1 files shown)


rzwitserloot:
Good idea, but we need to go one step further, and never attempt escaping at any point given that evidently escaping is simply not an option here. See 0d7260d. One worry I now have is that spaces in windows doesn't work.

I have confirmed that having a space in the path on macs now 'works' (in the sense that not escaping that space in the eclipse.ini `-javaagent:` line is the correct solution). Care to test on linux? Thanks for the idea and PR, @Stratehm !

Issue #2973, opened 2021-09-25
[BUG] Strict Encapsulation - Clashing Packages / Package org.objectweb.asm.signature in both module lombok and module org.objectweb.asm


GedMarc:
**Describe the bug**
The shade of ASM does not get placed in a lombok specified package. This causes test cases in JDK's with module-info's to fail due to duplicate package names in multiple modules

**To Reproduce**
Include ASM 9+ as a dependency in a module application

```
                 <dependency>
                        <groupId>org.ow2.asm</groupId>
                        <artifactId>asm</artifactId>
			<version>9.1</version>
                    </dependency>
```
module-info
```
requires org.objectweb.asm;
requires static lombok;
```
VM Boot Error
```
Error occurred during initialization of boot layer
java.lang.LayerInstantiationException: Package org.objectweb.asm.signature in both module lombok and module org.objectweb.asm

Process finished with exit code 1
```

Execute a test case

**Expected behavior**
VM boots

**Version info (please complete the following information):**
 - Lombok version 1.18.20
 - ```mvn test```

**Additional context**
All shades should be placed under the owned package name, and shouldn't reuse packages owned by other libraries

rzwitserloot:
<sidetrack rant>
That's a common shading strategy and yet also seems befuddlingly stupid. asm is asm is asm, why in the blazes would you want the exact same class file to exist in 85 differently named but otherwise identical flavours? Shading tools should mangle the names by including versions or hashes, not by project name. If a combination of build tools and JVM can't handle this, I think you should report bugs with them, then.
</sidetrack rant>

Lombok's shader works differently from both what common shaders do (the dumb thing; stripe in the project's name), and what they ought to do (the version). Instead, lombok modifies _the extension_, because lombok, intentionally, wants to not be there. At any time and in any phase of the process. Lombok's deps are for lombok only. Lombok is running as agent or AP and in that sense is 'outside the universe', or should be. That's why we've set it up that way.

The immediate problem, then, cannot possibly be explained by how the files are stored in the jar. They are stored in a way that could not possibly cause this problem.

However, we do _load_ them. This presumably is the source of the issue. However, the solution to isolation at a class loader level is customized loaders. Not name mangling.

@GedMarc's suggestion that lombok joins the lemmings off the cliff is denied. We shall not join that lunacy. It's not just for elegance reasons: That kind of renaming of things causes slowdowns in the build and is a big step.

[...]

PR #2977, opened 2021-09-28
Replicate JDK 17 Package Offense by Lombok, VM Kill - #2973

GedMarc:
Illegal packaging - using packages that are owned by other modules.

VM Boot Kill


GedMarc:
https://github.com/projectlombok/lombok/issues/2973

rzwitserloot:
Cannot reproduce; this is 'testing' for a non-existing problem. See issue.

Issue #2509, opened 2020-07-03
[BUG] Syntax highlighting for lombok annotations is broken


humphreygao:
**Describe the bug**
In Eclipse IDE, syntax coloring for some lombok annotations(such as `@Data`、`@Slf4j`、`@ToString`、`@EqualsAndHashCode`) are not correct.

**To Reproduce**
1) Install Eclipse lombok plug-in.
2) Via `Window > Preferences > Java > Editor > Syntax Coloring`, choose the `Java > Annotations` from preference page and change the color of annotations(for example, orange).

![lombok annotation color](https://user-images.githubusercontent.com/52394598/86490240-498f5e80-bd99-11ea-8642-c4e37e59fc5f.png)

**Version info (please complete the following information):**
 - Lombok version: 1.18.12
 - Platform: Eclipse IDE 2020-06

Rawi01:
I can not reproduce this issue. 
![img](https://user-images.githubusercontent.com/5850477/86509552-2096ba00-bde9-11ea-8d8c-1bad4637fec9.png)
Can you check if there is an error message? (Window > Show View > Error Log)

janrieke:
I have seen similar things, and I think it dates back to at least 1.18.2. However, I was unable to identify a common pattern or a way to reproduce.

rspilker:
Probably some of the positions of the generated AST nodes are off. We've also seen similar problems with `@Cleanup`, but have not yet put time in it to find the wrong positions yet

humphreygao:
@Rawi01 There is no error.

My computer OS is Windows 10. The syntax coloring is correct before the plugin was installed. You can try annotations one by one to see which annotation's color is wrong. When there are several annotations, some annotation's color may be okay(in the picture, `@Data` and `@ToString` is okay but not okay when they are alone).

Sometimes move the annotations up and down(using `Alt + Up/Down` shortcut) may affect the annotation color.

Rawi01:
Interesting, I know that this can happen if there are errors.

I just tried it using a new eclipse version + lombok 1.18.12 and it works as expected.

PR #2984, opened 2021-10-06
Fix syntax highlighting for EqualsAndHashCode

Rawi01:
This might be related to #2509

**Without the fix:** 
`@EqualsAndHashCode` is blue 
img1
After adding a new line half of the class looses its highlighting
img2

**After the fix:**
img3


rspilker:
Good find

PR #2993, opened 2021-10-09
To map from map

lioutall:
convert obj to map
convert map to obj

image

### DIFF
--- a/src/core/lombok/FromMap.java
+++ b/src/core/lombok/FromMap.java
@@ -1,0 +1,13 @@
+package lombok;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Target(ElementType.TYPE)
+@Retention(RetentionPolicy.SOURCE)
+public @interface FromMap {
+
+}
+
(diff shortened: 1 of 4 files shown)


janrieke:
_Disclaimer_: I'm not a lombok team member.

I'm sorry if i'm dampening your enthusiasm, but what exactly is the use-case for this feature?
To me, it appears to be somehow related to serialization. But why should Lombok deal with that, when there are lots of established frameworks around that are better suited for the job?

Or, to be more concrete, why don't you take Jackson and do this?

```
@Builder
@Value
public static class DTO {
	private int i;
	private String s;
}

ObjectMapper mapper = new ObjectMapper();
DTO dto = DTO.builder().i(1).s("s").build();
Map<String, Object> map = mapper.convertValue(dto, new TypeReference<Map<String, Object>>() {});
DTO dto2 = mapper.convertValue(map, DTO.class);
assertEquals(dto, dto2);
```

Jackson will do much more sophisticated stuff than lombok will ever do, e.g. type checks, type conversions, recursive (de)serialization, ...

rspilker:
I agree with @janrieke, this is the domain of serialization frameworks. There are too many open concerns, like loops,

lioutall:
I use it in ORM. Something like create instance with DB result (eg: Quarkus, Vert'x)
I accept this unmerged commit.

Issue #2693, opened 2021-01-02
Change lombok's property capitalization code. `Field xName` -> `getxName()`.


rzwitserloot:
This is a catch-all issue that tries to capture the relevant information from a ton of separate issue requests (to those reviewing bug reports, please mark as duplicate of just this issue).

The problem:

Lombok's name conversion tool, used by `@Getter` and `@Setter`, as well as `@With`, and therefore also `@Data` and `@Value`, turns a field that starts with a single lowercase letter followed by a capital into a capital followed by a capital: `@Getter private String uName;` generated `public String getUName() {return this.uName; }`.

However, various tools out there expect `getuName` instead, in this unique case where the second letter of the field/property name is a capital where the first is lowercase.

Unfortunately, there is no convention we can follow, because there are also tools out there that expected `getUName`. No matter what lombok does, some tooling / expectations will be left out in the cold. More generally then, because of these conflicts, __names like uName are problematic in any case, and you should try never to use such names.__

There is nothing lombok can do to solve all problems with single-lower-case-letter property names: If you decide to use fields named like this, sooner or later you're going to run into trouble. Quoting user [@Maaartinus](https://github.com/rzwitserloot/lombok/issues/1454#issuecomment-322196958):

> Javabeans is fubared. Some long time ago I was unhappy about Lombok mistreating my "uShaped" property and wrote manual etters as a workaround. IIRC it didn't work either, because some other tool didn't understand it in the opposite direction (setuShaped -> uShaped), so I ended using "shapedAsU" instead. 🙄

However, the majority of the community has standardized on `xName` -> `getxName()`, and thus we are now willing to accept a PR, even though this will cause no small amount of pain, as it'll be backwards incompatible, and we reserve the right to wimp out and delay integrating the Pull Request until a major update with more breaking changes.

Tools and specs that prefer `xName` -> `getxName`:
* [JavaBeans Specification (last updated: 1997!), §8.8](https://download.oracle.com/otndocs/jcp/7224-javabeans-1.01-fr-spec-oth-JSpec/). Quoting from this PDF, as it is often slow or impossible to download:

[...]

PR #2996, opened 2021-10-18
[Fix for #2693] Adding a new accessors flag - javaBeansSpecCapitalization

YonatanSherwin:
Adding a new accessors flag, which will allow to follow Java Bean Spec capitalization rules (e.g `aField` -> `getaField()`)

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -49,6 +49,7 @@ Takuya Murakami <tmurakam@tmurakam.org>
 Thomas Darimont <thomas.darimont@gmail.com>
 Till Brychcy <till.brychcy@mercateo.com>
 Victor Williams Stafusa da Silva <victorwssilva@gmail.com>
+Yonatan Sherwin <yonatansherwin@gmail.com>
 Yun Zhi Lin <yun@yunspace.com>
 
 By adding your name to this list, you grant full and irrevocable copyright and patent indemnity to Project Lombok and all use of Project Lombok in relation to all commits you add to Project Lombok, and you certify that you have the right to do so.


rzwitserloot:
Gracias, @YonatanSherwin I'm looking it over now. Can you add one more commit to this where you add your name to the `AUTHORS` file? Please note that this act (as explained at the bottom of that file) means you consent that your work will fall under lombok's copyright rules.

A few updates I'm currently making:

* The config key is becoming an enum, with valid values: "basic" and "beanspec". I'm not foreseeing more than 2, but attempting to attach 'true' and 'false' to 2 different strategies is a  bit unsightly, I think.
* Your code didn't handle builders, which also generates method names (such as `clearAddresses`) that need to adhere to the chosen configuration option.
* The option to set this option directly as param on the `@Accessors` annotation is going to go away. Our general rule is: If we can't imagine serious scenarios where one would wish to have different applications of a setting within a single package, we don't allow it to be an on-annotation setting, because we acknowledge that we have limited room for such settings (we've messed up somewhere if a lombok annotation has a menu of options as long as my leg, especially if most of them seem like a mistake if you set them on the annotation instead of as a project-wide config). If one day we realize that it _is_ common enough to offer it, we can always add it; removing it is much harder as that breaks backwards compatibility.

Said differently, if it really feels like the annotation parameter should get the following line in its documentation: "... but you probably dont want to use this and instead set a config key for the entire project", we don't add it :)

rzwitserloot:
I've finished the above updates. Next step is adding some docs, updating changelog, integrate your `AUTHORS` update, and push an edge release.

YonatanSherwin:
Thanks @rzwitserloot !
BTW i've created a new PR for fixing two tests failures:
https://github.com/projectlombok/lombok/pull/2998

PR #3080, opened 2022-01-04
don't generate checkerframework's @NotCalledMethods

### LINKED ISSUE #2908: [BUG] Bad Checker Framework annotation NotCalledMethods
When setting `checkerframework = 4.0`, lombok generates `org.checkerframework.checker.calledmethods.qual.NotCalledMethods` on the setter methods of `@Builder`/`@SuperBuilder`.
However, this annotation does not exist. It's still located in the old package: `org.checkerframework.checker.builder.qual.NotCalledMethods`

This results in a compile error.

**To Reproduce**

Compile 

    @lombok.Builder
    public class BuilderTest {
        String name;
    }

with lombok.config:

    lombok.addNullAnnotations = checkerframework
    checkerframework = 4.0

**Expected behavior**
Generate the correct annotation.

A better solutions would be (from my point of view) generating no annotation at all. If the Checker Framework will support this annotation eventually, lombok could start generating it again (with the help of the version check). However, I doubt the usefulness of such an annotation: Why should it be a misuse of the builder pattern to call a builder setter method twice? For instance, you could fill in some default values into a new builder and pass it to other methods to possibly overwrite those default. And when using `toBuilder()`, you will have no choice but calling a setter twice, because all methods must be marked as "called" on the return type of `toBuilder()` to avoid Checker Framework complaining about uncalled methods.

Furthermore, the minimum version for `@CalledMethods` (and `@This`) should be reduced, as it is already supported by Checker Framework 3.16.0 (maybe earlier).

 - Lombok version: 1.18.20
 - Checker Framework version: 3.16.0
 - Platform: javac 11.0.12

janrieke:
This PR removes generation of CheckerFramework's `@NotCalledMethods`. It was previously generated on builder setter methods to ensure those are not called twice. However, as described in #2908, CF currently does not support this annotation. Even if it would be, it is also semantically problematic, because you couldn't call setters methods on a builder created by `toBuilder`.

This fixes #2908 (PR for the bad `@This` will follow).


janrieke:
A PR for the bad `@This` is either based on this PR or would have conflicts if this PR is accepted. So I'll delay that until you merge (or reject) this PR, so I can resolve potential conflicts right away.

rspilker:
Thank

PR #3113, opened 2022-02-17
Fix enum logging

### LINKED ISSUE #2990: [BUG] @Log fails on nested enums
**Describe the bug**
Compilation fails when the annotation @Log is set on a nested enum.

**To Reproduce**
Here is an example:
```java
class Example {

  @Log
  enum Status  { OK, KO }
}
```
The following error appears during compilation:  
`@Log is not supported on non-static nested classes.`

**Expected behavior**
Should compile without error.
Was working fine in 1.18.20

 - Lombok 1.18.22
 - all platforms

**Additional context**
Addind the keyword `static` to the enum seems to workaround the problem.

varkart:
Fixes #2990

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -19,6 +19,7 @@ Jacob Middag <jacob@gaddim.nl>
 Jan Matèrne <jhm@apache.org>
 Jan Rieke <it@janrieke.de>
 Jappe van der Hel <jappe.vanderhel@gmail.com>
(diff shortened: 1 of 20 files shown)


rspilker:
Isn't the problem that `isStaticAllowed()` returns the wrong value for enums?

rzwitserloot:
Had to revert your changes and apply the principle to our `isStaticAllowed` method.

PR #3136, opened 2022-03-13
Search patched method also by parameters

### LINKED ISSUE #3134: [BUG] Eclipse 4.23 refactor/rename from within Java Source Editor throws java.lang.VerifyError: Bad type on operand stack
Given a fresh _Eclipse_ install from upcoming 4.23.0 Release (Candidate Build: 4.23RC2a), it appears that when Lombok is installed as java agent in eclipse.ini
```
...
-vmargs
-javaagent:xxx\eclipse\lombok-1.18.22.jar
...
```
the _Eclipse_ refactor feature no longer works from within the _Source Editor_ window.

**To Reproduce**
Given plain Java (no Lombok) ...
```
   var message = "Hallo World";
   System.out.println(message);
}
```
when trying to refactor/rename the `message` variable (e.g. select the text and use the context menu), _Eclipse_ throws a `java.lang.VerifyError.`

**Version info:**
 - Lombok version 1.18.22, but also a [latest build](https://github.com/projectlombok/lombok/commit/8c3ff7e9b08c4bf9c2b23cffb6042c31258e3ebb) from the 'master' branch has same issue
 - Platform Windows with Java 17.0.2 as shipped with the _Eclipse_ Release Candidate (Candidate Build: 4.23RC2a)

**Additional context**
Eclipse also has a _Code Outline_ view, from where one can initiate a refactor/rename e.g. on class fields. It appears that with that there are no issues.

Rawi01:
This PR fixes #3134

Eclipse added a new `findByNode` method to `LinkedNodeFinder`. Because both use the same name we need to include the parameters in the search to find the correct one. I also added a patch for the new method.

### DIFF
--- a/src/eclipseAgent/lombok/eclipse/agent/EclipsePatcher.java
+++ b/src/eclipseAgent/lombok/eclipse/agent/EclipsePatcher.java
@@ -332,10 +332,16 @@ public class EclipsePatcher implements AgentLauncher.AgentLaunchable {
 	
 	private static void patchHideGeneratedNodes(ScriptManager sm) {
 		sm.addScriptIfWitness(OSGI_TYPES, ScriptBuilder.wrapReturnValue()
-				.target(new MethodTarget("org.eclipse.jdt.internal.corext.dom.LinkedNodeFinder", "findByNode"))
+				.target(new MethodTarget("org.eclipse.jdt.internal.corext.dom.LinkedNodeFinder", "findByNode", "org.eclipse.jdt.core.dom.SimpleName[]", "org.eclipse.jdt.core.dom.ASTNode", "org.eclipse.jdt.core.dom.SimpleName"))
 				.target(new MethodTarget("org.eclipse.jdt.internal.corext.dom.LinkedNodeFinder", "findByBinding"))
 				.wrapMethod(new Hook("lombok.launch.PatchFixesHider$PatchFixes", "removeGeneratedSimpleNames", "org.eclipse.jdt.core.dom.SimpleName[]",
 						"org.eclipse.jdt.core.dom.SimpleName[]"))
+				.request(StackRequest.RETURN_VALUE).build());
+		
+		sm.addScriptIfWitness(OSGI_TYPES, ScriptBuilder.wrapReturnValue()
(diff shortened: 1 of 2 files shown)


Rawi01:
Eclipse commit: https://github.com/eclipse-mirrors/org.eclipse.jdt.ui/commit/64a3b5b9cfb663e786ba9a9ad3f8deddc3c85ea2

rzwitserloot:
The one question that comes up: That patch we do there - to remove generated simple names, should we also apply it to the overloaded `findByNode(ASTNode, Name)`? I don't know which eclipse code calls it, but presumably doing the same filtering there wouldn't hurt?

dstango:
Do you think it might be advisable to expand other MethodTarget-definitions with the parameter specifications in EclipsePatcher to prevent something similar happening with a future version of Eclipse? I guess adding overloaded methods might occur again and the error seems obscure enough that people don't know lombok needs an update, but go to the Eclipse site.

Or maybe any other ideas how to best deal with probable version incompatibilites? Would it be e.g. advisable for Eclipse to have some explicitly defined extension point lombok could hook into, instead of having to rely on reflection?

And another question: does the provided fix still work with older versions of Eclipse? As there are now hooks for two variants of findByNode -- what happens in an older Eclipse version? Is this silently ignored or will it cause refactoring to break in older versions?

Rawi01:
Yes, I think that expanding all definitions should prevent issues like this in the future. It will break some lombok related stuff but at least people will figure out that it is a lombok problem. I was also working on automated tests for the refactoring patches, this should also help to find problems like this easier.

Hooks would be nice but adding them is a bunch of work and it will only work in new eclipse/ecj versions.

The fix should also work in older versions, if it can not find a method to patch it simply does nothing.

dstango:
Okay, so Eclipse-hooks could only be a long term solution. I'm not deep enough in Eclipse internals yet, but if that changes some day I'm willing to contribute, as lombok is one of my favorite extensions :-).

Looking at the original stack trace in https://github.com/projectlombok/lombok/issues/3134 I guess the reason for the exception was that MethodTarget was specifying a method in an ambiguous way, and lombok simply used probably the first found of them. Probably it would be good to check for such situations explicitly, and reject or ignore such ambiguous MethodTargets. Thus it would (at least partially) break proper functioning of the lombok part, but not break the general Eclipse functioning per se.

Rawi01:
In this case the patcher patched all methods with the name and ignores return type and arguments. This actually is quite handy in cases where you do not care about it or if the patched code is stable. Adding return and argument types for all `MethodTarget`s should help to prevent a problem like this in the future. I started to work on this.

PR #3161, opened 2022-04-01
Fix documentation @NonNull

pehecoro:
NonNull vanilla code is using lombok. I believe the ideal would be to show just how it is without lombok.

### DIFF
--- a/website/usageExamples/NonNullExample_post.jpage
+++ b/website/usageExamples/NonNullExample_post.jpage
@@ -1,9 +1,7 @@
-import lombok.NonNull;
-
 public class NonNullExample extends Something {
 	private String name;
 	
-	public NonNullExample(@NonNull Person person) {
+	public NonNullExample(Person person) {
 		super("Hello");
 		if (person == null) {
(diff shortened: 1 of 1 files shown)


rzwitserloot:
Lombok removes all its annotations on a delombok run. _Except_ `@NonNull` which it leaves be.

That's important background information, given: we intentionally didn't remove it. For two (related) reasons:

* the examples have dual purpose: Partly to show what one would write when lombok isn't available, and partly to show what lombok actually does / what delombok will do. These 2 ought to be identical; in rare case where they aren't, we'll take it on a case by case basis, but generally lean towards 'it shows what lombok actually does, i.e. what delombok would generate, assuming you configure its format options to be as pretty as possible, e.g. don't generate `@Generated` annotations and the like. Because delombok doesn't remove this annotation, we don't remove it in the post snippet either.
* The reason we don't remove it is also relevant to 'how would you do this without lombok': The annotation isn't _just_ to cause lombok to generate code. It is also, on its own, documentary. That's also why the retention of that annotation is higher than all the other lombok annotations.

Issue #1841, opened 2018-08-28
Overloading a Builder's method for mass-adding to a Singular field removed the default convenience methods


CarlSalaets:
For a `@Singular List<E> elements` the `@Builder` generates 3 convenience methods:
- `element(E element)`
- `elements(Collection<? extends E> elements)`
- `clearElements()`

but if I try to overload the collection setter it no longer generates these 3 methods.
```java
@Builder
public class BuilderExample {
    @Singular
    public List<String> elements

    public static class BuilderExampleBuilder {

        /*
         * This overloaded method prevents the generation of the default methods.
         */
        public BuilderExampleBuilder elements(Stream<String> elementStream){
            return elements(elementStream.collect(toList()));
        }

        /*
         * This overloaded method is fine; it does not prevent the generation of the default methods.
         */
        public BuilderExampleBuilder element(Integer elementOfAnotherType){
            return element(String.valueOf(elementOfAnotherType));
        }

    }
}

Note: I'm not overwriting the generated methods, I'm overloading them to accept different types of arguments.
```

rspilker:
I would have expected and error saying `Manually adding a method that @Singular @Builder would generate is not supported. If you want to manually manage the builder aspect for this field/parameter, don't use @Singular.`

As far as I can see from the code, it only looks at the method names. One of the reasons is that it is very hard for us to see if you've created an overload or a replacement method. The latter will not really work since we have an (on purpose) undocumented internal representation that is optimized for further processing.

Maaartinus:
@rspilker I see the error you mentioned. However, overloads should be supported with `@Tolerate`. This is a typical use case for it, but it seems to be ignored, so I'd suggest to reopen.

CarlSalaets:
[@Tolerate](https://projectlombok.org/api/index.html?lombok/experimental/Tolerate.html) is still experimental for now, But that would be a good way to mitigate this issue.

jmoraleda:
I have encountered this exact issue. Currently I work around it but using a different name but it would provide a cleaner API if could overload the same name. Explicitly annotating the method [@Tolerate](https://projectlombok.org/api/index.html?lombok/experimental/Tolerate.html) would be very reasonable.

jmoraleda:
I remark that this issue is a duplicate of the older (albeit less well documented and commented on) https://github.com/projectlombok/lombok/issues/1498

PR #3193, opened 2022-05-12
[fixes #1498] Respect @Tolerate when using @Singular

### LINKED ISSUE #1498: @Tolerate should work with @Singular
`@Tolerate` is a nice feature to create overloaded methods alongside methods that will be genrated by lombok.

So let's say you want to provide an overload for a method created by the `@Singular` annotation in combination with the `@Builder` annotation, then it would be very convinient if I could just add the `@Tolerate` annotation to my overload.

erikhofer:
This fixes #1498 and provides a solution to #1841.

```java
import java.util.List;

import lombok.Builder;
import lombok.Singular;
import lombok.experimental.Tolerate;

@Builder
public class Foo {
	
	public static class FooBuilder {
		
		@Tolerate
		public FooBuilder bar(Integer i) {
			bar(i.toString());
			return this;
		}
	}
	
	@Singular
	private final List<String> bars;
}

```

This code currently produces the following error: `Manually adding a method that @Singular @Builder would generate is not supported. If you want to manually manage the builder aspect for this field/parameter, don't use @Singular.`

With this fix, it is possible to manually suppress this error with `@Tolerate`. It enables adding convenience methods with the same name as the generated methods.

### DIFF
--- a/src/core/lombok/eclipse/handlers/EclipseSingularsRecipes.java
+++ b/src/core/lombok/eclipse/handlers/EclipseSingularsRecipes.java
@@ -226,6 +226,7 @@ public class EclipseSingularsRecipes {
 				}
 				case METHOD: {
 					AbstractMethodDeclaration method = (AbstractMethodDeclaration) child.get();
+					if (isTolerate(child, method)) continue;
 					char[] name = method.selector;
 					if (name == null) continue;
 					if (getGeneratedBy(method) != null) continue;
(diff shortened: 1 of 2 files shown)


Rawi01:
`@Tolerate` documentation:
> Put on any method or constructor to make lombok pretend it doesn't exist, i.e., to generate a method which would otherwise be skipped due to possible conflicts.

I think it is reasonable to assume that someone tries to add this annotation and I also don't know a reason why it should not work. If we add this I would propose to also change the error message e.g. `or add the experimental @Tolerate annotation to your method`. If we don't allow this we should update the `@Tolerate` documentation

blacelle:
Could this be re-considered for merging?

PR #3199, opened 2022-05-19
Eclipse integration tests

Rawi01:
This PR adds support for eclipse integration tests.

The test runner starts the OSGI bundles and the dependencies, after that it is possible to use a lot of the eclipse features for test. All classes share the same classloader, thats why I added an additional patch to detect the bundle. There might be an easier solution for that but I couldn't find one.

I also added some `ant` targets to run the tests and download all the required dependencies. The dependency downloader also generates a eclipse user library that can be imported in eclipse.

It would be great if someone can try if all this stuff works.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,4 @@
 /website/lombokSupporters
 /pom.xml
 /jvm.locations
+/testenv
--- a/buildScripts/ivy.xml
+++ b/buildScripts/ivy.xml
@@ -43,7 +43,7 @@
 		<dependency org="org.apache.ant" name="ant-junit" rev="1.10.5" conf="build->default" />
 		
 		<!-- test deps -->
-		<dependency org="junit" name="junit" rev="4.8.2" conf="test->default; sources" />
+		<dependency org="junit" name="junit" rev="4.13.2" conf="test->default; sources" />
 		<dependency org="log4j" name="log4j" rev="1.2.17" conf="test->default; sources" />
 		<dependency org="org.apache.logging.log4j" name="log4j-api" rev="2.17.1" conf="test->default; sources" />
 		<dependency org="commons-logging" name="commons-logging" rev="1.2" conf="test->default; sources" />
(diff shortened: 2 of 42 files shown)


rzwitserloot:
This is awesome. The one problem is, this commit has 'hardcoded' linux.x64-86. This is one of those times where my main dev machine is a problem (as it's apple aarch64), so, the eclipse-oxygen-full test is never gonna work. But, I managed to get 2022-03 to work, but I had to make 2 modifications:

* The looong list of deps downloaded straight from `download.eclipse.org` needs to include 	`<arg value="org.eclipse.swt.cocoa.macosx.aarch64" />`.
* Because of dumb reasons, you need to add `<jvmarg value="-XstartOnFirstThread" />` to the `<junit>` block in `tests.ant.xml` - but that will cause the JVM to fail to start due to 'I do not know what that option means' on all OSes except MacOS.

I think we can update the ant scripts to automate all this. For my box:

* `${os.name}` resolves to `Mac OS X`.
* `${os.arch}` resolves to `aarch64`.
* `<condition name="foo" value="bar"><os name="Mac OS X" arch="aarch64" /></condition>` will set property `foo` to value `bar`. On any other machine it would leave `foo` unset.

I'm going to integrate this patch into master as it doesn't affect anything unless you explicitly want to (with `ant test.eclipse-oxygen-full` for example). I need somebody to figure out what os.name  / os.arch print on e.g. x86-64 linux, and on windows (at least 64-bit architectures), then I think I can make that ant script do the right thing on at least win64, linux64, mac-aarch64, and I guess mac-x86-64 is worth it still, too. I think the major source of x32 arch is raspberrys but surely nobody is going to hack on lombok on such a device, so that should cover most of it.

rzwitserloot:
I also did my best on making this OS/arch independent. I confirmed that gravitar systems on AWS (AMD-based linux) would work (they report architecture as `aarch64` and os name as `Linux`), I'm fairly confident this should work on the combinatorial explosion of mac/windows/linux vs x86-64/aarch64.

PR #3209, opened 2022-06-10
Generate @return and @param JavaDoc tags for @Getter and @Setter annotations

lars-sh:
I suggest to generate the JavaDoc tags `@return` and `@param` for the lombok annotations `@Getter` and `@Setter` unless they are specified by the developer manually.

Missing JavaDoc tags cause warnings in javac of JDK11 and later. That's why I came across this idea, which literally avoids developers to copy over JavaDoc code of really simple getters and setters for the `@return` or `@param` JavaDoc tags.

I hope this makes sense. Feel free to give me feedback to the general idea, the implementation, code style and so on.

This Pull Request contains multiple test cases for both lombok annotations inside `src/test/resources/transform/...`, while here comes one example to show the use of this request:

*Lombok code:*
```Java
/**
 * First name of the employee
 */
String firstName;
```

*Delomboked code:*
```Java
/**
 * First name of the employee
 */
@Getter
String firstName;

/**
 * First name of the employee
 * @return First name of the employee
 */
public String getFirstName() {
    return this.firstName;
}
```


lars-sh:
@rspilker, I really enjoy work.
Is there something I can do to support this Pull Request?

dstango:
This seems to be a somewhat different concept than the one already present, which allows to specify @param and @return on the field and then copies/moves it around: https://projectlombok.org/features/GetterSetter

As far as I've witnessed the development of lombok so far, I guess a new concept like this would first need some discussion and consideration before such a PR might be accepted.

I guess @rzwitserloot or @rspilker should say how to best proceed.

PR #3262, opened 2022-09-20
fixes #3259: allow @ExtensionMethod on interfaces

### LINKED ISSUE #3259: [BUG] Interfaces cannot be annotated with @ExtensionMethod
**Describe the bug**
Since Java 8 (2014) interfaces have `default` methods, and it makes perfect sense to use extension methods there.

```java
    // an interface - this will fail
    @ExtensionMethod(ProblemWithInterfaces.Extensions.class)
    public static interface GreeterAsInterface {
        // some code which would like to use extension methods
        ...
    }
```
https://github.com/marc-guenther/lombok-problems/blob/df013050404c29c0ce3b6b133e3182dc32643133/prob/lems/ProblemWithInterfaces.java#L17-L19

Unfortunately, in the Lombok sources, it is explicitly checked that `@ExtensionMethod` annotation is only allowed on `class` and `enum`.

**To Reproduce**
See here for an example project:
https://github.com/marc-guenther/lombok-problems

**Expected behavior**
The annotation should be allowed on `interface` as well. There is no technical reason why it shouldn't. In fact, I checked that it does in fact work, and use it like that for quite a while now without any problems (PR coming...)

 - Lombok version: 1.18.24
 - Platform: javac 1.8.0_282, Eclipse 2019-03 (4.11.0)

marc-guenther:
Simple change to allow `@ExtensionMethod` on interfaces, done for javac and Eclipse handlers. Using this without any problems for almost two months now. (fixes #3259)

### DIFF
--- a/src/core/lombok/eclipse/handlers/HandleExtensionMethod.java
+++ b/src/core/lombok/eclipse/handlers/HandleExtensionMethod.java
@@ -50,10 +50,10 @@ public class HandleExtensionMethod extends EclipseAnnotationHandler<ExtensionMet
 		int modifiers = typeDecl == null ? 0 : typeDecl.modifiers;
 		
 		boolean notAClass = (modifiers &
-				(ClassFileConstants.AccInterface | ClassFileConstants.AccAnnotation)) != 0;
+				(ClassFileConstants.AccAnnotation)) != 0;
 		
 		if (typeDecl == null || notAClass) {
-			annotationNode.addError("@ExtensionMethod is legal only on classes and enums.");
+			annotationNode.addError("@ExtensionMethod is legal only on classes and enums and interfaces.");
 			return;
 		}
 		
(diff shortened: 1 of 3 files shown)


marc-guenther:
Is there anything in particular which is blocking this PR?

It is a small, quite obvious two-line change, to allow @ExtensionMethod annotation on interfaces. Which is needed if you want to use extension methods in the `default` methods of interfaces (a feature which is around since 2014).

It doesn't break any tests, and I use it for months now, without any problems. 

Can I do anything to get this merged? Should I add a test case for it?

rzwitserloot:
We can't really accept any PRs unless you add your name to the AUTHORS file. However, because this commit essentially just removes stuff I think legally we're allright in this case. For next time :)

marc-guenther:
Cool, thanks.

About the AUTHORS file, I didn't find any info on that. If that is a requirement, it should be added to the Contributing page?

Issue #3267, opened 2022-09-26
[BUG] Broken link to feature page in Javadocs for @Log (and friends)


mjustin:
**Describe the bug**
The various log-related annotations have a broken link to the features page in their class-level Javadocs.  They link to https://projectlombok.org/features/Log, but that link does not work.  The working link appears to be https://projectlombok.org/features/log (the only difference being the case of the word "log" in the URL path).  The current link is for a 404 page containing the text "This is not the page you're looking for.".

Affected classes:

* [lombok.CustomLog](https://projectlombok.org/api/lombok/CustomLog)
* [lombok.extern.apachecommons.CommonsLog](https://projectlombok.org/api/lombok/extern/apachecommons/CommonsLog)
* [lombok.extern.flogger.Flogger](https://projectlombok.org/api/lombok/extern/flogger/Flogger)
* [lombok.extern.java.Log](https://projectlombok.org/api/lombok/extern/java/Log)
* [lombok.extern.jbosslog.JBossLog](https://projectlombok.org/api/lombok/extern/jbosslog/JBossLog)
* [lombok.extern.log4j.Log4j](https://projectlombok.org/api/lombok/extern/log4j/Log4j)
* [lombok.extern.log4j.Log4j2](https://projectlombok.org/api/lombok/extern/log4j/Log4j2)
* [lombok.extern.slf4j.Slf4j](https://projectlombok.org/api/lombok/extern/slf4j/Slf4j)
* [lombok.extern.slf4j.XSlf4j](https://projectlombok.org/api/lombok/extern/slf4j/XSlf4j)

**To Reproduce**
View the Javadoc file or HTML page of any of the affected classes.  In a web browser, open the [the project lombok features page for lombok log annotations](https://projectlombok.org/features/Log) link at the top of the documentation.

**Expected behavior**
The link should go to the working features page for "@Log (and friends)"

**Version info (please complete the following information):**
 - Lombok version: 1.18.24
 - Platform: N/A

PR #3275, opened 2022-10-06
Rename log to Log to comply with link in java docs

mucharafal:
Currently, links in Java docs have the path `feature/Log`, which such path doesn't exist on the website (there is `feature/log` instead)


edysli:
Fixes #3267

rzwitserloot:
Actually, it's the javadoc that is wrong. The reason we call the log feature page 'log' and not 'Log' is because 'Log' suggests that it is the docs for a hypothetical `@Log`. It's not; it's the docs for all the various log-related lombok annos.

Thanks for noticing and pointing the way to how to fix it :). I've updated the javadoc in just-pushed commit 731bb185077918af8bc1e6a9e6bb538b2d3fbbd8.

To cover for existing (wrong) URLs out in the wild, we're going to add a redirect from `/features/Log` to `/features/log`.

mucharafal:
Thanks for the feedback and for fixing the issue!

juriad:
aaa

PR #3299, opened 2022-11-17
The URL to constructor has been corrected : replaced "Constructor" with "constructor"

YassinHajaj:
Hi,

The existing links are dead.
image

Replaced them with the new links.
image


rspilker:
This change would fix future links, for now I've added a redirect from `Constructor` to `constructor`

PR #3335, opened 2023-01-21
[Eclipse 4.27] Lombok ignores lombok.config

### LINKED ISSUE #3332: [Eclipse 4.27] Lombok ignores lombok.config
Related issues: 
- https://github.com/redhat-developer/vscode-java#2887
- https://github.com/eclipse-jdt/eclipse.jdt.core/issues#181

Related Eclipse N&N - https://www.eclipse.org/eclipse/news/4.27/jdt.php#ecj-separated-from-core

Steps to reproduce:

- download Eclipse 4.27M1 - https://download.eclipse.org/eclipse/downloads/drops4/S-4.27M1-202301041800/
- install m2e - https://download.eclipse.org/technology/m2e/snapshots/latest/
- add lombok-1.18.24 agent to the Eclipse VM arguments
- import project[vscode-java-issue.zip](https://github.com/redhat-developer/vscode-java/files/10441007/vscode-java-issue.zip)

You will see `LOG cannot be resolve`.
The issue can't be reproduced in Eclipse 4.26

snjeza:
Fixes #3332

### DIFF
--- a/src/core/lombok/eclipse/EclipseAST.java
+++ b/src/core/lombok/eclipse/EclipseAST.java
@@ -21,7 +21,10 @@
  */
 package lombok.eclipse;
 
+import static lombok.eclipse.handlers.EclipseHandlerUtil.*;
+
 import java.io.File;
+import java.lang.reflect.Constructor;
 import java.lang.reflect.Field;
 import java.lang.reflect.Method;
 import java.net.URI;
@@ -30,13 +33,6 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
 
-import lombok.core.AST;
-import lombok.core.LombokImmutableList;
-import static lombok.eclipse.handlers.EclipseHandlerUtil.*;
-import lombok.permit.Permit;
-
-import org.eclipse.core.resources.ResourcesPlugin;
-import org.eclipse.core.runtime.Path;
 import org.eclipse.jdt.internal.compiler.CompilationResult;
 import org.eclipse.jdt.internal.compiler.ast.ASTNode;
 import org.eclipse.jdt.internal.compiler.ast.AbstractMethodDeclaration;
@@ -55,11 +51,25 @@ import org.eclipse.jdt.internal.compiler.ast.TypeDeclaration;
 import org.eclipse.jdt.internal.compiler.ast.TypeReference;
 import org.eclipse.jdt.internal.compiler.ast.Wildcard;
 
+import lombok.core.AST;
+import lombok.core.LombokImmutableList;
+import lombok.permit.Permit;
+
 /**
  * Wraps around Eclipse's internal AST view to add useful features as well as the ability to visit parents from children,
  * something Eclipse own AST system does not offer.
  */
 public class EclipseAST extends AST<EclipseAST, EclipseNode, ASTNode> {
+	private static ClassLoader jdtCoreClassLoader;
+
+	public static ClassLoader getJdtCoreClassLoader() {
+		return jdtCoreClassLoader;
+	}
+
+	public static void setJdtCoreClassLoader(ClassLoader jdtCoreClassLoader) {
+		EclipseAST.jdtCoreClassLoader = jdtCoreClassLoader;
+	}
+
 	/**
 	 * Creates a new EclipseAST of the provided Compilation Unit.
 	 * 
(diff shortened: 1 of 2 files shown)


matter-it-does:
This could use some early returns https://softwareengineering.stackexchange.com/questions/18454/should-i-return-from-a-function-early-or-use-an-if-statement

Rawi01:
I think this can be simplified by loading `EclipseWorkspaceBasedFileResolver` with the right classloader and calling only the `resolve` method via reflection.

I tried the patch and it seems to work, but I'm not sure if the way you store the classloader in the `EclipseAst` is safe. It uses the classloaded of the patched `Parser` class and if it is part of `jdt.core` it might break. 

I have thought about this but have not found a perfect solution yet. What we need is some kind of global cache through which we can find the right classloader for the call. We have to consider that it is theoretically possible that a bundle occurs several times.

Rawi01:
@snjeza I was trying to come up with a simpler patch and ended up with this one. It also has the advantage that all the class loading stuff is in the same place.

```diff
diff --git a/src/eclipseAgent/lombok/launch/PatchFixesHider.java b/src/eclipseAgent/lombok/launch/PatchFixesHider.java
index c7bdbc3..bd318f9 100755
--- a/src/eclipseAgent/lombok/launch/PatchFixesHider.java
+++ b/src/eclipseAgent/lombok/launch/PatchFixesHider.java
@@ -200,6 +200,13 @@
 			if (TRANSFORM != null) return;
 			
 			Main.prependClassLoader(prepend);
+			try {
+				ClassLoader currentClassLoader = Transform.class.getClassLoader();
+				Method prependParentMethod = Util.findMethod(currentClassLoader.getClass(), "prependParent", ClassLoader.class);
+				prependParentMethod.invoke(currentClassLoader, prepend);
+			} catch (Throwable t) {
+				// Ignore
+			}
 			Class<?> shadowed = Util.shadowLoadClass("lombok.eclipse.TransformEclipseAST");
 			TRANSFORM = Util.findMethodAnyArgs(shadowed, "transform");
 			TRANSFORM_SWAPPED = Util.findMethodAnyArgs(shadowed, "transform_swapped");

```

snjeza:
@Rawi01 I have tried your patch. It doesn't fix the issue described at https://github.com/redhat-developer/vscode-java/issues/2887
lombok.eclipse.EclipseAST still can't resolve org.eclipse.core.resources.ResourcesPlugin.class - See https://github.com/projectlombok/lombok/blob/master/src/core/lombok/eclipse/EclipseAST.java#L191

Rawi01:
You are right, sorry for wasting your time. I used the `Permit` class in an earlier version but thought that it would be better to replace it with `Util`. No idea why I did not notice during my test that this breaks everything. Added an updated diff below.

Lombok adds a `ShadowClassLoader` to every module. Instead of loading the `RessourcePlugin` by hand my idea is to prepend the `jdt.core` SCL to the `compiler.batch` SCL. This should enable lombok to use classes of `jdt.core` without additional workarounds.

```diff
diff --git a/src/eclipseAgent/lombok/launch/PatchFixesHider.java b/src/eclipseAgent/lombok/launch/PatchFixesHider.java
index c7bdbc3..5404ee1 100755
--- a/src/eclipseAgent/lombok/launch/PatchFixesHider.java
+++ b/src/eclipseAgent/lombok/launch/PatchFixesHider.java
@@ -60,6 +60,8 @@
 import org.eclipse.jdt.internal.corext.refactoring.SearchResultGroup;
 import org.eclipse.jdt.internal.corext.refactoring.structure.MemberVisibilityAdjustor.IncomingMemberVisibilityAdjustment;
 
+import lombok.permit.Permit;
+
 import static lombok.eclipse.EcjAugments.ASTNode_generatedBy;
 
 /** These contain a mix of the following:
@@ -200,6 +202,13 @@
 			if (TRANSFORM != null) return;
 			
 			Main.prependClassLoader(prepend);
+			try {
+				ClassLoader currentClassLoader = Transform.class.getClassLoader();
+				Method prependParentMethod = Permit.getMethod(currentClassLoader.getClass(), "prependParent", ClassLoader.class);
+				Permit.invoke(prependParentMethod, currentClassLoader, prepend);
+			} catch (Throwable t) {
+				// Ignore
+			}
 			Class<?> shadowed = Util.shadowLoadClass("lombok.eclipse.TransformEclipseAST");
 			TRANSFORM = Util.findMethodAnyArgs(shadowed, "transform");
 			TRANSFORM_SWAPPED = Util.findMethodAnyArgs(shadowed, "transform_swapped");
```

snjeza:
@Rawi01 It works fine.

snjeza:
I have closed this PR.

Issue #3334, opened 2023-01-19
[BUG] extension methods cannot be called on this in javac (works in Eclipse)


marc-guenther:
**Describe the bug**
Calling an extension method on `this` does not work. Assigning `this` to a local variable first, and then calling the extension method works fine.

In this example, `.hello()` is defined as an extension method on `MyClass`, and `this` is also an instance of that class:
```
// This fails with: error: cannot find symbol
this.hello();
//  ^
//  symbol: method hello()
// 1 error

// but works perfectly fine, when I assign "this" to a local variable first
MyClass a = this;
a.hello();

// even "val" is no problem
val b = this;
b.hello();
```
This only happens with `javac`, Eclipse has no problem with it.

**To Reproduce**
See here for an example project:
https://github.com/marc-guenther/lombok-problems
Example is in the file [ProblemWithJavacButNotEclipse2.java](https://github.com/marc-guenther/lombok-problems/blob/master/prob/lems/ProblemWithJavacButNotEclipse.java). Run `make problem2` to see the output of `javac`.

**Expected behavior**
Two things:
- javac and Eclipse should show the same behaviour
- `this.someExtensionMethod()` of course should work.

**Version info (please complete the following information):**
- Lombok version: 1.18.24
- Platform: javac 1.8.0_282, Eclipse 2019-03 (4.11.0)

PR #3338, opened 2023-01-26
Allow calling extension methods on `this` as receiver

dembanakh:
Addresses #3334 

Currently in Javac, whenever the call receiver is `this`, it is not considered as a candidate for extension method call.
https://github.com/projectlombok/lombok/blob/8bb4dc5daa8792e718e88c596812f232a2b73e6e/src/core/lombok/javac/handlers/HandleExtensionMethod.java#L174

This PR makes the behaviour consistent with that of Eclipse's handler, I believe. In other words, explicit `this` call would be considered as a possible extension method call.

### DIFF
--- a/test/transform/resource/after-delombok/ExtensionMethodThis.java
+++ b/test/transform/resource/after-delombok/ExtensionMethodThis.java
@@ -1,0 +1,14 @@
+class ExtensionMethodThis {
+	public void test() {
+		ExtensionMethodThis.Extensions.hello(this);
+		hello();
+	}
+	
+	private void hello() {
+	}
+	
+	static class Extensions {
+		public static void hello(ExtensionMethodThis self) {	
+		}
+	}
+}
--- a/AUTHORS
+++ b/AUTHORS
@@ -13,6 +13,7 @@ Christoph Dreis <christoph.dreis@freenet.de>
 DaveLaw <project.lombok@apconsult.de>
 Dave Brosius <dbrosius@mebigfatguy.com>
 Dawid Rusin <dawidrusin90@gmail.com>
+Demian Banakh <dembanakh01@gmail.com>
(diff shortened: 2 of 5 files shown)


dembanakh:
@Rawi01 Could you take a look at this by any chance? 😃

test/transform/resource/after-delombok/ExtensionMethodThis.java
Rawi01, line 1: This test fails because the method name is different

Rawi01 (changes requested):
- Please add your name to the AUTHORS file
- It would be great if you can rebase this branch on the current master. We recently enabled GitHub Actions to run the whole test suite and that way we do not have to run the tests locally. (Some eclipse test will fail but you can ignore that)

I'm not sure if extension methods should ever work like that but it should at least be consistent. It feels a little bit weird because you could simply add the method to your class instead of using extension method at all.

dembanakh:
Thanks, I fixed the failing test and rebased.

4 test suites are failing (the same 4 that fail on master), but the others pass successfully, so I hope that's alright.

While it indeed feels strange to call an extension method on `this`, I think this option is worth keeping. Consider e.g. the case where you have to use the extension method both inside and outside the receiver's class
```java
@ExtensionMethod(Extensions.class)
class A {
  public void reset() {
    this.clear();
    // ...
  }
}

@ExtensionMethod(Extensions.class) 
class B {
  private A a;

  public void run() {
    a.clear();
    // ...
  }
}

class Extensions {
  public static void clear(A a) {}
}
```

Rawi01 (approved):

Rawi01:
In your example you could also add the `clear` method to `A` but this still might be useful if the base class is a third party/framework class.

PR #3368, opened 2023-03-12
Generate JCTree clones to prevent duplicate nodes

### LINKED ISSUE #3364: [BUG] Expression appears twice in the generated AST
**Describe the bug**

When the `@Builder` annotation is used, a `JCExpression` for the the build method return type is [created](https://github.com/projectlombok/lombok/blob/e5c324c615c1e304b2a9a64bc7f8608a67b0a74e/src/core/lombok/javac/handlers/HandleBuilder.java#L305). This is then passed to [generateBuildMethod](https://github.com/projectlombok/lombok/blob/e5c324c615c1e304b2a9a64bc7f8608a67b0a74e/src/core/lombok/javac/handlers/HandleBuilder.java#L738), which uses it twice, when generating a `new` expression and when generating the method.

I don't know if this is the best fix, but it looks like
```
returnType = cloneType(maker, returnType, job.sourceNode);
```
around line 798 fixes it.

**To Reproduce**

```
package com.mycompany.app;

import lombok.Builder;
import lombok.Singular;
import java.util.List;

public class App 
{
    public static void main( String[] args )
    {
        System.out.println( "Hello World!" );
    }
}

@Builder
class BuilderClass {
    @Singular
    private List<Integer> someIntegers;
}
```

**Expected behavior**

I expected that expression nodes would only have a single parent in the AST.

 - Lombok version: 1.18.26 and master (e5c324c615c1e304b2a9a64bc7f8608a67b0a74e)
 - Platform (javac or eclipse, and if so, what is the output of `javac -version` / the version listed in the _about..._ dialog of eclipse.: javac 17.0.6

**Additional context**

Rawi01:
This PR fixe fixes #3364

I also added a new automated check to `RunTestsViaDelombok` to scan all tests for duplicate nodes. I simplified the copy annotation methods which will probably lead to a merge conflict with #3356.


rzwitserloot:
Hmm, due to some rebasing/merging, this has been merged in but github doesn't recognize it as such. Closing, but, _this entire patch is in main branch now_.

PR #3376, opened 2023-03-20
Find references for extension methods

### LINKED ISSUE #3373: [BUG] Eclipse doesn't find usages of extension-methods in "find references" nor rename-refactoring
**Describe the bug**
When using methods made accessible via `@ExtensionMethod` these usages of the methods are neither found by "References in Workspace" (Ctrl-G), nor are they changed when refactoring such methods' names.

**To Reproduce**
- Make some methods available via `@ExtensionMethod`:

  ```
  @ExtensionMethod(MyExtensions.class) // containing a static method foo(Object o)
  class AnyClass {
    void anyMethod() {
      Object o = new Object();
      o.foo();
    }
  } 
  ```
- Variations of the problem
  -  "References in Workspace"
      - move cursor on `foo`
      - press Ctrl-G
      - Result: search for usages of foo() doesn't contain class `AnyClass`
  - "Refactor"/"Rename"
      - move cursor on `foo`
      - press Alt-Ctrl-R
      - type `bar` as a new name and press return
      - Result: method name gets changed in Class `MyExtensions`, but usage in `AnyClass` does not. Results in syntax error in class `AnyClass`.
  - Invocation from class `MyExtensions`
    -  go to class `MyExtensions`
    - move cursor on `foo`
    - invoke one of the above operation
    - both operation don't find/change anything in class `AnyClass`
      
**Expected behavior**
- usages of `foo()` in class `AnyClass` should be found
- renaming of `foo()`'s name should affect usage in class `AnyClass`

 - Lombok version: 1.18.26
 - Eclipse: 2023-03

Rawi01:
This PR fixes #3373

To find references eclipse uses an index that is based on method parameters. A extension method definition always has one parameter more than the reference. To find all matching methods the patch modifies the search pattern to always assume varargs. The search logic removes non-matching methods in a later step.

There might be some unintended side effects...

### DIFF
--- a/src/eclipseAgent/lombok/eclipse/agent/EclipsePatcher.java
+++ b/src/eclipseAgent/lombok/eclipse/agent/EclipsePatcher.java
@@ -918,6 +918,13 @@ public class EclipsePatcher implements AgentLauncher.AgentLaunchable {
 			.request(StackRequest.THIS)
 			.wrapMethod(new Hook(PATCH_EXTENSIONMETHOD_COMPLETIONPROPOSAL_PORTAL, "getJavaCompletionProposals", I_JAVA_COMPLETION_PROPOSAL_SIG, "java.lang.Object[]", "java.lang.Object"))
 			.build());
+		
+		sm.addScriptIfWitness(OSGI_TYPES, ScriptBuilder.wrapReturnValue()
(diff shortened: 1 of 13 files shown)


dstango:
thank you for quickly taking care of the issue! :-)

PR #3377, opened 2023-03-21
Add GitHub actions for tests

Rawi01:
This PR adds GitHub actions to run all tests for PRs and pushed commits.

This does not change the behaviour of commits or pull requests if you do not mark the master branch as protected and enable that pull requests have to pass status checks. Even if you enable it you can bypass this as administrator if you do not disable that.


rspilker:
Let's see if these checks filter out mistakes, thanks!

PR #3378, opened 2023-03-23
Remove eclipse specific classes in ExtensionMethod patch signatures

Rawi01:
#3376 added a method that uses a eclipse only class in its signature. This breaks ecj. This PR replaces the type in the signature with `j.l.Object`.


snjeza:
The PR fixes https://github.com/projectlombok/lombok/issues/3403 and https://github.com/redhat-developer/vscode-java/issues/3046

rspilker:
Thanks

testforstephen:
@Rawi01 @rspilker Thank you for the great fix. We would like to adopt this patch in the VS Code Java extension as soon as possible. Could you please include this patch in the 1.18.27 edge release? thanks.

PR #3392, opened 2023-04-02
Support multi-release jars in delombok

### LINKED ISSUE #3389: [BUG] Delombok fails to resolve module name of multi-release jar
**Describe the bug**

When a multi-release jar file with no module-info.class in the base version and no Automatic-Module-Name, but with a module-info.class in the version for the currently running JDK version, delombok fails to correctly resolve the module name according to the included module-info.class and falls back to the jars name. If the jar is not named the same as the module this will lead to to errors during delomboking.

**To Reproduce**

1. Find or create such a multi-release jar (for example `org.jetbrains:annotations:24.0.1`)
2. Create the following hierarchy:
 ```
├─ src
│  ├─ org
│  │  └─ example
│  │     └─ Main.java
│  └─ module-info.java
└─ lib
    ├─ annotations.jar
    └─ lombok.jar
```
with `Main.java`
```java
package org.example;

import lombok.Getter;

@Getter // prevents delombok from skipping this file
public class Main {
        System.out.println("Hello World!");
    }
}
```
and `module-info.java`
```java
module org.example {
    requires static org.jetbrains.annotations;
    requires static lombok;
}
```
3. Compile with `javac -d out -p lib/annotations.jar:lib/lombok.jar src/org/example/Main.java src/module-info.java`. This will be successful.
4. Now try delomboking the source files with `java -jar lib/lombok.jar delombok --module-path lib/annotations.jar -d delombok src`. This will produce the following error:
```
/path/to/src/module-info.java:2: error: module not found: org.jetbrains.annotations
    requires static org.jetbrains.annotations;
```
5. Change the `module-info.java` to
```java
module org.example {
    requires static annotations;
    requires static lombok;
}
```
6. Try delomboking as in step 4. This time there won't be any errors.
7. Try compiling as in step 3. This will produce the following error:
```
src/module-info.java:2: error: module not found: annotations
    requires static annotations;
                    ^
1 error
```

**Expected behavior**
Delombok should choose the correct version of the multi-release jar according to the current jdk version and correctly resolve the module name according the the module-info. Step 4 should not produce any errors. Step 6 should however.

**Version info:**
```shell
$ java --version
openjdk 19.0.1 2022-10-18
OpenJDK Runtime Environment Temurin-19.0.1+10 (build 19.0.1+10)
OpenJDK 64-Bit Server VM Temurin-19.0.1+10 (build 19.0.1+10, mixed mode, sharing)
```
```shell
$ java -jar lib/lombok.jar --version
v1.18.24 "Envious Ferret"
```

**Additional context**
`com.sun.tools.javac.main.Main::compile` uses the following code to set up the multi-release version
```java
// init multi-release jar handling
if (fileManager.isSupportedOption(Option.MULTIRELEASE.primaryName) == 1) {
    Target target = Target.instance(context);
    List<String> list = List.of(target.multiReleaseValue());
    fileManager.handleOption(Option.MULTIRELEASE.primaryName, list.iterator());
}
```
defaulting to the currently running version of the jdk (overridable with the hidden option `--multi-release <version>`). `lombok.delombok.Delombok` is missing similar code resulting in a fallback to the jar file name in `com.sun.tools.javac.file.Locations.ModulePathLocationHandler.ModulePathIterator::inferModuleName`.

Rawi01:
This PR fixes #3389 

I also updated delombok to return `1` if there is an error and updated the action to run the test during the build.


victorwss:
I had a problem in a project thta uses lombok and delombok goes nuts due to this bug, even if using the lastest edge version. So, I manually applied your patch to Delombok.java into lombok's edge, recompiled lombok and added it to my project. Worked like a charm! For me, this should be included in next version!

victorwss:
@Rawi01 Any reason why this was still not integrated? It is a great work.

reneleonhardt:
It's sad when contributors fixing old bugs are not even being appreciated for their efforts with a review... 😞

rzwitserloot:
@reneleonhardt You should like into how the XZ hack went down. It's entitled meanspirited whiners like you that hurt open source. You. You hurt open source. Cut it out.

Final notice, next complaint will get you banned from this project.

PR #3420, opened 2023-04-30
Ignore annotations during array detection 

### LINKED ISSUE #3416: [BUG] @ToString generates bad String if field marked as "@NonNull T @NonNull [] name"
**Describe the bug**
`@ToString` annotation generates bad String 

code:
```java
@Value
class Foo {
    @NonNull String @NonNull [] strings;
}
log.debug("Foo : {}", new Foo(new String[]{"foo", "bar"}));
```

Output:
```
Foo : Foo(strings=[Ljava.lang.String;@293bc4da)
```

**Version info:**
 - Lombok 1.18.26
 - javac 20.0.1

Rawi01:
This PR fixes #3416 

I also replaced the reflection code for type annotations in HandleEqualsAndHashcode with the handle methods and refactored the handler to use static final fields (allows JVM optimizations) and method guards (no exception for Java < 8).

The last commit fixes a problem with the Zulu JDK and the removal of record comopenent annotations.


mregulski:
It appears this also fixes #3434

PR #3425, opened 2023-05-10
Copy annotations on record components

### LINKED ISSUE #3315: [BUG] @JsonUnwrapped not copied to @Builder methods in Eclipse IDE for record
**Describe the bug**
In Eclipes IDE lombok fails to copy the Jackson annotation @JsonUnwrapped from a **record** to the method on @Builder's. This works fine outside of Eclipse IDE. This issue seems related to https://github.com/projectlombok/lombok/issues/2736

Notice, this seems to work fine if used with **class** rather than **record**

**To Reproduce**
Run the following test case, this will fail in Eclipse as the @JsonUnwrapped is not copied to the builder() class method inner(). The same test case will work outside of Eclipse.
```
package test;

import static org.assertj.core.api.Assertions.assertThat;

import org.junit.jupiter.api.Test;

import com.fasterxml.jackson.annotation.JsonUnwrapped;

import java.util.Arrays;

import lombok.Builder;
import lombok.extern.jackson.Jacksonized;

class RecordAnnotationsNotCopiedTest {
    @Builder
    @Jacksonized
    record Example(@JsonUnwrapped Inner inner) {}
    
    record Inner(String s) {}
    
    @Test
    void testLombok() throws NoSuchMethodException, SecurityException {
        var method = Arrays.asList(Example.builder().getClass().getMethods()).stream().filter(m -> m.getName().equals("inner")).findFirst().get();
        assertThat(method.getAnnotations().length).isEqualTo(1);

    }
}
```

**Expected behavior**
This should also work in Eclipse IDE.

 - Lombok v1.18.24
 - Platform Eclipse  Version: 2022-06 (4.24.0)

Rawi01:
This PR fixes #3315

There are some failing unrelated tests that were fixed in https://github.com/projectlombok/lombok/commit/1599bc1ce4a2440322af7a1183828c5f65dcc237.

### DIFF
--- a/src/core/lombok/eclipse/handlers/EclipseHandlerUtil.java
+++ b/src/core/lombok/eclipse/handlers/EclipseHandlerUtil.java
@@ -809,12 +809,12 @@ public class EclipseHandlerUtil {
 	 * Searches the given field node for annotations and returns each one that is 'copyable' (either via configuration or from the base list).
 	 */
 	public static Annotation[] findCopyableAnnotations(EclipseNode node) {
-		AbstractVariableDeclaration avd = (AbstractVariableDeclaration) node.get();
-		if (avd.annotations == null) return EMPTY_ANNOTATIONS_ARRAY;
 		List<Annotation> result = new ArrayList<Annotation>();
 		List<TypeName> configuredCopyable = node.getAst().readConfiguration(ConfigurationKeys.COPYABLE_ANNOTATIONS);
-		
-		for (Annotation annotation : avd.annotations) {
+		for (EclipseNode child : node.down()) {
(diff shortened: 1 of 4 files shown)


velo:
Hi @Rawi01 I have a PR with a test for `@JsonUnwrapped` and it doesn't seem to fix it

https://github.com/projectlombok/lombok/pull/3423

I originally found the issue cause I use `record` classes, but made an example with regular class cause I was also able to reproduce it there

Rawi01:
@velo Seems like your issue is different. This PR fixes a general eclipse problem affecting all annotations on record components.

PR #3475, opened 2023-08-04
Add option to mark SuperBuilder as root

mzcu:
I have a kotlin abstract class (coming from a shared library) that I extend in java code which makes an attempt to use `@SuperBuilder` annotation.

This doesn't work because, as explained in the javadoc, all classes in the hierarchy (well, at least ones which are present in `extends` clauses of annotated classes)  must be annotated with `@SuperBuilder`.

This change adds option to mark a class as the root of a super-builder hierarchy: `@SuperBuilder(root=true)`, which relaxes the constraint mentioned in the previous paragraph.

In case this fits well with the existing design and you indicate that you are willing to merge, I'd be happy to polish it and update any docs or tests which I missed.

### DIFF
--- a/src/core/lombok/experimental/SuperBuilder.java
+++ b/src/core/lombok/experimental/SuperBuilder.java
@@ -33,7 +33,8 @@ import lombok.Singular;
  * The SuperBuilder annotation creates a so-called 'builder' aspect to the class that is annotated with {@code @SuperBuilder}, but which works well when extending.
  * It is similar to {@code @Builder}, except it is only legal on types, is less configurable, but allows you to {@code extends} other builder-able classes.
  * <p>
- * All classes in the hierarchy must be annotated with {@code @SuperBuilder}.
+ * All classes in the hierarchy must be annotated with {@code @SuperBuilder}. To relax this requirement, set {@link SuperBuilder#root()} to {@code true}
+ * on the top-most class in the hierarchy which should have a generated builder.
  * <p>
  * Lombok generates 2 inner 'builder' classes, which extend the parent class' builder class (unless your class doesn't have an extends clause).
  * Lombok also generates a static method named {@code builder()}, and a protected constructor that takes 1 argument of the builderclass type.
@@ -81,4 +82,12 @@ public @interface SuperBuilder {
 	 * @return The prefix to prepend to generated method names.
 	 */
 	String setterPrefix() default "";
+
+	/**
+	 * If <code>true</code>, ignore extends clause of the annotated class and do not produce builders for any of its
+	 * superclasses. This will only work if superclasses of the annotated class have default constructors.
+	 *
+	 * @return Whether to stop with super-builder generation at this class.
+	 */
+	boolean root() default false;
 }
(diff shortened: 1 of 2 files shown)


janrieke:
I understand your use case. In fact, I have [answered a similar question on StackOverflow](https://stackoverflow.com/a/61763012/10026074).

What you purpose is not a real solution to the problem, it's just a hack to make the code compile in some cases (e.g. you must have a default constructor on the extended class). What you'll get (in those compiling cases) is something that looks like a builder, but is incomplete: it lacks setter methods for the fields of the superclasses.

Lombok should never generate such code.
You can still implement the manual solution linked above, but then you know the risks and can take precautions that it doesn't break expectations.

I'm sorry, but I don't see a way how such a feature could be implemented in a way that doesn't cause hundreds of bug reports.

mzcu:
First of all, thanks for maintaining Lombok!

> What you'll get (in those compiling cases) is something that looks like a builder, but is incomplete: it lacks setter methods for the fields of the superclasses. Lombok should never generate such code

Yes, and that's fine because user explicitly requested this by setting a flag

> You can still implement the manual solution linked above, but then you know the risks and can take precautions that it doesn't break expectations

What is the advantage of delomboking, modifying the builder manually and checking-in auto-generated code over setting a flag? In both cases one is explicitly stating their intentions. Of course it can be better customized by writing custom code, but that's a truism.

In the end, it really depends on number of users who face similar issue in the wild. Adding complexity to the feature would be warranted only if there are enough problem reports similar to this one

> I'm sorry, but I don't see a way how such a feature could be implemented in a way that doesn't cause hundreds of bug reports.

Can't argue with that, you are the one who is getting the bug reports

janrieke:
Just a few more words on why I vote against such a feature. To boil it down, your proposed solution has two major problems:
1. It produces an incomplete builder, so you have to customize it in almost all cases.
2. It produces compile errors in many cases, and lombok users will be unable to fix them without customization.

Just one of those problems would be enough to not include it. Lombok is primarily a "boilerplate buster", which means its features should work without customization in the vast majority of the cases. As you said: There is no advantage in delomboking/customization, but your proposed solution often does require it, too.

If you look at the linked StackOverflow answer, you see that technically speaking this solution does not require customization of lombok-generated code. It inserts a new class which contains all of the builder logic that's missing because of the absent `@SuperBuilder` on the superclasses. That makes it easy to, e.g., write an ArchUnit test that ensures every field of the superclass has a matching builder setter. That's what I had in mind when writing "take precautions". Of course, you could all do that with your solution in a customized `@SuperBuilder`, but then: why add an error-prone feature if there is a (similarly complex) working solution already?

mklueh:
Hello,

I'm running into the same problem using Quarkus Panache, where my BaseEntity extends PanacheEntityBase (class from Quarkus Framework) and all my entities are extending the BaseEntity. 

**I can't use the Builder pattern for any of my entities because of this restriction.**

For me, it would be **perfectly fine**, if the fields of PanacheEntityBase **were not exposed** via builder at all and thus the **builder would be "incomplete"**, as it is just the primary id. 

This class does have the implicit default constructor.

Instead of having some extra attribute like proposed `@SuperBuilder(root=true)`, **why can't Lombok just assume the last `@SuperBuilder` annotated class in the hierarchy as the root, not matter what and just fail if the extended root class does not have a default constructor?**

PR #3487, opened 2023-08-21
Patch-1-NonNull-VanillaJava-Example

xandersavvy:
As we are using vanilla js example here ,the vanilla js non null should not be there


Rawi01:
This is intentional, there are multiple closed PRs for this already (e.g. #3161)

PR #3494, opened 2023-08-31
Remove javac support for getter on enum members

### LINKED ISSUE #3427: [BUG] Eclipse handlers don't support enum constants
**Describe the bug**
For enum constants JDT leaves null TypeReference in `org.eclipse.jdt.internal.compiler.ast.FieldDeclaration`, but `lombok.eclipse.handlers.HandleGetter` doesn't ready for it so there will be several problems:

1. NPE at [isBoolean](https://github.com/projectlombok/lombok/blob/master/src/core/lombok/eclipse/handlers/EclipseHandlerUtil.java#L1425)
2. Null return type at [createGetter](https://github.com/projectlombok/lombok/blob/master/src/core/lombok/eclipse/handlers/HandleGetter.java#L241) so it will be generated as void method
3. [Not static modifier ](https://github.com/projectlombok/lombok/blob/master/src/core/lombok/eclipse/handlers/HandleGetter.java#L202)

Looks like It can be reproduced with other annotation handlers.

**To Reproduce**
Case:
```
public enum EnumTest {
    @Getter
    AAA
}
```
Result:
```
Lombok annotation handler class lombok.eclipse.handlers.HandleGetter failed
java.lang.NullPointerException
at lombok.eclipse.handlers.EclipseHandlerUtil.isBoolean(EclipseHandlerUtil.java:1425)
at lombok.eclipse.handlers.HandleGetter.createGetterForField(HandleGetter.java:193)
at lombok.eclipse.handlers.HandleGetter.createGetterForFields(HandleGetter.java:164)
at lombok.eclipse.handlers.HandleGetter.handle(HandleGetter.java:153)
at lombok.eclipse.HandlerLibrary$AnnotationHandlerContainer.handle(HandlerLibrary.java:106)
at lombok.eclipse.HandlerLibrary.handleAnnotation(HandlerLibrary.java:237)
at lombok.eclipse.TransformEclipseAST$AnnotationVisitor.visitAnnotationOnField(TransformEclipseAST.java:245)
at lombok.eclipse.EclipseNode.traverse(EclipseNode.java:110)
```

**Expected behavior**
Javac handler works well in this case and generate method:
```
public static EnumTest getAAA() {
    return AAA;
}
```

**Version info:**
Platform Eclipse: 

-  Version: 2023-03 (4.27.0)
-  Build id: 20230309-1520
-  Lombok v1.18.26 "Envious Ferret" is installed

Rawi01:
This PR fixes #3427

This is a breaking change for `javac` but I cannot imagine why someone want to generate a static getter for an enum member.

### DIFF
--- a/src/core/lombok/eclipse/handlers/HandleGetter.java
+++ b/src/core/lombok/eclipse/handlers/HandleGetter.java
@@ -168,7 +168,7 @@ public class HandleGetter extends EclipseAnnotationHandler<Getter> {
 	public void createGetterForField(AccessLevel level,
 			EclipseNode fieldNode, EclipseNode errorNode, ASTNode source, boolean whineIfExists, boolean lazy, List<Annotation> onMethod) {
 		
-		if (fieldNode.getKind() != Kind.FIELD) {
+		if (fieldNode.getKind() != Kind.FIELD || fieldNode.isEnumMember()) {
 			errorNode.addError(GETTER_NODE_NOT_SUPPORTED_ERR);
 			return;
 		}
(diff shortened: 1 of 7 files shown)


rzwitserloot:
Any 'breaking change' that only breaks obviously buggy / 'just to make a point' code is, in my book, actually a good thing :) Better to throw a compiler error in a coder's face than silently compiling gobbledygook.

PR #3520, opened 2023-10-11
Feature/lombok tostring exclude

95sharmavarun:
In this change, I've created a new method level annotation "ExcludeToString" which will add support for records in @ExcludeToString  annotation to exclude params to not show in toString method.


Rawi01:
This is a spring aspect and not a lombok handler. If you want to exclude fields you can already do that.

Issue #3305, opened 2022-11-24
Undocumented (or confusing) behaviour of  Lombok generator with org.jetbrains.annotations.NotNull


trokhinas:
I tried to use `org.jetbrains.annotations.NotNull` for static code analyse by IDEA like this:
```
import lombok.Data;
import lombok.NoArgsConstructor;
import org.jetbrains.annotations.NotNull;
import org.jetbrains.annotations.Nullable;

@Data
@NoArgsConstructor
public class LombokTest {
    @NotNull
    private String field1;

    @Nullable
    private String field2;

    public static void main(String[] args) {
        final LombokTest test = new LombokTest();
        test.setField1(null); // <-- here throws an exception
    }
}
```

But when i tries to call `test.setField1(null)` i got an exception:
```
Exception in thread "main" java.lang.NullPointerException: field1 is marked non-null but is null
```
It seems like Lombok intepret any NotNull annotation for generating null checks or it is a bug? If i would want to add null checks i would use `lombok.NonNull` that was designed specifically to GENERATE checks in source code. But i just want to use advantages of static analyse by IDEA and do not want any real null checks.

trokhinas:
I am using Lombok 1.18.22 and Lombok 1.18.12.

marc-guenther:
Different people like different behaviour. See #3294 for comparison. ;)

Also I'm confused now. Apparently `org.jetbrains.annotations.NotNull` and `javax.validation.constraints.NotNull` are treated differently by Lombok?

trokhinas:
@marc-guenther, integration with other libraries and frameworks is always cool. But as I said, if I want to force null checks on my objects, I would rather use `lombok.NonNull` as it was designed for that. First I want to find out if there is any documentation for this function or if it is an unexpected lombok generator behavior and a bug.

The only solution I've found so far is to "de-lombok" this code with IDEA.

trokhinas:
In the lombok changelog, the latest entry states that 'BUGFIX: Turns out treating @NotNull as an annotation that indicates lombok should generate nullcheck guards causes all sorts of problems.' in [v0.10.2 (November 1st, 2011)](https://projectlombok.org/changelog#:~:text=v0.10.2%20(November%201st%2C%202011)). But nothing about `org.jetbrains.annotations.NotNull`. Is there a chance that the problem is related to the same name for the annotations?

prdoyle:
It is documented [here](https://projectlombok.org/features/NonNull):

[...]

PR #3525, opened 2023-10-18
Fix confusing behaviour of NonNull annotations.

rPraml:
See #3305 

This PR fixes the inconsistent behaviour, that non-lombok `@NonNull` annotations are honored at fields, but not on method/constructor parameters.

It also adds a config option to completely turn off or on the use of foreign nonNull annotations

### DIFF
--- a/src/core/lombok/core/handlers/HandlerUtil.java
+++ b/src/core/lombok/core/handlers/HandlerUtil.java
@@ -78,8 +78,10 @@ public class HandlerUtil {
 	public static int primeForNull() {
 		return 43;
 	}
-	
-	public static final List<String> FOREIGN_NONNULL_ANNOTATIONS, LOMBOK_NONNULL_ANNOTATIONS, BASE_COPYABLE_ANNOTATIONS, COPY_TO_SETTER_ANNOTATIONS, COPY_TO_BUILDER_SINGULAR_SETTER_ANNOTATIONS, JACKSON_COPY_TO_BUILDER_ANNOTATIONS;
+
+	private static final List<String> NONNULL_ANNOTATIONS, LOMBOK_NONNULL_ANNOTATIONS;
+
+	public static final List<String> BASE_COPYABLE_ANNOTATIONS, COPY_TO_SETTER_ANNOTATIONS, COPY_TO_BUILDER_SINGULAR_SETTER_ANNOTATIONS, JACKSON_COPY_TO_BUILDER_ANNOTATIONS;
 	static {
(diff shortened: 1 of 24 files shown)


src/core/lombok/eclipse/HandlerLibrary.java
```diff
@@ -219,9 +219,15 @@ public long handleAnnotation(CompilationUnitDeclaration ast, EclipseNode annotat
 		TypeResolver resolver = new TypeResolver(annotationNode.getImportList());
 		TypeReference rawType = annotation.type;
 		if (rawType == null) return Long.MAX_VALUE;
-		
+
 		String fqn = resolver.typeRefToFullyQualifiedName(annotationNode, typeLibrary, toQualifiedName(annotation.type.getTypeName()));
-		if (fqn == null) return Long.MAX_VALUE;
+		if (fqn == null) {
+			if (isNonNullAnnotation(annotationNode, annotation)) {
+				fqn = lombok.NonNull.class.getName();
```
rPraml, line 228: This fixes the inconsistent behaviour mentioned in https://github.com/projectlombok/lombok/issues/3305#issuecomment-1765993862

We run in this code path, if an unknown annotantion is found on an method or constructor paramerter. In this case we check, if this could be an alternative nonNull annotation. If so, we treat it like the lombok.NonNull annotation.

src/core/lombok/core/handlers/HandlerUtil.java
```diff
@@ -980,4 +989,13 @@ public static String getParamJavadoc(String methodComment, String param) {
...
 		return null;
 	}
+
+	/**
+	 * Determines the list of non-null annotations that should be used.
+	 * @param useForeignAnnotations the config parameter, if we want to use foreign non-null annotations (default: true).
+	 * @return depending on the config, it either returns a list of all known non-null annotations or just containing the lombok.NonNull annotation
+	 */
+	public static List<String> nonNullAnnotations(Boolean useForeignAnnotations) {
+		return Boolean.FALSE.equals(useForeignAnnotations) ? LOMBOK_NONNULL_ANNOTATIONS : NONNULL_ANNOTATIONS;
```
rPraml, line 999: Here is the switch logic, which non-null annotations should lombok use. By default, it uses all known nonNull-annotations. if `lombok.nonNull.useForeignAnnotations` is set to false, only the lombok.NonNull annotations (there is currently only one, but we are prepared, if an experimental one is added)

So people, who do not want that lombok use foreign annotations can turn it off now.

It would be also possible to specify a list of nonNull annotations in the configuration. What do you think?

rzwitserloot:
This change would be backwards incompatible and is therefore vetoed.

PR #3550, opened 2023-11-19
[FEATURE] Add internString boolean flag on @Setter Annotation to reuse JVM string pool if you want

shlteater:
**Describe the feature**
Add internString boolean flag on @Setter Annotation. 

```Java
	/**
	 * If you want your setter to invoke {@link java.lang.String#intern() String.intern()} on input string to reuse JVM string pool, you can turn this to true.
	 * Note: Can only take effect to String field. And a field level @Setter annotation can override type level @Setter annotation on this feature.
	 * 
	 * @return Whether or not use {@link java.lang.String#intern() String.intern()} on input string to reuse JVM string pool.
	 */
	boolean internString() default false;
```

Example1
```
import lombok.Setter;

@Setter(internString = true)
public class SetterInternOnClass {
	private String str1;
    private java.lang.String str2;
	private int int1;
}
```
Will be compiled as 
```
public class SetterInternOnClass {
	private String str1;
	private java.lang.String str2;
	private int int1;

	public void setStr1(final String str1) {
		this.str1 = str1 == null ? null : str1.intern();
	}

	public void setStr2(final java.lang.String str2) {
		this.str2 = str2 == null ? null : str2.intern();
	}

	public void setInt1(final int int1) {
		this.int1 = int1;
	}
}
```

Example2
```
import lombok.Setter;

@Setter
public class SetterInternOnField {
	@Setter(internString = true)
    private String str1;
	@Setter(internString = true)
    private java.lang.String str2;
	private String str3;
}
```
Will be compiled as 
```
public class SetterInternOnField {
	private String str1;
	private java.lang.String str2;
	private String str3;

	public void setStr3(final String str3) {
		this.str3 = str3;
	}

	public void setStr1(final String str1) {
		this.str1 = str1 == null ? null : str1.intern();
	}
	
	public void setStr2(final java.lang.String str2) {
		this.str2 = str2 == null ? null : str2.intern();
	}
}
```

If u have a large number of Java POJOs in you project memory.
And the POJO has some string field that cannot be simply treated hard coded as java enum.
U may have a choice to reduce memory footprint in this case.

Example
```
import lombok.Setter;

@Setter
public class BankAccount {
    private String accountNumber;
    @Setter(internString = true)
    private String currencyCode;
}
```
All cuurenyCodes in the world may change sometimes and cannot be hard coded as enum.
We can get into this.

```
import lombok.Setter;

@Setter
public class BankAccount {
    private String accountNumber;
    @Setter(internString = true)
    private String currencyCode;

        public void setAccountNumber(final String accountNumber) {
        this.accountNumber= accountNumber;
    }

        public void setCurrencyCode(final String currencyCode) {
        this.currencyCode= currencyCode== null ? null : currencyCode.intern();
        }
    }
```

### DIFF
--- a/src/core/lombok/Setter.java
+++ b/src/core/lombok/Setter.java
@@ -92,4 +92,12 @@ public @interface Setter {
 	@Retention(RetentionPolicy.SOURCE)
 	@Target({})
 	@interface AnyAnnotation {}
+	
+	/**
+	 * If you want your setter to invoke {@link java.lang.String#intern() String.intern()} on input string to reuse JVM string pool, you can turn this to true.
+	 * Note: Can only take effect to String field. And a field level @Setter annotation can override type level @Setter annotation on this feature.
+	 * 
+	 * @return Whether or not use {@link java.lang.String#intern() String.intern()} on input string to reuse JVM string pool.
+	 */
+	boolean internString() default false;
 }
--- a/src/core/lombok/eclipse/handlers/HandleBuilder.java
+++ b/src/core/lombok/eclipse/handlers/HandleBuilder.java
@@ -1050,7 +1050,7 @@ public class HandleBuilder extends EclipseAnnotationHandler<Builder> {
 		Annotation[] methodAnns = EclipseHandlerUtil.findCopyableToSetterAnnotations(bfd.originalFieldNode);
 		if (methodAnns != null && methodAnns.length > 0) methodAnnsList = Arrays.asList(methodAnns);
 		ASTNode source = job.sourceNode.get();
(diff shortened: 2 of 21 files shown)


Geker:
String.intern()  have some performance issues, i don't think this is really necessary

rzwitserloot:
Lombok's purpose is all about boilerplate. Boilerplate boils (heheheh) down to two properties:

1. The code you have to write is formulaic. A very simple robot / script could do it.
2. The job is __common__.

This feature does not clear the bar: The problem is that bolded word.

You'd think: Eh, so what, even if it helps only a few people, it helps em, right?

No, for two reasons:

1. People who use lombok need to know this feature exists or at least their eyeballs need to pass over it when perusing the docs.
2. We need to maintain it.

These are very small costs. But the benefit is also very small. It's about a value/cost proposition. Both are low, but in net balance, the 'cost' side is higher than the 'benefit' side. Hence, regrettably, PR denied.

PR #3552, opened 2023-11-19
Prevent inlining for OperatorId constants 

### LINKED ISSUE #3536: [BUG] Annotations generate incorrect operators in Eclipse
**Describe the bug**
Lombok is using an outdated version of Eclipse JDT Core, causing an error in reading some OperatorIds in Eclipse's AST.
Eclipse recently changed the OperatorIds to resolve a bug seen here: https://github.com/eclipse-jdt/eclipse.jdt.core/issues/316.
with the corresponding change here: https://github.com/eclipse-jdt/eclipse.jdt.core/commit/80a104f07bfab34a2f8f49c1b8993de035d116ab
This results in the OperatorId for "!=" (int 29 / bin 11101) truncating to the value of "+" (int 14 / bin 1110) when read by Eclipse.

**To Reproduce**
Project reproducing the issue: 
[test.tar.gz](https://github.com/projectlombok/lombok/files/13252803/test.tar.gz)

1. Install Eclipse 4.29, Lombok 1.18.30, and ASTView plugin
2. View AST of Person.java, find lombok's generated "equals()" method. "other.getId() != null" will have an operator value of "+", seen below:
image

**Expected behavior**
Operator should be "!=" for source generated by Data annotation.

 - Lombok Version: 1.18.30
 - Eclipse Version: 2023-09 (4.29.0)

**Additional context**
None

Rawi01:
This PR  fixes #3536

The new stub utilizes conditional expressions to prevent inlining and force the resolution of the constants at runtime. Additionally I added the latest eclipse as test target and updated the pipeline to use these targets. To standardize the generated eclipse test output between different versions, lombok now hides implicit constructors in the print method.

### DIFF
--- a/.github/workflows/ant.yml
+++ b/.github/workflows/ant.yml
@@ -24,9 +24,9 @@ jobs:
           path: |
             ivyCache
             lib
-          key: ${{ runner.os }}-${{ hashFiles('**/ivy.xml') }}
+          key: ivy-${{ hashFiles('**/ivy.xml') }}
           restore-keys: |
-            ${{ runner.os }}-
+            ivy-
         
       - name: Build with Ant
         run: ant -noinput dist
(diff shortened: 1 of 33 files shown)


rspilker:
Initially I was looking for different values for `!=`, until I realized that this change only affects the compiled code and isn't used at runtime. Nice trick.

PR #3560, opened 2023-12-01
Explain builder default gotchas

mjaggard:


website/templates/features/Builder.html
```diff
@@ -66,7 +66,7 @@
...
+			Finally, applying <code>@Builder</code> to a class is as if you added <code>@AllArgsConstructor(access = AccessLevel.PACKAGE)</code> to the class and applied the <code>@Builder</code> annotation to this all-args-constructor. This only works if you haven't written any explicit constructors yourself or created one with @NoArgsConstructor. If you do have an explicit constructor, put the <code>@Builder</code> annotation on the constructor instead of on the class. Note that if you put both `@Value` and `@Builder` on a class, the package-private constructor that `@Builder` wants to generate 'wins' and suppresses the constructor that `@Value` wants to make.
```
feech, line 69: chatGPT rephrases it:
Applying <code>@Builder</code> to a class is equivalent to adding <code>@AllArgsConstructor(access = AccessLevel.PACKAGE)</code> to the class and applying the <code>@Builder</code> annotation to this all-args constructor. This mechanism is effective only when explicit constructors or constructors created with <code>@NoArgsConstructor</code> are absent. If an explicit constructor is present, it's recommended to place the <code>@Builder</code> annotation on the constructor instead of the class. It's worth noting that when both <code>@Value</code> and <code>@Builder</code> are used in a class, the package-private constructor intended by <code>@Builder</code> takes precedence and suppresses the constructor specified by <code>@Value</code>

As non-native speacker i find it's easier to understand.

feech:
Thanks, Mat
It looks good to me.

rzwitserloot:
I accepted this then immediately updated the text. The gotcha on the gotcha is: `@RequiredArgsConstructor` can also cause this 'issue' if you try real hard (by using `@NonFinal`, for example). See commit 84cb36d25.

PR #3572, opened 2023-12-21
Allow CI workflow to test against JDK 22-ea

mmoayyed:


Rawi01:
This change doesn't work because the `setup-java` action does not support early access versions. That's why we use a seperate step for the early access test setup.

Rawi01:
This was added

PR #3584, opened 2024-01-14
Adding support for SequencedCollection, SequencedSet and SequencedMap from JDK 21

msamoylych:
Adding support for `java.util.SequencedCollection`, `java.util.SequencedSet` and `java.util.SequencedMap` from JDK 21 for `@Singular` annotation (`@Builder`)

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -33,6 +33,7 @@ Mart Hagenaars <marthagenaars@gmail.com>
 Martin O'Connor <38929043+martinoconnor@users.noreply.github.com>
 Martin Panzer <postremus1996@googlemail.com>
 Mateusz Matela <mateusz.matela@gmail.com>
+Maxim Samoylych <msamoylych@gmail.com>
 Michael Dardis <git@md-5.net>
 Michael Ernst <mernst@alum.mit.edu>
 Michiel Verheul <cheelio@gmail.com>
--- a/src/core/lombok/core/GuavaTypeMap.java
+++ b/src/core/lombok/core/GuavaTypeMap.java
@@ -37,7 +37,10 @@ public final class GuavaTypeMap {
 		m.put("java.util.Map", "ImmutableMap");
 		m.put("java.util.Collection", "ImmutableList");
 		m.put("java.util.List", "ImmutableList");
-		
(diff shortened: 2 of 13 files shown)


rspilker:
Thanks for the contribution! I haven't reviewed the whole code yet.

I did copy the java 22 gradle fix directly to master, thanks for that solution, because it's not part of this fix. Please remove it from this pull request.

Rawi01:
- You have disabled the Eclipse test. My guess is that you did this to avoid the error messages about missing classes. There are two ways to handle this:
  * Add the test, mark the errors as expected and call it a day
  * Add the current runtime to the eclipse compiler classpath. IIRC this will cause some split module errors. The last time I managed to workaround it but I think in this case we can no longer do so and have to change something somewhere
- I think the mapping between the sorted guava type and the sequenced collections is incorrect. There also is no test for this, do we ever call this?
- +1 for updating the website

msamoylych:
> You have disabled the Eclipse test. My guess is that you did this to avoid the error messages about missing classes. There are two ways to handle this:
Add the test, mark the errors as expected and call it a day
Add the current runtime to the eclipse compiler classpath. IIRC this will cause some split module errors. The last time I managed to workaround it but I think in this case we can no longer do so and have to change something somewhere

Eclipse IDE 2023-12 test uses JDK 21 unless specified `compiler.compliance.level` but in [ant.yml](https://github.com/projectlombok/lombok/blob/fd55c7aa93f3afc08c8694d3f9e59032cf42aedf/.github/workflows/ant.yml#L125C28-L125C28) is specified JDK 17 to run this test. Therefore [BuilderSingularSequencedCollections](test/transform/resource/before/BuilderSingularSequencedCollections.java) test performs (`// version 21:`) and fails due to error messages about missing classes in JDK 17.
I added the test with expected errors

> I think the mapping between the sorted guava type and the sequenced collections is incorrect. There also is no test for this, do we ever call this?

There are other suitable classes in Guava now. I think [BuilderSingularRedirectToGuava](https://github.com/projectlombok/lombok/blob/master/test/transform/resource/before/BuilderSingularRedirectToGuava.java) test is sufficient

reneleonhardt:
Awesome contribution! Is the new code covered and ready to be merged after you added `@Generated`?

PR #3589, opened 2024-01-17
Add NonNull annotations to test cases for testing annotation copying

### LINKED ISSUE #3318: [BUG] Custom nullable annotations are not copied to constructors and methods for Value/Data classes
**Describe the bug**
In my project and libs I use custom nullability annotations that are almost one-to-one copies of spring ones, the purpose is to avoid depending on spring-core. I wanted to configure lombok to add them from fields to constructors and methods (getters/setters) for Value/Data annotated classes, but for some reason annotations are not added.

At the same time, when lombok is configured to use `spring` or `checkerframework` annotations they are added correctly.

In other words, sample class
```java
@Value
public class Sample {
    @Nullable
    String prop1;
    String prop2;
}
```

Produce different results depending on `Nullable` annotation.

For spring with (`lombok.addLombokGeneratedAnnotation=true, lombok.addNullAnnotations=spring`)
```
import org.springframework.lang.NonNull;
import org.springframework.lang.Nullable;

public final class Sample {
    @Nullable
    private final String prop1;
    private final String prop2;

    @Generated
    public Sample(@Nullable String prop1, String prop2) {
        this.prop1 = prop1;
        this.prop2 = prop2;
    }

    @Nullable
    @Generated
    public String getProp1() {
        return this.prop1;
    }
```
For custom (`lombok.addLombokGeneratedAnnotation=true, lombok.addNullAnnotations=CUSTOM:com.example.annotations.NonNull:com.example.annotations.Nullable`)
```
import com.example.annotations.NonNull;
import com.example.annotations.Nullable;
import lombok.Generated;

public final class Sample {
    @Nullable
    private final String prop1;
    private final String prop2;

    @Generated
    public Sample(String prop1, String prop2) {
        this.prop1 = prop1;
        this.prop2 = prop2;
    }

    @Generated
    public String getProp1() {
        return this.prop1;
    }
```

**To Reproduce**
I've prepared a sample project that shows the problem: https://github.com/zeldigas/lombok-custom-nullable-problem. In `spring` module you can find usage of spring annotations and in `custom` - custom ones that are defined in this module as well (for simplicity)

**Expected behavior**
Both modules in sample project are working the same way.

 - Lombok version - `1.18.24`
 - Platform - javac temurin-11.0.17 on Windows 10

burdoto:
Closes https://github.com/projectlombok/lombok/issues/3318

This PR introduces `@NonNull` annotations as part of `@Data` and `@Value` classes' fields in order to be copied from fields to constructors, getters and setters as requested in https://github.com/projectlombok/lombok/issues/3318#issuecomment-1894757426.

### DIFF
--- a/test/transform/resource/after-delombok/DataPlain.java
+++ b/test/transform/resource/after-delombok/DataPlain.java
@@ -1,6 +1,6 @@
 class Data1 {
 	final int x;
-	String name;
+	@lombok.NonNull String name;
 	@java.lang.SuppressWarnings("all")
 	public Data1(final int x) {
 		this.x = x;
@@ -9,12 +9,13 @@ class Data1 {
 	public int getX() {
 		return this.x;
 	}
+	@lombok.NonNull
 	@java.lang.SuppressWarnings("all")
 	public String getName() {
 		return this.name;
 	}
 	@java.lang.SuppressWarnings("all")
-	public void setName(final String name) {
+	public void setName(@lombok.NonNull final String name) {
 		this.name = name;
 	}
 	@java.lang.Override
--- a/test/transform/resource/after-delombok/ValuePlain.java
+++ b/test/transform/resource/after-delombok/ValuePlain.java
@@ -1,8 +1,9 @@
 final class Value1 {
 	private final int x;
+	@lombok.NonNull
 	private final String name;
 	@java.lang.SuppressWarnings("all")
-	public Value1(final int x, final String name) {
+	public Value1(final int x, @lombok.NonNull final String name) {
 		this.x = x;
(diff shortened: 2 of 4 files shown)


burdoto:
After writing and testing the modified Test case, I noticed that this does in fact function with the `@lombok.NonNull` annotation.
The reported issue applies to external annotations such as `org.intellij.lang.annotations.Language` (as reported in https://github.com/projectlombok/lombok/issues/3318#issuecomment-1886710636) that are affected. However, for the test case I dont have enough project insight to correctly introduce external annotations to this test case (let alone since maven fails to import the whole project).

PR #3604, opened 2024-02-12
Fix type detection of nested annotations

facboy:
Fix type detection of annotations that are nested within the same (or other) classes. I came across this bug when trying to copy an annotation from a field to a constructor, where that annotation was an inner class of the field that the annotated class belonged to, eg:

```java
@RequiredArgsConstructor
@Getter
public class SimpleNestedAnnotation {

	@TA private final int someVal;

	@Retention(RetentionPolicy.RUNTIME)
	public @interface TA {}
}
```

Lombok fails to find the annotation and so doesn't copy it to the constructor. There is a test project that reproduces this error too: https://github.com/facboy/lombok-bug

### DIFF
--- a/buildScripts/vm-finder.ant.xml
+++ b/buildScripts/vm-finder.ant.xml
@@ -143,7 +143,7 @@ and rerun the build; this build is capable of finding VMs automatically on many 
 			<matches pattern="^\s*$" string="${jvm.loc}" />
 		</condition>
 		<fail if="jvm.loc.aborted">aborted</fail>
-		<fail if="jvm.loc.invalid">.
+		<fail>.
 
 ERROR: That does not appear to be a valid location; ${jvm.loc}/bin/${exe.java} should exist.
 			<condition>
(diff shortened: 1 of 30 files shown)


rzwitserloot:
Hi @facboy - your commit's message is a bit terse and there's no other context. Which can be allright, we can always review the patch. However, you ran your code autoformatter so that's going to take hours. Without some indication as to what you're trying to address with this we're not going to review this; we'd have to close it, unreviewed. Which is sad.

Can you update this PR, or, probably better, open a new one, with a bit more context as to what it is trying to accomplish, and ensure that you _do not_ run autoformatters, because that makes timely review impossible.

The build file update looks relevant but is clearly not related to 'Fix type detection of nested annotations' - a separate PR would be best, but if you must include it here - at least make that a separate commit.

without further chat: Close this unmerged 2024-10-20.

facboy:
geez, i didn't even notice the formatting change, will fix that.

also will address the other comments.

facboy:
Addressed comments (hopefully), and added some more detail to the description.

PR #3624, opened 2024-03-12
Add support for Eclipse 2024-03 (4.31)

### LINKED ISSUE #3620: [FEATURE] Add support for Eclipse 2024-03 (4.31)
**Describe the feature**
While working on a fix for broken Javadoc in Eclipse, I realised that I needed to add new patches for the upcoming Eclipse release. There also seem to be some changes that might affect the handling of string literals and delegate methods. I created this issue to track the fixed stuff and allow everyone to report additional problems in the same place.

Everyone who wants to use Eclipse 2024-03 (release date is 13.03.2024).

**Latest pipeline build**
[lombok.jar](https://github.com/projectlombok/lombok/actions/runs/8256670904/artifacts/1320428636)

Rawi01:
This PR fixes #3620

I couldn't find a better solution to fix the delegate stuff in Eclipse. The changed method signature is really hard to handle. I hope the reflection stuff doesn't affect performance too much...

### DIFF
--- a/.github/workflows/ant.yml
+++ b/.github/workflows/ant.yml
@@ -106,11 +106,13 @@ jobs:
         - eclipse-202212
         - eclipse-202309
         - eclipse-202312
+        - eclipse-202403
         - eclipse-I-build
         - eclipse-oxygen-full
         - eclipse-2022-03-full
         - eclipse-2023-09-full
         - eclipse-2023-12-full
+        - eclipse-2024-03-full
         - eclipse-I-build-full
         - ecj11
         - ecj14
--- a/buildScripts/ivy.xml
+++ b/buildScripts/ivy.xml
@@ -33,6 +33,7 @@
 		<conf name="eclipse-202212" />
 		<conf name="eclipse-202309" />
 		<conf name="eclipse-202312" />
+		<conf name="eclipse-202403" />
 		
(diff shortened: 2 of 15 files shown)


rspilker:
Wow, that was quite some hard work!

It looks good to me, but I wasn't able to test or validate all scenario's. But given that it's currently broken, and we do have tests, also for older versions, I'm willing to send it out there.

@rzwitserloot and @Rawi01, do you agree with my reasoning?

rspilker:
I did rerun the failing checks, regarding Eclipse-I (I do not know what version that is) and the problem seems to be:

```
[junit] Lombok annotation handler class lombok.eclipse.handlers.HandleBuilder failed
    [junit] java.lang.NoSuchMethodError: 'void org.eclipse.jdt.internal.compiler.ast.CaseStatement.<init>(org.eclipse.jdt.internal.compiler.ast.Expression, int, int)'
    [junit] 	at lombok.eclipse.Eclipse.createCaseStatement(Eclipse.java:285)
```

rspilker:
https://github.com/projectlombok/lombok/actions/runs/8116751220 was the last successful run of the Eclipse-I tests, but the first run that broke it seems to have nothing to do with the suddenly broken tests.

PR #3626, opened 2024-03-14
Support unnamed variables

Rawi01:
This PR adds support for JEP 456 (Unnamed Variables & Patterns) in delombok. JEP 456 was integrated in JDK22.


rspilker:
I expect that after the rebase the jdk22 test will suddenly work, so I'll just merge this and deal with the consequences.

PR #3643, opened 2024-03-29
[docs] Add a note Builder.Default limitation with records

kraiss:
As mentioned by @mjustin in https://github.com/projectlombok/lombok/issues/3547#issuecomment-1843451391, `Builder.Default` is not supported with records.

This PR intends to document this limitation on the website and a possible workaround. 

I did not find documentation to run the website locally so there may be a styling/alignment issue. How can I check it?


rspilker:
`ant website.open` should do the job.

kraiss:
Thanks, I'm not familiar with `ant`. Rendering is okay if you're okay with adding this to the documentation

PR #3646, opened 2024-04-01
feat: allow generics with SuperBuilder customization

ntroutman:
Trying to customize @SuperBuilder for classes that contain generics will fail due to a check in `javac/handlers/HandleSuperBuilder.constructorExists` assuming that the only generics on the builder are the ones introduced by the builder itself. This means attempting to do:

```
@SuperBuilder
public class Parent<T> {
  int field;

  protected Parent(ParentBuilder<T, ?, ?> b) {
    if (b.field == 0)
      throw new IllegalArgumentException("field must be != 0");
      this.field = b.field;
    }
}
```

Lombok fails with the error `error: constructor Parent(ParentBuilder<T,?,?>) is already defined in class Parent`


janrieke:
Let me know if you need any support, e.g. a intermediate review etc.

src/core/lombok/javac/handlers/HandleSuperBuilder.java
```diff
@@ -1160,7 +1160,7 @@ private boolean constructorExists(JavacNode type, String builderClassName) {
 						if (lastIndexOfDot >= 0) {
 							typeName = typeName.substring(lastIndexOfDot+1);
 						}
-						if ((builderClassName+"<?, ?>").equals(typeName))
+						if (typeName.startsWith(builderClassName) && typeName.endsWith("?, ?>"))
```
janrieke, line 1163: The code before was not optimal already, as String comparisons tend to break quickly. Imagine something like Java extending the diamond operator also for ?, so that you can write:
`protected Parent(ParentBuilder<> b) {...}`
Thus, we should use the opportunity to improve it. I suggest doing an `instanceof` check and only use the generic `toString` as a fallback:
```
String typeName;
JCTree paramType = md.params.get(0).getType();
if (paramType instanceof JCTypeApply) {
	typeName = ((JCTypeApply)paramType).clazz.toString();
} else {
	typeName = paramType.toString();
}
int lastIndexOfDot = typeName.lastIndexOf('.');
if (lastIndexOfDot >= 0) {
	typeName = typeName.substring(lastIndexOfDot+1);
}
if (builderClassName.equals(typeName))
	return true;
```

janrieke (changes requested):

src/core/lombok/javac/handlers/HandleSuperBuilder.java
ntroutman: Thanks, I'll work on that. I won't like have a chance until a couple weeks from now.

rzwitserloot: Yeah, using `toString()` to compare generics is a nono, though I find the idea of falling back on that in case a series of `instanceof` fails a great idea. @ntroutman still working on it, or?

rsebestik:
Hello, this feature is somehow still not fixed, are there any plans to finish this PR?

janrieke:
Are you still on this, @ntroutman? If not, I'd take over.

ntroutman:
Updated with the suggested approach of using JCTypeApply to check the type of the "builder" parameter.

janrieke:
Please add yourself to the AUTHORS file.

rsebestik:
@janrieke hello, can it be approved and possibly merged? thank you.

janrieke:
I don't have the permissions, that's a job for the maintainers.

Rawi01 (commented):
- The two failing checks can be ignored, thats a general build problem that we have to fix and is unrelated to your changes. 
- If I haven't missed anything there now is no test for the simple customization case anymore. I know that your changes doesn't break it but to be safe during future refactorings I would love to have this code path covered.

rsebestik:
@janrieke are you willing to finalize the changes as requested please?

janrieke:
@ntroutman: Can you rebase this PR with the current master and re-add the simple customization test case?

PR #3649, opened 2024-04-04
Update Javadoc for @Data

sergvasylchak:
The javadoc on Data.java says it generates a constructor, which is the equivalent of `@RequiredArgsConstructor` behavior. In reality though, the constructor is only generated if no explicitly defined constructors (including explicit Lombok constructors) already exist. The behavior is expected and documented [here](https://projectlombok.org/features/Data).

The javadoc alone, however, often causes confusions, since there is no clear explanation on why the constructor in `@Data` may or may not be generated, which leads to a need of additional digging in Lombok documentation.

### DIFF
--- a/src/core/lombok/Data.java
+++ b/src/core/lombok/Data.java
@@ -28,7 +28,8 @@ import java.lang.annotation.Target;
 
 /**
  * Generates getters for all fields, a useful toString method, and hashCode and equals implementations that check
- * all non-transient fields. Will also generate setters for all non-final fields, as well as a constructor.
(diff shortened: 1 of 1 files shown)


rzwitserloot:
The javadoc is not intended to reproduce every little niggle; the features page on the site is the source of every caveat. However, given that the javadoc was basically lying (saying that it was equivalent to a stack of annotations, _including_ `@RequiredArgsConstructor`, which is only true if there is no explicit constructor), this update is on net worth adding. Accepted.

PR #3669, opened 2024-05-20
Add javadoc for generated constructors

### LINKED ISSUE #933: Javadoc for @XArgsConstructor
Two things I note w.r.t. javadoc using `@RequiredArgsConstructor`:
1. Generated javadoc shows the default constructor only, no the one generated by lombok
2. If there were a generated lombok constructor javadoc, I'd expect to see the field javadoc copied into the `@param` javadoc entries

Rawi01:
This PR fixes #933

Lombok now adds a javadoc comment to a generated constructor if there is any field with a description.

### DIFF
--- a/src/core/lombok/core/handlers/HandlerUtil.java
+++ b/src/core/lombok/core/handlers/HandlerUtil.java
@@ -974,7 +974,7 @@ public class HandlerUtil {
 
 	public static String getParamJavadoc(String methodComment, String param) {
 		if (methodComment == null || methodComment.isEmpty()) return methodComment;
-		Pattern pattern = Pattern.compile("@param " + param + " (\\S|\\s)+?(?=^ ?@)", Pattern.MULTILINE | Pattern.CASE_INSENSITIVE);
+		Pattern pattern = Pattern.compile("@param " + param + " (\\S|\\s)+?(?=^ ?@|\\z)", Pattern.MULTILINE | Pattern.CASE_INSENSITIVE);
 		Matcher matcher = pattern.matcher(methodComment);
 		if (matcher.find()) {
 			return matcher.group();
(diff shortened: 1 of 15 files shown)


archer-321:
Thanks for working on this feature!

Would it make sense to allow a `-- CONSTRUCTOR --` section similar to `-- GETTER --`? This wouldn't work for classes with multiple constructor annotations unless the syntax is changed to `-- ALL CONSTRUCTOR --`, `-- REQUIRED CONSTRUCTOR --` or similar, but it would allow for more flexible constructor documentation.

Looking at the example generated documentation in the tests, I immediately noticed the capitalization for the `@param` parameter documentation. Most projects I'm aware of (including [the OpenJDK source code](https://github.com/openjdk/jdk/blob/2c1da6c6fa2e50856ea71c0e266961171bee1037/src/java.base/share/classes/java/util/HashMap.java#L440-L443)) use lowercase descriptions while starting class item documentation with a cardinal letter. More control over the generated constructor documentation would help keep the module documentation uniform.

Rawi01:
@archer-321 Thanks for the feedback. The intended solution was to copy `@param` tags to the constructor if they are present. I missed that it doesn't work at all because of different bug that I fixed years ago but was never merged. Cherry picked that commit and it should now work as planned.

rzwitserloot:
Speaking of javadoc, @Rawi01 there's the markdown [JEP 467](https://openjdk.org/jeps/467) - the only effect I think that has on lombok is that we should copy the 'this is markdown style' flag to the javadoc node we generate, other than that, I don't think this affects either scanning for `--- GETTER ---` or scanning for `@param` / `@return`, which is unchanged in the markdown flavour. We can only move on this once all relevant tools (javac and ecj I guess) are updated to be capable of handling JEP 467 which probably isn't yet today? I haven't checked.

This does not affect accepting this PR at all, just tangentially related.

PR #3673, opened 2024-05-23
Mark injected fields as generated

Rawi01:
This PR enables marking for fields generated by `@Log`, `@Synchronized` and `@Locked`. It was disabled years ago to fix a bug but it was never enabled again.

### DIFF
--- a/src/core/lombok/eclipse/handlers/HandleLockedUtil.java
+++ b/src/core/lombok/eclipse/handlers/HandleLockedUtil.java
@@ -1,3 +1,24 @@
+/*
+ * Copyright (C) 2024 The Project Lombok Authors.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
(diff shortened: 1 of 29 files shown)


rzwitserloot:
This is actually a fairly high impact PR. Before this PR, sticking e.g. `@Log4j` on a class where you _never_ write to the log causes your editor/javac to emit a warning: You declared a private field that you never use. In e.g. eclipse his results in a yellow wavy underline under `@Log4j` that it's unused.

With this update, due to this automatically shoving `@SuppressWarnings` on the field, this behaviour is modified; you no longer get that warning.

@rspilker and I are on the fence. In some sense, the warning go: The point of the warning is twofold:

* It's "faster" if you don't have the field at all, and it's pointless; it can go away.
* You put in the effort to declare a field which you never use. Surely what you intended for your code to do, and what it actually does, are not in sync: Your code has a bug or oversight but it's not clear what; the compiler warning serves to remind you your code doesn't make sense.

Of those 2 points, the first point is, as far as we're considered, nearly irrelevant. It's all about that second point. And it's mostly irrelevant specifically for `@Log4j` and co. Ordinarily, declaring a field just to never use it is indicative of an oversight, but declaring "I might wanna log some stuff" and then never logging strikes us as highly likely to be immaterial.

So, that means.. this actually isn't all that high an impact. Or it is, but almost entirely a _good_ thing: You can now e.g. tell your IDE to just stick `@Log4j` on all new source files by default without getting annoyed at the 'you aren't using the logger!' warning you used to get. But, is this conclusion ("it does not matter / this is basically upside, yay") correct? Are we missing a situation where this update that removes the warning is an unwanted update by some lombok users?

Technically this PR is great, just, waiting for feedback on whether this is too high impact to just integrate without spending some time on e.g. docs, or, gasp, considering a mid-point version update.

rzwitserloot:
Uh, nevermind - the PR just changes how eclipse works. Our output for javac has added the suppress warnings for a decade or so already. Let's just go with suppresswarnings all everywhere then.

PR #3678, opened 2024-05-29
Standardize @Retention usage in Lombok annotationsfix: Standardize @Retention usage in annotations

xpmxf4:
### Description
This pull request standardizes the usage of the @Retention annotation across Lombok's annotations. Specifically, it changes the usage in the @Builder annotation from `@Retention(SOURCE)` to `@Retention(RetentionPolicy.SOURCE)`. 

### Reason for Change
The other annotations in the Lombok project, such as @Getter and @Setter, use `@Retention(RetentionPolicy.SOURCE)`. For the sake of consistency and readability, this change ensures that all annotations follow the same convention.

### Changes Made
- Updated `@Retention(SOURCE)` to `@Retention(RetentionPolicy.SOURCE)` in the @Builder annotation.

### Impact
This change does not affect the functionality of the annotations but improves the consistency and maintainability of the codebase.

### Verification
No additional tests are needed as this change only affects the annotation declaration style.

Please review and let me know if there are any concerns or further changes needed.

Thank you!

### DIFF
--- a/src/core/lombok/Builder.java
+++ b/src/core/lombok/Builder.java
@@ -111,7 +111,7 @@ import java.lang.annotation.Target;
  * @see Singular
  */
 @Target({TYPE, METHOD, CONSTRUCTOR})
-@Retention(SOURCE)
+@Retention(RetentionPolicy.SOURCE)
 public @interface Builder {
 	/**
 	 * The field annotated with {@code @Default} must have an initializing expression; that expression is taken as the default to be used if not explicitly set during building.


rzwitserloot:
There's a `@Target` annotation immediately above that also isn't consistent.

This isn't a meaningful update unless:

* You go the other way - Builder is right ,Setter is wrong: You want the static imports and the `@RetentionPolicy(SOURCE)`, you've moved us to a more consistent, but less desired place. The PR should update Getter, Setter etc instead.
* It needs to cover _all_ the annotations. So, also `@Target`, which also needs to move to a 'use star-static imports' mode.

NB: Stylistically it's star imports, because annotations predetermine what the values can be; The `RetentionPolicy.` adds nothing useful.

PR #3683, opened 2024-06-07
[new-annotation] @Adapter annotation

csorbazoli:
Test cases included for both the javac and eclipse test suites
There is a lombok.javac.handlers.HandleAdapter and lombok.eclipse.agent.PatchAdapter implementation.
Apart from the unit tests I've tried it with eclipse (STS 4.17.1) and wrote unit tests for my classes using the new `@Adapter` annotation to test the generated methods as well.
It all looks good and quite useful.

Annotation suggestion - @Adapter

There are some cases, when you need to write a class that implements a certain interface, but you don't need all of the interface methods, then either you leave the class abstract, or use some Proxy wrapper or let the IDE create all the methods with an empty body. The latter one is not so nice, as there are many empty method blocks, or TODO comments, that does not look good or needs some manual effort to clean them up. Also these methods are not doing anything, so they are not worth the time to cover them with tests. That leads to "boilerplate" code, that obscures the code.

This often happens, when you need to write some test doubles (mocks) or using the nullables methodology for creating simple implementations that can be injected instead of the "real" dependencies.
In the first case, usually a mocking framework is the right solution, but even if that's available in some cases it's better to implement a simple and straightforward mock class, instead of setting up the expected answers using the mock framework itself.
In case of the nullables, the mocking framework is not a good option, as the nullable implementation is part of the production code (like a dry-run option), so the usage of test framework dependencies are discouraged.
In both cases, the main focus is on those specific methods that are significant for the business logic at hand, and all the other interface methods could be disregarded.
For example, the service is injecting a DAO component, that has many general methods provided by the framework (e.g. Spring), but the service only needs the findByNameAndExpirationDate method and nothing else. Then a the "nullable" object should implement only this method and some extra initialization/setup logic, but nothing else from the base DAO interface. 

My suggestion is a lombok annotation with the following features.
Name: @Adapter
Attributes:
* of = interface(s)
  Optional attribute to control which interface's methods should be generated.
  As a default behaviour, all the abstract methods of given class should be auto-generated by lombok.
* throwException = Exception class
  Optional attribute. If not specified, then the default implementation does nothing (void method) or returns a default value according to the method declaration (0, false, null, empty collection, etc).
* message = Exception message
  Optional attribute as an additional setting for the throws option.
* suppressThrows = true/false (default: false)
  The generated methods are replicating the same throws declaration as the interface declares them. In case of 'suppressThrows = true' the generated method overrides are omitting these throws declarations, considering that these methods would never throw the declared errors.
* silent = true/false
  If true, then the generated methods return default value (e.g. null) instead of throwing UnsupportedOperationException
  
"Adapter" is a common phrase for classes that provides default implementations for some interfaces and can be used for implementing simple custom adapters by extending them. Maybe the most known example is the EventListener and EventAdapter, where you only need to extend those onX methods that are relevant for given case.

Additional notes. The extended interfaces can have super interfaces. So, all the unimplemented methods should be autogenerated.
But the interface may declare default method bodies, therefore these methods does not require generating an override.
Basically, the @Adapter annotation is aiming fix the "unimplemented methods" compilation error, by providing the default method overrides. This way only those methods needs to be implemented that requires special behaviour.

Code samples.

Let's assume we have this interface declared.

```
public interface FooInterface {
   void doStuff();
   
   int countMyObjects(User user) throws UserNotFoundException;
  
   String getStatus() throws IllegalStatusException;
}
```

Default adapter settings:

```
@Adapter
public class NullableFooInterface implements FooInterface {
  @Override
  public String getStatus() {
    return "TestStatus";
  }
}
```

That results the following code:

```
public class NullableFooInterface implements FooInterface {
  @Override
  public String getStatus() {
    return "TestStatus";
  }
  @Override
  public void doStuff() {
    throw new UnsupportedOperationException();
  }
  @Override
  public int countMyObjects(User user) throws UserNotFoundException {
    throw new UnsupportedOperationException();
  }
}
```

Silent mode:

```
@Adapter(silent = true)
public class NullableFooInterface implements FooInterface {
}
```

That results the following code:

```
public class NullableFooInterface implements FooInterface {
  @Override
  public String getStatus() {
    return null;
  }
  @Override
  public void doStuff() {
    // empty method
  }
  @Override
  public int countMyObjects(User user) throws UserNotFoundException {
	return 0;
  }
}
```

Custom adapter settings:

```
@Adapter(of = FooInterface.class, throws = UnsupportedOperationException.class, message = "NOT_IMPLEMENTED", suppressThrows = true)
public class NullableFooInterface implements FooInterface {
  @Override
  public String getStatus() {
    return "TestStatus";
  }
}
```

That results the following code:

```
public class NullableFooInterface imlements FooInterface {
  @Override
  public String getStatus() {
    return "TestStatus";
  }
  @Override
  public void doStuff() {
	throw new UnsupportedOperationException("NOT_IMPLEMENTED");
  }
  @Override
  public int countMyObjects(User user) {
	throw new UnsupportedOperationException("NOT_IMPLEMENTED");
  }
}
```

See more use cases in the unit tests.


rzwitserloot:
It's a shame you put in all this effort before contacting us. Because, unfortunately, PR denied.

The problem is twofold:

## It's.. kinda out there

In the distant past this PR would have made sense. But now that interfaces have `default`, it's... still relevant, but less so.

It's opinionated: Given that it surely as a feature isn't there to do what `default` can already do, it's more about mocking implementations, for example. And mocking is.. contentious. And we fall on the 'properly written code should not have this'.

"We do not like it" is not a showstopper. But it does make it harder to accept a PR.

## Lombok does not need to fix this

Which means this is a death knell for the PR: Lombok isn't what you need.

`java.lang.reflect.Proxy` can trivially do this:

```java
java.lang.reflect.Proxy.newProxyInstance(someClassLoader, FooInterface.class,(proxy, method, args) -> {
  throw new UnsupportedOperationException();
});
```

is all you need. You can festoon it with the customization options fairly easily, including returning `0 / '\0' / false / null`.

Alternatively, the adapter class this generates is a completely separate class file. Which means a plain jane annotation processor can do this job. Granted, lombok does it better, but all lombok features are adjudicated on a cost v. benefit analysis, and the benefit side of the scale is significantly reduced because it's just the 'lombok does it better' part, not the 'lombok does it at all' that features like e.g. `@Setter` get.

csorbazoli:
Thank you for the positive feedback and the detailed response.
The effort I've put in is not wasted. Learned a lot and if nothing else, then it was a great experiment.
You're right, that there are other solutions even if Lombok could do it better. Just need to find the one that suits me best.
I know about the Proxy, and I am using it some cases, but I don't really like the resulting Proxy layer.
I'm going to experiment further to get around those things that bother me.

Issue #3713, opened 2024-07-29
[FEATURE] Builder @Ignore 


CrypticCabub:
It's getting a little rediculous the amount of traction this old issue is getting without even being noticed by the maintainers of this project. On the off-chance that the old closed FR is simply being filtered out before anybody actually sees it I am opening a new issue with a reference back to the original one that has 20+ people asking for it:

https://github.com/projectlombok/lombok/issues/2307

janrieke:
While @rspilker is correct that there is a simple workaround for `@Builder` (by putting `@Builder` onto a custom constructor), this is not the case for `@SuperBuilder`. `@SuperBuilder` is only valid as annotation on a class, so you cannot create your own constructor. There is also no other way to exclude a field for `@SuperBuilder`.

Thus I propose adding `@SuperBuilder.Exclude`, which only affects `@SuperBuilder`. Anyone who needs this feature for a `@Builder` and does not like the workaround could switch over to `@SuperBuilder`.

Pros:
- `@SuperBuilder` is still experimental, so even if this feature would require changing some aspect of the generated code (which i doubt), it won't be much of a problem.
- No change to the widely used `@Builder`.

Cons:
- Switching from `@Builder` to `@SuperBuilder` causes additional type parameters on builder variables. However, you can use `var` from Java 10 on if that annoys you.

To discuss:
- How do we deal with field initialization when there is not initializing expression? `null` for Objects, default value for primitives? Or error out when there is no initializer? (For the sake of simplicity, I vote for the latter.)

CrypticCabub:
That could potentially work, and I admittedly have never used @superbuilder to know what the difference between how it and @builder interact with parent/child relationships. 

Personally I don't currently see the issue with adding an @Builder.exclude() option to the base builder class that would have no impact on generated code when not used, and putting @Builder on a custom constructor is not as easy of a workaround as it sounds because the most common use-case for builder.exclude that I am familiar with is on large JPA entities where the custom constructor workaround results in a very ugly and hard-to-validate constructor since for say a 20-property JPA object with a JPA-managed variable that needs excluding, you now have a 19-arg constructor that you have to visually identify which property is missing. In comparison a simple exclude annotation and a comment on why that property is excluded would very concisely convey that the property is managed by the framework and should not be interacted with otherwise

Issue #3283, opened 2022-10-19
[FEATURE] provide @Builder.Exclude


KrawN:
Hi,
there are so many requests on this feature everywhere that I would just paste a link to the prematurely closed previous request:
[https://github.com/projectlombok/lombok/issues/2307](https://github.com/projectlombok/lombok/issues/2307)
Thank you for hearing the users

Issue #2307, opened 2019-12-09
[FEATURE] provide @Builder.Exclude


nimo23:
**Describe the feature**
We can use the `@Builder` on class level. However, with this, we cannot omit some fields from the builder. Please provide the field annotation `@Builder.Exclude` to exclude fields from the builder pattern.

```
@Builder
public class User {

    // should not be included in the builder pattern
    @Builder.Exclude
    private int id;

    private String name;

}
```

rspilker:
I do miss in what common scenario you would want to exclude fields from the builder.

The way to solve this is by putting `@Builder` on a constructor instead of a type.

Basically, you can put `@Builder` on a method, and we will generate a Builder for all parameters, where the `build()` invokes the method. In this sense, the constructor is also a method.

For convenience, we allow `@Builder` on a type. Then we generate an `@AllArgsConstructor`, and act as if that was annotated with `@Builder`.

nimo23:
>I do miss in what common scenario you would want to exclude fields from the builder.

I have this class:

```
@Builder
@NoArgsConstructor
@AllArgsConstructor
public class User {

// many properties
...

@Builder.Default
private Task task = new Task();

@Builder.Exclude // does not exist
private Type type;

}
```

For convenience, we allow @Builder on a type. Then we generate an @AllArgsConstructor, and act as if that was annotated with @Builder.

With type you mean class? Note that `@AllArgsConstructor` is only created if no `@NoArgsConstructor` is on the class. See https://github.com/rzwitserloot/lombok/issues/2305.

>@Builder on a constructor or on method

yes, but with a class having many properties we have to maintain a large constructor or large parameter list.

nimo23:
@rspilker look at https://stackoverflow.com/questions/30717640/how-to-exclude-property-from-lombok-builder then you understand the use cases why having a `@Builder.Exclude` makes sense. 

Actually, If I have many properties within a class and only want to exclude one property from the builder then I am forced to put a large constructor or static method with @Builder annoation instead of having the @Builder on class level. With `@Builder.Exclude` I do not need to add a constructor or static method manually.

@rspilker So please make `@Builder.Exclude`  available.

eliast101:
This issue needs to be reopened.

rzwitserloot:
@eliast101 No.

Kyle-Holub:
[...]

PR #3723, opened 2024-08-14
[POC]  @Builder.Exclude for a field in cases where @Builder/@SuperBuilder is on a type

a1shadows:
relates to #2307 , #3283 , #3713 

## Motivation
I know that #2307 has been closed but from the discussion thread I realized that many others were having the same issues I was: that some fields need not be part of a builder (or whatever creation pattern the class is using).

I understand that there is the option of creating a static method/constructor and annotating it with `@Builder` to achieve exclusion of field(s). However, I would argue that a similar argument could have been made for default values as well. Yet, we have a `@Builder.Default` which is a super useful feature to have!

My use cases specifically was with regards to JPA Entity classes where I wanted to hide certain non-insertable/non-updatable fields in a JPA Entity from users attempting to build an Entity object. While using `@Setter`, this was simple enough with `@Setter(AccessLevel.PRIVATE)`, but I wanted to use `@Builder` since the entities were large and complex. There I had no options with regards to excluding these fields. Not excluding them would mislead the users since such fields are not persisted in the database even when initialized so I was forced to use Setters.

I did not want to badger the maintainers with excessive argumentation so I thought a "POC" might better demonstrate what such an idea would like and would take the discussion forward more constructively. I would greatly appreciate any and all feedback.

## Implementation 
The basic idea is that if the `@Builder`/`@SuperBuilder` annotation is on a type, no setter methods will be generated in the builder for the fields marked with `@Builder.Exclude`.  This means that although the builder class will contain all the fields(so as not to break any existing logic of mapping the builder class to the annotated class through a constructor, generated for builder or maybe through `@AllAgsConstructor`). However,  it will not be possible to set the value of the excluded field before calling the build() method.
It is somewhat similar in use(not implementation) to `@Builder.Default`:
```
@Builder
public class User {

    // field setter method will not be generated in builder
    @Builder.Exclude
    private int id;

    private String name;

}
```
The builder for this class will not contain a `User.UserBuilder id(int id)` method.

 I've added a warning similar to the one for `@Builder.Default` for when the annotation is not on type.  As is the case with `@Builder.Default`, the `@Builder.Exclude` annotation holds no meaning when used without a `@Builder` on the type. if the `@Builder` is on a constructor or a static method, the exclusion will not take place

I've added a few test cases to demonstrate its working. 

I'll add more test cases to the PR if the direction looks promising!

### DIFF
--- a/src/core/lombok/Builder.java
+++ b/src/core/lombok/Builder.java
@@ -120,6 +120,13 @@ public @interface Builder {
 	@Retention(SOURCE)
 	public @interface Default {}
 
+	/**
+	 * The field annotated with {@code @Exclude} will be excluded from the builder;
+	 */
+	@Target(FIELD)
+	@Retention(SOURCE)
+	public @interface Exclude {}
+
 	/** @return Name of the method that creates a new builder instance. Default: {@code builder}. If the empty string, suppress generating the {@code builder} method. */
 	String builderMethodName() default "builder";
 	
--- a/src/core/lombok/Builder.java
+++ b/src/core/lombok/Builder.java
@@ -121,7 +121,7 @@ public @interface Builder {
 	public @interface Default {}
 
 	/**
-	 * The field annotated with {@code @Exclude} will be excluded from the builder;
+	 * The field annotated with {@code @Exclude} will be excluded from the builder.
 	 */
 	@Target(FIELD)
 	@Retention(SOURCE)
--- a/src/core/lombok/Builder.java
+++ b/src/core/lombok/Builder.java
@@ -121,11 +121,12 @@ public @interface Builder {
 	public @interface Default {}
 
 	/**
-	 * The field annotated with {@code @Exclude} will be excluded from the builder.
+	 * The field annotated with {@code @Exclude} will be excluded from the builder. Excluded objects will be initialized as null
+	 * and primitives with their default value.
 	 */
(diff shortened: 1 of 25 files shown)


a1shadows:
@rspilker @rzwitserloot  could you please take a look at this whenever possible?

src/core/lombok/Builder.java
```diff
@@ -120,6 +120,13 @@
 	@Retention(SOURCE)
 	public @interface Default {}
 
+	/**
+	 * The field annotated with {@code @Exclude} will be excluded from the builder.
```
Ordiel, line 124: Maybe we should consider adding a legend to indicate that Objects will be set to null and primitives to its defaul value

a1shadows: yes, i can do that. on it

a1shadows: done

isasmendiagus (approved):

agustingroh (approved):

isasmendiagus:
@rspilker @rzwitserloot Hi, we need this feature as well. Could you please take a look? 
Thanks

a1shadows:
Damn arrogance has a limit. I'll keep my feature then.

alionell:
Why was this PR not merged? It is as if the maintainers do not want to add this super useful feature just out of spite.

PR #3740, opened 2024-09-10
Exclude the classloader for jasperreports-maven-plugin due to a classloader conflict with ECJ

Postremus:
I am currently in the process of renaming the plugin from jasperreports-plugin to jasperreports-maven-plugin.


rzwitserloot:
Hi @Postremus - thanks! Nice of you to think of other libraries that have your GAV hardcoded :)

Much appreciated.

PR #3744, opened 2024-09-11
[fixes #3116] Add multi round support for mapstruct

### LINKED ISSUE #3116: [BUG] Mapstruct mappers are wrong if lombok needs to create Builders in later rounds
**Describe the bug**
the lombok implementation of `AstModifyingAnnotationProcessor` in https://github.com/projectlombok/lombok/blob/560ded50127b12bd8168f8ae2bada47253f15bc8/src/bindings/mapstruct/lombok/mapstruct/NotifierHider.java#L19-L33 does not work correctly when lombok gets called in a subsequent round to modify a file emitted by a different AP in an earlier round (that contains `@SuperBuilder` or `@Builder`). 

Instead of reporting to mapstruct if a given type is "complete" lombok reports "complete" for all files as long as it has run once.

**To Reproduce**
I've uploaded a reproducer project at https://github.com/fraenkelc/bug-lombok-mapstruct-ap

Steps:
1. create an AP that writes a source file which contains the `@SuperBuilder` (probably also affects `@Builder`) lombok annotation
2. Use this AP together with Lombok, mapstruct and the lombok-mapstruct binding in a project
3. Write a Mapper that maps **to** the generated file
4. compile
5. Mapstruct will run too early and not see the modified ast and in turn not use the lomboks builder
  https://github.com/fraenkelc/bug-lombok-mapstruct-ap/blob/aa68ee421703115af6cd7c19d1045d3e110d506c/README.md?plain=1#L21-L43

**Expected behavior**
Lombok should correctly delay Mapstruct until it is done, e.g. by implementing AstModifyingAnnotationProcessor with a per-type logic instead of the current "ran once and done" logic.

 - 1.18.22
 - Java 11.0.14.1 (Eclipse Adoptium)

kanchev1:
Report to MapStruct that a type is completed when there aren't any Lombok annotations left on it. Lombok annotations are removed whenever a class is processed. This way, annotations which require multiple rounds to process are also correctly handled, and MapStruct processing will be delayed until Lombok completely finishes processing required types.

Fixes issues with e.g. @FieldNameConstants and @Builder (see #3116)

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -1,6 +1,7 @@
 Lombok contributors in alphabetical order:
 
 Adam Juraszek <juriad@gmail.com>
+Aleksandar Kanchev <136312841+kanchev1@users.noreply.github.com>
 Aleksandr Zhelezniak <lekan1992@gmail.com>
 Amine Touzani <ttzn.dev@gmail.com>
 Andre Brait <andrebrait@gmail.com>
--- a/src/bindings/mapstruct/lombok/mapstruct/NotifierHider.java
+++ b/src/bindings/mapstruct/lombok/mapstruct/NotifierHider.java
@@ -1,35 +1,37 @@
 package lombok.mapstruct;
 
-import java.lang.reflect.Field;
-
-import javax.lang.model.type.TypeMirror;
-
 import org.mapstruct.ap.spi.AstModifyingAnnotationProcessor;
 
+import javax.lang.model.element.AnnotationMirror;
+import javax.lang.model.type.TypeMirror;
+import java.util.List;
+
+/**
+ * Report to MapStruct that a type is completed when there aren't any Lombok annotations left on it. Lombok annotations
+ * are removed whenever a class is processed. This way, annotations which require multiple rounds to process are also
+ * correctly handled, and MapStruct processing will be delayed until Lombok completely finishes processing required types.
+ */
 class NotifierHider {
 	
 	public static class AstModificationNotifier implements AstModifyingAnnotationProcessor {
-		private static Field lombokInvoked;
-		
-		@Override public boolean isTypeComplete(TypeMirror type) {
-			if (System.getProperty("lombok.disable") != null) return true;
-			return isLombokInvoked();
-		}
-		
-		private static boolean isLombokInvoked() {
-			if (lombokInvoked != null) {
-				try {
-					return lombokInvoked.getBoolean(null);
-				} catch (Exception e) {}
+
+		@Override
+		public boolean isTypeComplete(final TypeMirror typeMirror) {
+			final List<? extends AnnotationMirror> annotationMirrors = typeMirror.getAnnotationMirrors();
+			if (annotationMirrors == null || annotationMirrors.isEmpty()) {
 				return true;
 			}
-			
-			try {
-				Class<?> data = Class.forName("lombok.launch.AnnotationProcessorHider$AstModificationNotifierData");
-				lombokInvoked = data.getField("lombokInvoked");
-				return lombokInvoked.getBoolean(null);
-			} catch (Exception e) {}
+
+			for (final AnnotationMirror annotationMirror : annotationMirrors) {
+				final String annotationName = String.valueOf(annotationMirror);
+				// check for ClaimingProcessor's SupportedAnnotationTypes
+				if (annotationName.startsWith("@lombok.")) {
+					return false;
+				}
+			}
+
 			return true;
+
 		}
 	}
 }
--- a/src/launch/lombok/launch/AnnotationProcessor.java
+++ b/src/launch/lombok/launch/AnnotationProcessor.java
@@ -39,10 +39,6 @@ import sun.misc.Unsafe;
 
 class AnnotationProcessorHider {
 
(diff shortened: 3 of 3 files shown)


rzwitserloot:
Thanks, @kanchev1 ! I just restored the 'hey if lombok is entirely disabled, just tell mapstruct we are definitely not going to modify that AST anymore' and applied the house style.  Good work.

rgrunber:
Hi all, I think this commit introduced a regression in Lombok.

I have the following steps to reproduce :

1. `git clone https://github.com/dxaraujo/vscode-java/` (this is a sample project that shows the issue)
2. Enter the project and set the `lombok.version` property to 1.18.35 in the pom.xml.
3. Now go into https://github.com/projectlombok/lombok and build against the commit before this one, 268b6145 (`ant dist`).
4. Run `mvn install:install-file -Dfile=dist/lombok-1.18.35.jar -DgroupId=org.projectlombok -DartifactId=lombok -Dversion=1.18.35 -Dpackaging=jar`, which should install the buld artifact into the local Maven repository.
5. Go into the "vscode-java" sample project above and run `mvn clean verify`. It will use the Lombok we just built, and succeed.
6. Go back to the lombok project and checkout and build this commit, 04c9755b (`ant dist`).
7. Run `mvn install:install-file -Dfile=dist/lombok-1.18.31.jar -DgroupId=org.projectlombok -DartifactId=lombok -Dversion=1.18.35 -Dpackaging=jar` (Note the jar is lombok-1.18.31.jar but we'll treat it as 1.18.35 to simplify things)
8. Perform (5) again, and you'll see the build fails with :

```
[ERROR] COMPILATION ERROR : 
[INFO] -------------------------------------------------------------
[ERROR] /home/rgrunber/sample-projects/issue-3836/src/main/java/com/vscode/java/mapper/UserMapper.java:[9,8] No implementation was created for UserMapper due to having a problem in the erroneous element java.util.ArrayList. Hint: this often means that some other annotation processor was supposed to process the erroneous element. You can also enable MapStruct verbose mode by setting -Amapstruct.verbose=true as a compilation argument.
[INFO] 1 error
[INFO] -------------------------------------------------------------
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
```

testforstephen:
@rgrunber Could you open an issue for this in lombok repository?

rzwitserloot:
This issue will now be reverted, and, we think, this approach of fixing it is a dead end. As far as we can tell:

1. The 'old' version has lombok itself track whether it has run or not, and uses that to answer mapstruct when it asks lombok: "Hey, is this type complete?". There are 2 problems with this; [A] lombok has its own rounds system and perhaps its an issue that we start answering 'yeah we are done' to the question right away, perhaps we should wait until our rounds are done (but this is only relevant if mapstruct asks us in the middle of lombok's processing, and I don't think it does, so this is probably irrelevant), and [B] if the annotation processing part of javac itself causes multiple rounds, which usually happens when APs generate new source files, then the newly generated files are immediately considered as 'yeah it is done' by lombok because lombok has already ran a full cycle. Even if that particular is completely new, and contains unprocessed lombok stuff.
2. The 'new' version ignores any bookkeeping that lombok does and instead simply checks if there is lombok stuff in the type that is being asked about. __But this is impossible__.

The impl simply asks the `TypeMirror` representing type X in mapstruct's question to lombok: "Hey, is type X complete; i.e. are you done adding methods and fields and such to it?" - do you have any annotations on you (on __the type__) that are lombok annotations?

This isn't complete! Lombok can trigger on annotations on local variables (such as `@Cleanup`), as well as `var`/`val`. Yes, _right now_ those 'shows up in method bodies' lombok annotations don't affect the completeness of the type (they don't add new methods), but those aren't handcuffs we wanna put on. We don't want to close the door on ever doing that.

That's why its a dead end. Yes, the `TypeMirror` can be recursed to scan for lombok-related annotations _anywhere_ inside it, but that's still not enough; `TypeMirror` does not allow you to drill all the way into method bodies, but it has to, to properly answer the question "is lombok completely done with this?"

Separately, lombok might want to leave some lombok artefacts behind. We already do that: `@lombok.NonNull` can remain; it doesn't _currently_ conflict with kanchev's PR because NonNull is not a thing that can exist on the type itself, but that will immediately throw the PR's approach into chaos if you add a mechanism to drill into the content of the TypeMirror.

Hence, the strategy chosen seems fundamentally unworkable.

We _think_ the following strategy _would_ work:

Instead of 'globally' tracking "has lombok run yet?", track it per source file (in AST terms, per compilation unit). If lombok has completed an entire cycle on some source file, then that type _is_ complete. It's not about whether lombok stuff remains in it (it might, stuff like `@NonNull` may remain), it's about whether lombok considers all lombok-related stuff in it as 'processed'. Assuming lombok is the only AP around that modifies types themselves, then a single cycle is sufficient to deem a compilation unit as 'as far as lombok is concerned, it is complete'.

I guess storing some sort of hashset containing _every_ type being compiled in one go is possibly a slight memory hog, but it's probably fine. In the end shoving 20k strings each ~40 chars long into a hashset isn't all that much memory these days.

PR #3749, opened 2024-09-19
Patch SourceTypeConverter to avoid parsing empty strings

### LINKED ISSUE #3748: [BUG] @Synchronized causes error markers in Java editor with Eclipse 2024-09 (4.33.0)
**Describe the bug**
Since the update to Eclipse 2024-09 (4.33.0), classes that contain one or more `@Synchronized` annotations cause "_CLASSNAME_ cannot be resolved to a type" errors in other classes that reference these classes.

Note that these errors only appear when the source file is actually opened in a Java editor. They neither show up in the Outline nor the Package Explorer. It seems they are just pseudo errors, because the compilation produces a valid class file and actually loading and executing the compiled classes also works. 

Removing the `@Synchronized` makes the error disappear. Putting some other lombok annotations like `@Builder` or `@EqualsAndHashCode` on the (referenced) class also removes the error, while other annotations like `@Getter` or `@ToString` do not affect the error.

There is no error in the Eclipse logs when this problem is present.

**To Reproduce**
Create a new Eclipse project with just these two files:

```
package test;
import lombok.Synchronized;

public class ClassWithSynchronized {
	@Synchronized
	public void test() {
	}
}
```
and
```
package test;

public class UsingClass {
	private ClassWithSynchronized test;
}
```

When opening `UsingClass` in the Java editor, you'll see an error marker at line 4 with the error message "ClassWithSynchronized cannot be resolved to a type".

**Expected behavior**
No error markers present in referencing classes.

 - Lombok 1.18.34 
 - eclipse 2024-09 (4.33.0) (uses bundled Java `org.eclipse.justj.openjdk.hotspot.jre.full.win32.x86_64_21.0.4.v20240802-1551`)

Rawi01:
This PR fixes #3748

### DIFF
--- a/src/eclipseAgent/lombok/eclipse/agent/EclipsePatcher.java
+++ b/src/eclipseAgent/lombok/eclipse/agent/EclipsePatcher.java
@@ -875,6 +875,14 @@ public class EclipsePatcher implements AgentLauncher.AgentLaunchable {
 				.target(new MethodTarget(SOURCE_TYPE_CONVERTER_SIG, "convertAnnotations", ANNOTATION_SIG + "[]", I_ANNOTATABLE_SIG))
 				.wrapMethod(new Hook("lombok.launch.PatchFixesHider$PatchFixes", "convertAnnotations", ANNOTATION_SIG + "[]", ANNOTATION_SIG + "[]", I_ANNOTATABLE_SIG))
 				.request(StackRequest.PARAM1, StackRequest.RETURN_VALUE).build());
+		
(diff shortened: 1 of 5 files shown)


rzwitserloot:
Ha! That trick of replacing `assertTrue(list.isEmpty())` with `assertEquals(Collections.emptyList(), list)` is, I presume, so that if the assertion fails, the JUnit IDE renderer thingie will show you the non-empty list, instead of simply telling you: This assertion failed?

PR #3750, opened 2024-09-20
Support new JDK23 CommentStyle

Rawi01:
This PR updates lombok to support the updated `CommentStyle` enum. I also added JDK 23 as test target.


reneleonhardt:
Amazing work, thank you! 🚀

reneleonhardt:
You don't test against JDK 24, why is a test against an Eclipse integration build not allowed to fail?
For me testing a released version like 2024-09 (two weeks old) would be more important.
https://projects.eclipse.org/releases/2024-09

@Rawi01 in case you didn't notice, there is no explaining comment but I guess docker should be tested against LTS versions and the latest JDK, you didn't replace 22 with 23 there 😉
https://github.com/projectlombok/lombok/pull/3750/files#diff-861032010d7892d288389586a0f5b825c3f8583a54524f96f5c382a898ce55a5L169

PR #3765, opened 2024-10-19
Close ZipOutputStream in eclipse dependency downloader

Rawi01:
This PR fixes the broken eclipse tests


rzwitserloot:
I added another commit on top; perhaps a strong word but I _hate_ `closeQuietly` and hence it could not stand.

One problem with my updates is that most linter tools are likely to dislike it, but, the problem is, all linters are bad and wrong, at least in regards to streams. An annoyance; do we ugly up our code to satisfy linters, or do we die on the hill?

I'm a romantic, so, you know the choice I prefer :)

PR #3801, opened 2024-12-23
update gradle syntax to support both `.gradle` and `.gradle.kts`

ZZZank:
`.gradle.kts`, unlike `.gradle`, is using Kotlin instead of Groovy as the base of its scripting language. The old syntax is problematic in `.gradle.kts` becasue:
- Kotlin string cannot be wrapped by single quote, aka `'some string'` is not allowed and should be `"some string"` instead
- Kotlin requires brackets for method calling, aka `f someArg` is not allowed and should be `f(someArg)` instead

So I made some changes to the syntax so that the installation guide is valid for both `.gradle.kts`, unlike `.gradle`

build.gradle.kts:
example-kotlin

build.gradle:
example-groovy

### DIFF
--- a/website/templates/setup/gradle.html
+++ b/website/templates/setup/gradle.html
@@ -23,11 +23,11 @@ repositories {
 }
 
 dependencies {
-	compileOnly 'org.projectlombok:lombok:${version}'
-	annotationProcessor 'org.projectlombok:lombok:${version}'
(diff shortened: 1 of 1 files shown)


rzwitserloot:
Ooh, you managed to make that work for both forms, that's quite convenient. Gracias, @ZZZank!

PR #3807, opened 2025-01-08
Update onXExample_pre.jpage

jevanlingen:
Update `onX` example, to make the JDK 8+ `onConstructor` syntax more clear.

Notice, the written documentation is already on point:
> On javac7, to use any of the 3 onX features, you must wrap the annotations to be applied to the constructor / method / parameter in` @__(@AnnotationGoesHere)`. On javac8 and up, you add an underscore after onMethod, onParam, or onConstructor.

### DIFF
--- a/website/usageExamples/experimental/onXExample_pre.jpage
+++ b/website/usageExamples/experimental/onXExample_pre.jpage
@@ -7,7 +7,8 @@ import javax.persistence.Id;
 import javax.persistence.Column;
 import javax.validation.constraints.Max;
 
-@AllArgsConstructor(onConstructor=@__(@Inject))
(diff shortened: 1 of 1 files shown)


jevanlingen:
The documentation also mentioned:
> Possibly java 9 will offer (much) better ways of supporting this feature.

I don't know if java 9 did support this better, but at least that line should be updated (or removed) as well.

rzwitserloot:
Thanks @jevanlingen – I removed the JDK7 examples. We support all the way back to JDK6 as long as it's not much effort, but as a policy our docs don't cater to anything older than JDK8. Based primarily on reports from the trenches: JDK8 is still in quite some use, but JDK7 and below is quite rare.

JDK9 does not support this better. That hopeful note has not panned out. I removed it and wiped away a tear.

PR #3811, opened 2025-01-17
Avoid generate @Delegate method body NPE

eeoun:
Avoid generate @Delegate method body NPE when deletegate target is null on javac's HandleDelegate

Add a null check before delegate call,now is only for javac

before:
```java
    @Generated
    public void asd() {
        this.getSd().asd();
    }

    @Generated
    public String getExt() {
        return this.getSd().getExt();
    }

    @Generated
    public void setExt(final String ext) {
        this.getSd().setExt(ext);
    }
```
after
```java
    @Generated
    public void asd() {
        if (this.getSd() != null) {
            this.getSd().asd();
        }
    }

    @Generated
    public String getExt() {
        return this.getSd() == null ? null : this.getSd().getExt();
    }

    @Generated
    public void setExt(final String ext) {
        if (this.getSd() != null) {
            this.getSd().setExt(ext);
        }
    }
```

### DIFF
--- a/src/core/lombok/experimental/Delegate.java
+++ b/src/core/lombok/experimental/Delegate.java
@@ -65,4 +65,11 @@ public @interface Delegate {
 	 * @return For each method (not already in {@code java.lang.Object}) in these types, skip generating a delegate method (overrides {@code types()}).
 	 */
 	Class<?>[] excludes() default {};
+
+	/**
+	 * If configuration turn to true. When field is null and target type delegate method return as a references type, just return a null instead of throw a NPE
+	 *
+	 * @return Is it needed to avoid NPE make the utmost effort.
+	 */
+	boolean skipAndReturnNullOnNullRef() default false;
 }
--- a/src/core/lombok/javac/handlers/HandleDelegate.java
+++ b/src/core/lombok/javac/handlers/HandleDelegate.java
@@ -23,6 +23,8 @@ package lombok.javac.handlers;
 
 import static com.sun.tools.javac.code.Flags.*;
 import static lombok.core.handlers.HandlerUtil.handleExperimentalFlagUsage;
+import static lombok.javac.Javac.CTC_EQUAL;
+import static lombok.javac.Javac.CTC_BOT;
 import static lombok.javac.handlers.JavacHandlerUtil.*;
 
 import java.lang.annotation.Annotation;
@@ -72,13 +74,8 @@ import lombok.core.AST.Kind;
 import lombok.core.AnnotationValues;
 import lombok.core.HandlerPriority;
 import lombok.experimental.Delegate;
-import lombok.javac.FindTypeVarScanner;
-import lombok.javac.JavacAnnotationHandler;
-import lombok.javac.JavacNode;
-import lombok.javac.JavacResolution;
+import lombok.javac.*;
 import lombok.javac.JavacResolution.TypeNotConvertibleException;
-import lombok.javac.JavacTreeMaker;
-import lombok.javac.ResolutionResetNeeded;
 import lombok.permit.Permit;
 import lombok.spi.Provides;
 
@@ -111,7 +108,7 @@ public class HandleDelegate extends JavacAnnotationHandler<Delegate> {
 		
 		@SuppressWarnings("deprecation") Class<? extends Annotation> oldDelegate = lombok.Delegate.class;
 		deleteAnnotationIfNeccessary(annotationNode, Delegate.class, oldDelegate);
-		
+
 		Type delegateType;
 		Name delegateName = annotationNode.toName(annotationNode.up().getName());
 		DelegateReceiver delegateReceiver;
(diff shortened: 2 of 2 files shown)


eeoun:
There is also e ecj's implement version. But i've not ever test event compile it. But I think it's ok. I'll add in PR if you think is ok.

[commit f426b66bc5ef7cdee88c4bffeb163155a411ae00](https://github.com/eeoun/lombok/commit/f426b66bc5ef7cdee88c4bffeb163155a411ae00)

victorwss:
Not sure if this is a good idea. Let's see this case:

```java
public interface Foo {
    /**
     * Do a foo thing. Never, ever returns {@code null}.
     * It is very important that in no circunstances {@code null} is ever returned.
     * @return Anything, everything or something, except {@code null}.
     */
    public String foo();
}
```

```java
public class FooDecorator implements Foo {
    @Delegate(types = Foo.class)
    private Foo myFooField = null;
}
```

eeoun:
> Not sure if this is a good idea. Let's see this case:
> 
> ```java
> public interface Foo {
>     /**
>      * Do a foo thing. Never, ever returns {@code null}.
>      * It is very important that in no circunstances {@code null} is ever returned.
>      * @return Anything, everything or something, except {@code null}.
>      */
>     public String foo();
> }
> ```
> 
> ```java
> public class FooDecorator implements Foo {
>     @Delegate(types = Foo.class)
>     private Foo myFooField = null;
> }
> ```

Well in this case, the code runs and stops immediately throwing a NPE. It May useful for some scenario want expose bug earlier in test time.

In the samethime in some use case like DTO case, need expose additioal field but don't want add it to core model
```java
class RestControllerShopItemReturnDTO{
    
    private String someAdditionalShitFieldFromShitService;

    @Delegate
    private ShopItemDTO item = null;

    @Delegate
    private PriceDTO price = null;
}

class ShopItemDTO{

}
```

according to query condition it may contains price or not.

May be add a method in annoation ` @Delegate` defination makes developer control it is a good idea. And it won't change the default behavier compare with old versions.
```java
boolean skipWhenFieldNull() default false;
```

finally it looks like this
```java
class RestControllerShopItemReturnDTO{
    
    private String someAdditionalShitFieldFromShitService;

    @Delegate(skipWhenFieldNull=true)
    private ShopItemDTO item = null;

    @Delegate(skipWhenFieldNull=true)
    private PriceDTO price = null;
}

class ShopItemDTO{

}
```

Rawi01:
I converted this to a draft because:
- I'm not sure if @rzwitserloot and @rspilker approve of the idea.
- It doesn't handle primitive values.
- There are no tests.
- It introduces a breaking change.

While the actual modification is straightforward and could be adjusted to avoid being a breaking change, making it work with primitives might be a bit more complex. I recommend waiting for Reinier or Roel to approve the idea before investing time in implementing it.

eeoun:
Yes it is, primitive type is a serious issue. I'll try more solutions.

rzwitserloot:
I'm going to deny this. You're right, `null` + `@Delegate` is slightly painful.

Except, it isn't. You're using null wrong.

That's an overwrought statement, in the sense that I'm talking to the java community at large and not you specifically. The java community uses null wrong.

SQL has the right idea. `null` does not mean 'default' or 'empty' or 'basic'.

It should mean only one thing: __unknown__.

If you want to delegate to an object, and the thing that points at the object you want to delegate to (the field) has value `null` that means: Delegate to.. um... this.. er, object.. which.. uh.. I don't know what it is supposed to be right now.

NPE is correct. __Silently doing nothing is seriously wrong__. It leads to bugs where you do a thing and nothing happens. No errors. No stack trace. Just.. nothing. Quite literally this is 50x worse. In that it takes you 50x more time to find this bug than one that gives you a stack trace. Hence, even if this code ends up 'doing what you wanted' 48 out of 49 times, you've made things worse, not better.

In some ways this mission is either David taking a swing at Goliath, or possibly Don Quixote tilting at windmills, and I don't think lombok is going to just go ahead and fix the java community's long standing idiotic take on what `null` is supposed to be. It's more of a policy: Trying to cater to all these variant takes on `null` simply can't possibly lead to good things. This 'null means unknown, therefore, embrace the NullPointerException' so happens to remain fairly consistent, and is generally easy to do. Hence, we prefer that view.

And under that view, this PR is never gonna happen.

rzwitserloot:
NB: If this is you and you wanted `null` to mean default, then the solution is to create a 'dummy': A version of the class that represents whatever your erroneous use of null was meant to represent.

For example, if you're writing this:

```java
String x = someGetter();
if (x == null || x.isEmpty()) doAThing();
```

You've done it wrong. Clearly you intend `null` to mean the same thing as `""`. So __return that instead__. `someGetter()` shouldn't have returned null in the first place. It should have returned `""`.

In the somewhat exotic scenario where there __is__ a meaningful difference from the point of view of `someGetter()` between `null` and `""`, but, this particular code that calls it so happens to want to do the same thing in both scenarios, either that `if` is correct, but, lombok should not attempt to cater to it (the `||` in your if matches semantics: You have some code that should run if one of two completely unrelated conditions is true). If you have more code to run that deals with `x`, you should instead follow up and fix:

```java
String x = someGetter();
if (x == null) x = "";
```

and go from there.

For strings and lists this is usually simple (generally this 'default' thing people incorrectly think `null` should mean already exists: `""` and `List.of()`), for other objects: Copy List's homework. Make a method that produces an immutable object that does this correct default thing for all methods that your class has.

Once you have such a thing, make sure you field is never null, instead, it's a ref to this 'default/dummy' thing, and now delegate does what you want, in a non-magical way.

PR #3815, opened 2025-02-04
docs(locked): fix introduced version

tiennm99:
`@Locked` was introduced in lombok v1.18.32, but it's v1.20 in the docs and website


Rawi01:
Thanks!

PR #3816, opened 2025-02-05
 Add Configuration Key lombok.builder.toBuilder (Fixes #3758)

### LINKED ISSUE #3758: [FEATURE]
According to the testing methodology adopted in the company, the template of test object is created before the tests. Then in each test, only one field of the copy recieved via `toBuilder` is changed. When testing GrapgQL services, we get POJO classes by generating code using `mapstruct-processor` and `lombok-mapstruct-binding`. We have encountered a problem that it is not possible to enable `toBuilder` in any way in this case. Please consider adding an appropriate configuration key to `lombok.config` that sets `toBuilder = true` by default for all classes in scope
```
lombok.builder.toBuilder = true
```
This will give developers more control over the result of code generation, especially on testing projects.

Mahmoud-Khawaja:
## Description  
This PR introduces the `lombok.builder.toBuilder` configuration key, allowing `toBuilder()` to be enabled globally for all `@Builder` annotations via `lombok.config`. If not explicitly set in the annotation, the configuration value is applied (default: `false`).  

## Changes  
- Added `BUILDER_TO_BUILDER` key in `ConfigurationKeys.java`.  
- Integrated config key handling in `HandleBuilder.java`.  
- Ensured explicit annotation settings override configuration values.  

This provides more flexibility in testing and code generation.  
Fixes #3758.


rzwitserloot:
I'm not entirely sure lombok.config is the right place for this sort of thing. Config is always fighting with itself: On one hand, lombok never was intended to be silent magic - the source file on its own without knowing what `lombok.config` contains ought to be enough to at least guesstimate what's happening.

Then again, your class still has `@Builder` on it; it should therefore be somewhat obvious where `toBuilder()` is coming from.

I'll discuss this one with @rspilker before merging.

One consideration fighting against this, @Mahmoud-Khawaja, is OpenJDK's plans for `with`. It might be limited to records (all the more reason to use `toBuilder()` more!), but it would look something like:

```java
LocalDate x = fetchSomeLocalDate();
LocalDate y = x with {
  year = 10;
  month = 12;
};
```

If that was part of OpenJDK, it reduces the boilerplate considerations (the 'bang' part of the 'bang for the buck' equation that lombok features have to balance out).

rzwitserloot:
Reviewed the code:

* The merge conflict is trivial; `master` now has a new config key added at the end. The order between yours and the new one is not important.
* Your one commit contains the update, but no docs (that's allright, we can write them), no tests (acceptable, we'll write them), and a whole boatload of weird docker stuff. That has to go.
* Please add your name to the `AUTHORS` file for legal reasons.

If you could add the name to this PR we can accept it. And we'd take it as a kindness if you rebase it, remove the docker stuff.

Mahmoud-Khawaja:
done, thanks!

cruftex:
ChatGPT is telling me already that the option is existing, but it seems the PR is still in progress. It would be great to have it included.

@Mahmoud-Khawaja 
Can you please check your PR and clean it? Please check the changed files to contain only the intended changes. There is docker and Gradle things that that are unrelated. I think its best to squash in to a single commit and then force push.

rzwitserloot:
@cruftex Just a life tip: "ChatGPT told me X" means nothing. LLM output has potentially useful _creative_ value and totally, utterly pointless truth value.

blacelle:
@rzwitserloot Do you think this would be merged soon, for availability into edge? TY

PR #3826, opened 2025-02-12
Set missing source position for receiver annotation

### LINKED ISSUE #3823: [BUG] Eclipse-I build broken
**Describe the bug**
The current eclipse-I build is broken.

**To Reproduce**
See https://github.com/projectlombok/lombok/actions/runs/13164121540/job/36740043178

 - Lombok version: latest
 - Platform: Eclipse-I

Rawi01:
This PR fixes #3823

The position was missing for a long time but was never noticed because our test system was using an old JLS version that doesn't support receivers. I also bumped the java version we use to run the full eclipse tests because the upcoming version requires JDK21 or newer. I also added the missing eclipse 2024-12 test target.

### DIFF
--- a/.github/workflows/ant.yml
+++ b/.github/workflows/ant.yml
@@ -100,6 +100,7 @@ jobs:
         - eclipse-202403
         - eclipse-202406
         - eclipse-202409
+        - eclipse-202412
         - eclipse-I-build
         - eclipse-oxygen-full
         - eclipse-2022-03-full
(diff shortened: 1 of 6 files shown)


mmoayyed:
Would I be correct in stating that this PR also solves this problem https://github.com/google/error-prone/issues/4918?

Rawi01:
No, this PR only adds positions in eclipse for receiver annotations.

The Error Prone check probably fails because it assumes the string list is never empty. However, for code generated by lombok, the source positions of generated nodes are set to the generating annotation, which result in an empty list of strings. Adding a empty check before getting the last element should fix the issue.

rzwitserloot:
I implemented a small part of this because I was a dunce and didn't check PRs before delving into a failing test.

rzwitserloot:
merged via cherry pick to avoid a really mess merge christmas tree in git.

PR #3837, opened 2025-03-04
Add jspecify Nonnull/Nullable annotations

moonfruit:
The `spring-framework` has now abandoned its own definition of `Nonnull/Nullable` annotations in favor of those defined by `jspecify`.
As one of the most successful Java open source project, I'm sure more and more people will follow it and start using `jspecify`.
So I think it's important for `lombok` to provide predefined configuration templates for `jspecify` to simplify its use.


Rawi01 (changes requested):
Thanks for your contribution :+1: 

- Please update the comment in `ConfigurationKeys` and the documentation in `configuration.html` to include the new key

moonfruit:
OK. I'm done. Please help me check if there are any more changes that need to be made.

knutwannheden:
At some point it would be awesome if Lombok could support scoped annotations like JSpecify's `@NullMarked`. Without that I doubt that Lombok will be able to leverage JSpecify very much.

rzwitserloot:
This has been merged as commit 0d0aba6. Not quite sure why github hasn't picked up on that.

PR #3856, opened 2025-03-30
Fix github workflow YAML configuration

mmoayyed:
Github Actions workflow appears to be failing because of invalid YAML configuration and indentation issues. This PR should fix that. 

image


rzwitserloot:
Whoops, that was me. Thanks!

Issue #3769, opened 2024-10-28
[BUG] The subclass field does not overwrite the parent class


qq31715879:
```java
package com.kael.test.dto;

import com.fasterxml.jackson.annotation.JsonIgnore;
import lombok.AllArgsConstructor;
import lombok.Getter;
import lombok.NoArgsConstructor;
import lombok.Setter;
import lombok.experimental.SuperBuilder;


@Setter
@Getter
@SuperBuilder
@AllArgsConstructor
@NoArgsConstructor
public class TestParentDTO {

    @JsonIgnore
    private String demo;
    private String ddd;
}

package com.kael.test.dto;

import com.jugg.agile.framework.core.util.io.serialize.json.JaJson;
import lombok.AllArgsConstructor;
import lombok.Getter;
import lombok.NoArgsConstructor;
import lombok.Setter;
import lombok.experimental.SuperBuilder;


@Setter
@Getter
@SuperBuilder
@AllArgsConstructor
@NoArgsConstructor
public class TestChildrenDTO extends TestParentDTO {

    private String demo;

    public static void main(String[] args) {
        TestChildrenDTO testChildrenDTO = new TestChildrenDTO();
        testChildrenDTO.setDemo("ddemo");
        testChildrenDTO.setDdd("ddd");

        System.out.println(JaJson.toString(testChildrenDTO));
    }
}

```

Subclass fields have removed annotations  @JsonIgnore

------------------------------------------
version 1.8.12 can runsuccess
![image](https://github.com/user-attachments/assets/459d0cda-7378-4163-8e0d-94317e7c18ec)

-------------------------------------
but High version failed to run
![image](https://github.com/user-attachments/assets/d8b43935-1b23-4c0d-8bff-da265ded1108)

janrieke:
Things probably changed when Lombok started copying certain annotation to the setter with 1.18.16, including `@JsonIgnore`. Jackson can sometimes be confusing when it comes to detecting which annotations belong to a property, because builds a combination of annotations on the field and its corresponding setters and getters.

With 1.18.12, the annotations was only on the field of the superclass, and Jackson correctly ignores only the field from the parent class. However, once Lombok copies it to the setter, I suspect that Jackson considers this setter method as if it belongs to the property defined in the subclass.

I tend towards "this is a Jackson bug", although you could argue that the setter method is overridden and thus it's the same method. On the other hand, `@JsonIgnore` is not inheritable, so Jackson should not consider it for the subclass property.

[...]

Issue #3270, opened 2022-09-29
[BUG] Should @Jacksonized add @JsonProperty to fluent methods?


mjaggard:
**Describe the bug**
When I apply `@Builder` and `@Jacksonzied` along with `@Data` or `@Value` and `@Accessors(fluent = true)` my object gets deserialised correctly but serialisation doesn't work and I end up with `{}`

**To Reproduce**
```
@Value
@Builder
@Jacksonized
@Accessors(fluent = true)
public static class MyObject {
        String objectName;
}

objectMapper.writeValueAsString(MyObject.builder().objectName("name").build());
```

**Expected behavior**
The output should be `{"objectName":"name"}`

**Version info (please complete the following information):**
 - Lombok version 1.8.22
 - All platforms

**Additional context**
I'm migrating from Immutables which solves this by adding `@JsonProperty` to all of the Immutable object's fluent getters.

mjaggard:
Possibly related: https://github.com/projectlombok/lombok/issues/3265

janrieke:
That's not a bug, that's a new feature. `@Jacksonized` only affects `@(Super)Builder`. Thus, it is only relevant to deserialization.

We can talk about extending it to `@Accessors`, but I'm quite sure that would mean more than just adding `@JsonProperty`.

mjaggard:
I'm aware that `@Jacksonized` is only there for Builders but arguably a user is reasonable in expecting it to work correctly for Serialisation too which is why I labelled it "BUG".

`@JsonProperty("fieldname")` on the getter seems to solve the problem for me.

rzwitserloot:
Fixed by @janrieke

Issue #3263, opened 2022-09-22
[BUG] lombok update from 1.18.12 to 1.18.20 change  jackson format


likaer:
**Describe the bug**
I used jackson-databind 2.10.3 and spring-web 5.2.5 to parse response body as json.
Assume I have a request api like this
```
@PostMapping(PREFIX_PATH + "/findFull")
ChildVO anyway(@RequestBody AnyQueryDTO queryDTO);


@Data
public class ChildVO extends BaseVO {

    private LocalDateTime createTime;

    private LocalDateTime updateTime;

}

@Data
public abstract class BaseVO implements Serializable {
     //same as DateUtils.DATE_TIME_FORMAT
     public static final String DATE_TIME_FORMAT = "yyyy-MM-dd HH:mm:ss";

    @JsonFormat(pattern = DateUtil.DATE_TIME_FORMAT)
    private LocalDateTime createTime;

    @JsonFormat(pattern = DateUtil.DATE_TIME_FORMAT)
    private LocalDateTime updateTime;

}

```

**To Reproduce**

when I use version of **1.18.12**,   **objectMapper** output **2022-09-21T11:28:24**

when I use version of **1.18.20**,   **objectMapper** output **2022-09-21 11:28:24**

why?

janrieke:
Probably because Lombok started copying `@JsonFormat` to the getter. Before (without the annotation on the getter), Jackson does not see this annotation anywhere in the actual class `ChildVO`. Now Lombok copies it to the getter, and Jackson recognizes it because the setter is (although inherited) visible in the `ChildVO` class.

PR #3860, opened 2025-04-01
New Feature: @Jacksonized @Accessors

### LINKED ISSUE #3265: Generate fluent methods plus Getter and Setters
Using Accessors with fluent and Data will not generate getters and setters but only fluent methods.

eg.:
```
@Accessors(fluent = true, chain = true)
@Data
public class TestFluent{

  private String test;
}
```

Only a fluent method will be generated: 
```
  public TestFluent test(final String test) {
    this.test = test;
    return this;
  }
```

I would also need getters and setters like:
```
  public String getTest() {
    return test;
  }

  public void setTest(final String test) {
    this.test = test;
  }
```

janrieke:
The PR adds a new feature for the existing `@Jacksonized` annotation. Up to now, `@Jacksonized` only affected `@Builder` or `@SuperBuilder`. With this PR, `@Jacksonized` can be used in combination with `@Accessors(fluent=true)`. 

Jackson identifies corresponding accessors following the beanspec conventions. If you use fluent accessors, Jackson does not make the connection between the field and the corresponding accessors, leading to strange behavior when de-/serializing. You can manually mitigate this by adding `@JsonPropery("propertyName")` annotations on the field, the setter and the getter (using `@Getter/@Setter(onMethod_={@JsonPropery("propertyName")})`). However, this is cumbersome and error-prone (because when renaming a field you also have to manually adjust all annotations).

With `@Jacksonized @Accessors(fluent=true)`, lombok will automatically create the relevant annotations such that Jackson correctly identifies fluent accessors.

You can simply write classes like this:
```
@Jacksonized @Accessors(fluent = true)
@Getter @Setter
public class MyDTO {
	private int field;
}
```
Lombok will generate fluent getters and setters with the required Jackson annotations, such that this class can be de-/serialized by Jackson without further modifications.

This resolves #3265, #3270.
Related issue: #2141

**Further changes**:
Together with this new feature, this PR changes a lombok behavior that has caused some issues since its introduction in v1.18.16. With that version, lombok started automatically copying certain Jackson annotations to the setter. However, this behavior cannot be switched off, and it may even be harmful in certain situations (e.g. #3769, #2769, #3263): Jackson considers some annotations inheritable, and this "virtual inheritance" only affects annotations on setter/getter, but not on private fields. Automatically copying annotations therefore may change the de-/serialization behavior of subclasses.

In general, it is not necessary to copy those annotations at all: Jackson considers the Java field and the corresponding setter and getter methods to be connected (given that Jackson can "beanspec"-identify the connection), and it is sufficient for an annotation to be present on one of those to become effective _for all_. Thus, the new default behavior with this PR is not to copy, but you can switch back to the previous behavior using `lombok.copyJacksonAnnotationsToAccessors = true`.


rzwitserloot:
Awesome work. Nothing of note during review; just used it as a springboard to make a couple of style things more consistent.

PR #3888, opened 2025-05-27
Support records in upcoming eclipse version

### LINKED ISSUE #3883: [BUG] @Builder annotation broken with Java records
See https://github.com/redhat-developer/vscode-java/issues/4050

**Describe the bug**
Annotating a Java record with `@Builder` produces `The constructor ... is undefined`

**To Reproduce**
Use vscode-java 1.42.0 and with a project that uses lombok, simply annotate a record with `@Builder`.

```java
package org.example;

import java.util.List;
import lombok.Builder;

@Builder
public record Example(List<String> list) {
}
```

This probably happens with any Lombok version because vscode-java 1.42.0 uses JDT-LS 1.47.0, which is using the latest JDT Core. Recently, JDT Core made some updates to records : https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3925 , https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3980 so likely that lombok needs to adjust some of its usage of JDT Core.

Rawi01:
This PR fixes #3883

Eclipse changed how they handle records. Instead of generating fields early on it now does it in a later stage. To keep the support in lombok we now map the record components to fields ourself and add theses fields to our AST. To keep the test output the same I added a patch that removes the implicit fields from the output if they are present.

### DIFF
--- a/.github/workflows/ant.yml
+++ b/.github/workflows/ant.yml
@@ -94,6 +94,9 @@ jobs:
           - eclipse-oxygen
           - eclipse-202006
           - eclipse-202006-jdk8
+          - eclipse-202403
+          - eclipse-202503
+          - eclipse-I-build
           - eclipse-oxygen-full
           - eclipse-202403-full
           - eclipse-202503-full
--- a/buildScripts/setup.ant.xml
+++ b/buildScripts/setup.ant.xml
@@ -205,6 +205,8 @@ This buildfile is part of projectlombok.org. It sets up the build itself.
 					<!-- osgi.extender dependecies -->
 					<arg value="osgi.bundle:org.apache.felix.scr" />
 					<arg value="osgi.bundle:org.apache.aries.spifly.dynamic.bundle" />
+					<!-- Unresolvable until we add property based dependency resolution -->
+					<arg value="osgi.bundle:org.eclipse.swt.svg" />
 					<!-- Real dependencies -->
 					<arg value="osgi.bundle:org.eclipse.jdt.core" />
 					<arg value="osgi.bundle:org.eclipse.jdt.ui" />
--- a/src/core/lombok/eclipse/EclipseAST.java
+++ b/src/core/lombok/eclipse/EclipseAST.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2009-2021 The Project Lombok Authors.
+ * Copyright (C) 2009-2025 The Project Lombok Authors.
  * 
  * Permission is hereby granted, free of charge, to any person obtaining a copy
  * of this software and associated documentation files (the "Software"), to deal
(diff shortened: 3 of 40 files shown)


rgrunber (approved):
I tried this out and it works well for me. Records with a `@Builder` annotation no longer emit an error. "Document does not match AST" also remains fixed.

durvalprintes:
I built the branch and tested the new jar, the errors disappeared.

hashpyrit:
The problem still occurs if you you do custom constructor field validation for your record:
```
public record FailedEvent(Long assetId, DigitalAssetIdentifier assetIdentifier, @NonNull String user, String failureMessage) {

    @Builder <--- this line gets the error that the constructor is not defined
    public FailedEvent {
        if (assetId == null && assetIdentifier == null) {
            throw new IllegalArgumentException("Either assetId or assetIdentifier must be provided");
        }
    }
}
```


I downloaded lombok-edge.jar today as well as I tried the lombok-1.18.39-4050.jar.  Fails with both.  I'm using Spring Tools For Eclipse 4.31.0 which is using Eclipse JDT 3.20.200.v20250528-1830.

vkushnir-onseo:
Hello! Dear @Rawi01 just cloned this branch, run `ant dist` and tried generated jar with eclipse and issue still there. Few possible cases: I did smth wrong or fix is not complete or eclipse again changed logic related to records handling
```
Version: 2025-06 (4.36.0)
Build id: 20250605-1316
Lombok v1.18.37 "Edgy Guinea Pig" - 2025-07-09 is installed. https://projectlombok.org/
```

Rawi01:
@vkushnir-onseo This branch should generate a build of version 1.18.39. You either built the wrong branch or made a mistake in your eclipse setup. Rather than building it yourself, you can also download the artifact from the GitHub Actions run. (https://github.com/projectlombok/lombok/actions/runs/15284242422?pr=3888)

alphademirkaya-hub:
@Rawi01 thanks a lot it works!

rzwitserloot (commented):
@Rawi01 Without spending some time delving into why we added 'hey if you see TerminalToken being loaded, apply this patch' is no longer needed – trying to get a grip on why we can simply the fixing of `dom.rewrite.TokenScanner`. Can you confirm we no longer need it / OSGI check takes care of it?

For context: Up to ~3 months ago, this fix was _only_ applied if we witnessed TerminalToken. 3 months ago, this was updated to 'either if you see any OSGI signs _or_ if you see TerminalToken, apply it' and now we've arrived at: "If you see any OSGI signs, apply it. Don't worry about TerminalToken".

EDIT: Hmm, not sure github makes it possible to see what this is referring to. It's about commit 'Use simpler addScriptIfWitness', 98d980d.

rgrunber:
@rzwitserloot I think that commit was related to https://github.com/projectlombok/lombok/pull/3885#issuecomment-2910704421 . Upstream replaces `TerminalTokens` with `TerminalToken` in newer versions. I think the logic should be "if you see `TerminalToken`, we're using a newer Eclipse runtime, so apply the patch".

Issue #3835, opened 2025-02-25
[FEATURE] A single annotation to produce both getter and setter (e.g. @GetSetter)


cchai-newtrek:
**Describe the feature**
A single annotation to produce both getter and setter (e.g. ```@GetSetter```, similar to ```@Data``` but without the hashcode and toString feature), since it is very common to produce both getter and setter in a java POJO, so we can have 3 options: ```@Getter```, ```@Setter``` or both ```@GetSetter``` to further reduce the boilerplate

**Describe the target audience**
Reduce boilerplate, tired of repeatedly annotate with ```@Getter``` and ```@Setter``` many times

**Additional context**
NIL

fabriciuMF:
Could it be @GSet, it is better to remember, and also, if there is an specific implementation of the get or set, we should avoid the default getter and setter and prioritize the customized method instead?

PR #3890, opened 2025-05-28
Add a single annotation replacing the use of @Getter and @Setter both

JNGDSOFT:
This PR is for the issue #3835

Adds a new annotation that combines **Getter** and **Setter** annotations into a single one **Gset**

I may have missed some required configurations details - particularly, this annotation **doesn´t support parameters** yet and **I haven't been able to test it fully** due to some local environment issues, that´s why I wanted to share this early version in case it's helpful or can be built upon.

Feedback is very welcome. Thanks!

PD: "GSet" name were based on a suggestion from @fabriciuMF


rust2:
I think the name `@Property` would suit it better

Issue #2280, opened 2019-11-02
[FEATURE] @Log(accessLevel)


DaBlick:
I checked the Wiki to see if this is a duplciate but to me, the wiki seems empty.  Apologies if this is a duplicate.

### Describe the Feature

I would like to be able to choose some alternative access level modifiers to `private `when I define a logger.    
This might be done with something like

        @Log( accessLevel = "protected")

### Target Audience 

While I won't claim that something other than `private` is always a _best practice_, I do feel like there are a few common scenarios (such as choosing `protected `in a `abstract `base class) where it's quite defensible.

- Writing an abstract base class that provides a shared logger to subclasses

- Exporting a class logger via a public getter either `public `or to other members of a `package`

rzwitserloot:
exporting your logger would presuppose that you name or categorize the logger in a way that is different from categorizing it according to class names, because, it makes no sense for code in class X to send log statements to a logger it got from elsewhere; a logger named 'class Y'.

Hence, this feature makes no sense to me. When do you want to have an 'exported' logger that nevertheless should have the name of the class it is from?

DaBlick:
What we want to do is to be able to define a logger in a superclass and have it shared among all subclasses.     While this isn't the typical use case, I feel it is perfectly valid.

Part of what we're trying to achieve is, ironically, to avoid having "boilerplate" Lombok annotations.

We are not "exporting" it publicly, but to subclasses.    (And potentially other things in the same package).

janrieke:
I really think this is a bad idea. For instance, if you configure your output to contain the line in the source code, this will result in maximum confusion if someone is later trying to find the location of a log line.

Having a static logger field will have almost no effect on initialisation performance and memory footprint, and no effect on runtime performance. Am I missing a reason why you would want this?

DaBlick:
@janrieke The purpose here is not to minimize runtime performance.   

The goal, as I said before, is to minimize “boilerplate code” in this case, a set of “boilerplate” annotations and bean properties on methods.   The particular use cases here are controller and service layer classes which for our app tend to be very very small.

So the real goal here is to minimize to have the code in each concrete subclass be highly focused on things that are SPECIFIC to that subclass and not things that every sibling subclass does (define loggers, etc.)

rzwitserloot:
[...]

PR #3922, opened 2025-08-19
@Logger(AccessLevel)

liampace:
This pull request is related to Issue #2280.
When this issue was initially presented it was mildly controversial because it would require the ability to directly set the logger's name. Fast forward to today and now the `topic` parameter is officially supported within the various logging annotations, thus making the main drawback null and void. I have provided an implementation for this feature as while the average developer may not need it, those working with micro-services, embedded development, and game programming to name a few would greatly benefit from it.

This feature works just like `@Getter/@Setter`'s `access` parameter with one caveat:
- The default value for `access` is `AccessLevel.PRIVATE`

Where is what the feature looks like:
```java
package <org>.graphics;

@Log(access = AccessLevel.PROTECTED, topic="RenderPipeline")
public abstract class Graphics2D {
    public static class Graphics2DFactory {
        public Graphics2D create(String classname) {
            // create and return instance
        }
   }
    // ... draw/render methods 
}
```
```java
package <org>.graphics.opengl;
import static <org>.graphics.Graphics2D.log;

// This class can now use the provided logger from Graphics2D
class OpenGLGraphics2DImpl extends Graphics2D {
    // implement methods from Graphics2D
}
```
```java
package <org>.graphics.vulkan;
import static <org>.graphics.Graphics2D.log;

// This class can now use the provided logger from Graphics2D
class VulkanGraphics2DImpl extends Graphics2D {
    // implement methods from Graphics2D
}
```

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -26,6 +26,7 @@ John Paul Taylor II <johnpaultaylorii@gmail.com>
 Karthik kathari <44122128+varkart@users.noreply.github.com>
 Kevin Chirls <kchirls@users.noreply.github.com>
 Liu DongMiao <liudongmiao@gmail.com>
+Liam Pace <liam.hollum.pace@gmail.com>
 Luan Nico <luannico27@gmail.com>
 Maarten Mulders <mthmulders@users.noreply.github.com>
 Manu Sridharan <msridhar@gmail.com>
--- a/docker/gradle/files/classpath/.gitignore
+++ b/docker/gradle/files/classpath/.gitignore
@@ -1,0 +1,5 @@
+build/
+.gradle/
+gradle/
+gradlew
+gradlew.bat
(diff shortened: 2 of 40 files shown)


rzwitserloot:
`access` is not misleading, I doubt anybody is going to misunderstand what it would do. I dislike introducing lombok features that have no purpose other than to enable bad code, but, 

> those working with micro-services, embedded development, and game programming to name a few would greatly benefit from it.

That's all stuff I don't have much experience with. The damage is minimal because it's hard to misunderstand what `access` would do, and I'm not particularly convinced about the veracity of 'non-private loggers bad' .

Accepted. I used a more logical git workflow to my sensibilities (rebase it on master, then merge and push it), but evidently github no longer understands this. Your commits are in as normal, I just moved them around in history.

PR #3945, opened 2025-09-18
[BUG] Transient properties in classes with @Jacksonized @Accessors(fluent=true) get serialized

liampace:
This pull request is related to Issue #3936.

Currently the `@Jacksonized` annotation mistakenly applies `@JsonProperty` to transient fields. The new behavior has `@Jacksonized` explicitly apply `@JsonIgnore` to all transient fields.

For example the following sample code:
```java
@Jacksonized
@Accessors(fluent = true)
public class Foo {
    String foo;
    transient String transientFoo;
}
```
would be equivalent to:
```java
public final class Foo {
    @JsonProperty("foo")
    private final String foo;
    @JsonIgnore
    private final transient String transientFoo;
    ...
}
```

PR #3948, opened 2025-09-18
[BUG] Lombok Tests Failing

liampace:
Initially discovered on PR #3945, some unit tests unexpectedly started to fail due to the deprecation of `AccessLevel.MODULE`. This PR attempts to fix these tests and any others along the way. Right now, the failing tests for the various `@Log` annotations new access level feature are now passing locally. I'm still searching for if any other annotations are unexpectedly failing tests.

### DIFF
--- a/test/transform/resource/after-delombok/LoggerLog4j2Access.java
+++ b/test/transform/resource/after-delombok/LoggerLog4j2Access.java
@@ -5,6 +5,7 @@ class LoggerLog4j2AccessPublic {
 	public static final org.apache.logging.log4j.Logger log = org.apache.logging.log4j.LogManager.getLogger(LoggerLog4j2AccessPublic.class);
 
 }
+@SuppressWarnings("deprecation")
 class LoggerLog4j2AccessModule {
 	@java.lang.SuppressWarnings("all")
 	@lombok.Generated
--- a/test/transform/resource/after-delombok/LoggerCommonsAccess.java
+++ b/test/transform/resource/after-delombok/LoggerCommonsAccess.java
@@ -5,6 +5,7 @@ class LoggerCommonsAccessPublic {
 	public static final org.apache.commons.logging.Log log = org.apache.commons.logging.LogFactory.getLog(LoggerCommonsAccessPublic.class);
 
 }
+@SuppressWarnings("deprecation")
 class LoggerCommonsAccessModule {
 	@java.lang.SuppressWarnings("all")
 	@lombok.Generated
--- a/test/transform/resource/after-delombok/LoggerCustomAccess.java
+++ b/test/transform/resource/after-delombok/LoggerCustomAccess.java
@@ -5,6 +5,7 @@ class LoggerCustomAccessPublic {
 	public static final MyLogger log = MyLoggerFactory.create(LoggerCustomAccessPublic.class);
 
 }
+@SuppressWarnings("deprecation")
 class LoggerCustomAccessModule {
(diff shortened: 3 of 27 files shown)


liampace:
I noticed when building the tests using javac6 locally, the `EqualsAndHashCodeOfAndExcludeError.java` message test was failing. I was wondering if this is an actual bug or if the tests were simply not updated:

Terminal Output:
```
[junit] ***** EqualsAndHashCodeOfAndExcludeError.java *****
    [junit] [EqualsAndHashCodeOfAndExcludeError.java] Expected message '2:60 You must use constant literals in lombok annotations; they cannot be references to (static) fields.' but got message '2 ERROR You must use constant literals in lombok annotations; they cannot be references to (static) fields.'
    [junit] **** Expected ******
    [junit] 2:60 You must use constant literals in lombok annotations; they cannot be references to (static) fields.
    [junit] 7:160 You must use constant literals in lombok annotations; they cannot be references to (static) fields.
    [junit] ****  Actual  ******
    [junit] 2 ERROR You must use constant literals in lombok annotations; they cannot be references to (static) fields.
    [junit] 7 ERROR You must use constant literals in lombok annotations; they cannot be references to (static) fields.
    [junit] ****  Actual File  ******
    [junit]    1 // Generated by delombok at Thu Sep 18 16:50:01 EDT 2025
    [junit]    2 // skip-compare-contents
    [junit]    3 final class EqualsAndHashCodeErrorOf {
    [junit]    4        int x;
    [junit]    5 }
    [junit]    6
    [junit]    7 final class EqualsAndHashCodeErrorExclude {
    [junit]    8        int x;
    [junit]    9 }
    [junit]   10
    [junit]   11 class Const {
    [junit]   12        static final String A = "A";
    [junit]   13 }
    [junit]   14
    [junit] *******************
```

liampace:
It seems that the CI checks here are happy now. Let me know if there is anything I may have missed and I'll get it fixed asap.

test/transform/resource/after-delombok/LoggerLog4j2Access.java
```diff
@@ -1,10 +1,12 @@
+//version 8:
```
Rawi01, line 1: Any idea why only this test doesn't work for JDK6 and all the other Logger*Access tests pass without any problems?

Rawi01 (commented):
Thanks for fixing the tests :+1: 

- Instead of adding `@SuppressWarnings` everywhere it is also possible to add a `.messages` file for a test. Our test system compares the generated messages with this file and the test passes if the content matches. Using these files would lead to a smaller and simpler PR but your approach is fine too, just want to let you know for your next PR :)

test/transform/resource/after-delombok/LoggerLog4j2Access.java
liampace: 1) I should already be under the AUTHORS file since I contributed to the `@Log` access feature but I'll double check.
2) The reason that I added the `//version 8:` to the `LoggerLog4j2Access` test is because the regular `LoggerLog4j2` test also uses it. This puts me under the impression that Log4j2 (at least for the most recent versions) does not support versions under JDK-8
3) Thank you for the feedback, I'll make sure to take that into consideration for the future

Rawi01: 1. Sorry missed that :)
2. Makes sense. Spend a few minutes to figure out why it was added in the first place, turns out that the Log4j2 version we use for our tests doesn't support Java 6 anymore.

Issue #3947, opened 2025-09-18
[BUG] Lombok val can be used only once with JDK25 and `--release 8`


asheldon:
**Describe the bug**
When compiling with OpenJDK25, if `val` is used _more than once_ and the target platform is Java 8 or Java 9, all uses of val result in errors like the following:

```
MutliVal.java:6: error: incompatible types: long cannot be converted to val
        val two = 2L;
```

**To Reproduce**

MultiVal.java
---
```
import lombok.val;

public class MutliVal {
    // This does not compile with JDK 25, Lombok 1.18.42 and --release 8 or --release 9.
    public static void jdk25failure() {
        val one = 1L;
        val two = 2L;
    }
}
```

Compiling:
```
javac MutliVal.java -classpath lombok.jar -proc:full --release 8
```

**Expected behavior**
Compilation to succeed as it does with JDK 21 and earlier.

**Version info (please complete the following information):**
 - Lombok version - 1.18.42
 - Platform - javac 25

**Additional context**
n/a

PR #3956, opened 2025-09-25
Suppress messages during resolution

### LINKED ISSUE #3954: [BUG] Random errors during compilation of classes annotated with ExtensionMethod using JDK 25
**Describe the bug**
After migration from JDK 24 to JDK 25, the following compilation errors start to appear in the classes/interfaces annotated with the `@ExtensionMethod`:
```
java: modifier public,static not allowed here
java: modifier default not allowed here
```
The compilation succeeds if the `@ExtensionMethod` annotation is removed.

 - Lombok version 1.18.42
 - Javac version 25

Rawi01:
This PR fixes #3954 and #3947 

-  To suppress messages during resolution lombok injected a temporary message collection into the current diagnostic handler and restored the original one afterwards. Instead of using this error prone approach lombok now installs a  `DiscardDiagnosticHandler` that is designed exactly for this purpose and is also used by the compiler itself.
- The removeAllBetween method now uses a `LinkedList` to be compatible with all known versions of the `Log` class. I also noticed that this code was broken a long time and even if I remove it all our tests pass. Might be possible to remove all that stuff.
- I also noticed that the `compare` method in `JavacResolution` doesn't work because the used `Name::compareTo` returns different results in Java 8

### DIFF
--- a/src/core/lombok/javac/CompilerMessageSuppressor.java
+++ b/src/core/lombok/javac/CompilerMessageSuppressor.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2011-2021 The Project Lombok Authors.
+ * Copyright (C) 2011-2025 The Project Lombok Authors.
  * 
  * Permission is hereby granted, free of charge, to any person obtaining a copy
  * of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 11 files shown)


asheldon:
I've tested this change against the actual code that demonstrated https://github.com/projectlombok/lombok/issues/3947 and it appears to be fixed.

rschmitt:
Any updates on this?

PR #3958, opened 2025-09-25
[FEATURE] Proof of Concept for @After

liampace:
Having looked at the wiki for the [`@After`](https://github.com/projectlombok/lombok/wiki/FEATURE-IDEA:-@After) feature, I have decided to give a crack at implementing it. Right now it is in the beginning stages and none of the features are implemented yet but I'll post updates here as development progresses.


Rawi01:
I created a PR for something similar a few years ago: #2522

liampace:
Would it be okay for me to continue developing this as it is the more generalized version of the concept or would it be considered a duplicate?

PR #3970, opened 2025-10-16
Fix broken folding in eclipse

Rawi01:
In the latest eclipse version the new folding mechanism is broken. This PR fixes this problem. I also added eclipse 2025-09 as test target.

### DIFF
--- a/.github/workflows/ant.yml
+++ b/.github/workflows/ant.yml
@@ -96,10 +96,12 @@ jobs:
           - eclipse-202006-jdk8
           - eclipse-202403
           - eclipse-202503
+          - eclipse-202509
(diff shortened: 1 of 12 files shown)


rzwitserloot:
Ah, fantastic. We've been tasked to give a keynote at JFall which is in 3 weeks, but I'll see if I can sneak it in somewhere. There are a few more wonky things about the current release that suggest we should release a new version soon.

rzwitserloot:
nevermind - not merged as is, this branch pulled in lots of unrelated stuff. Manually merging....

rzwitserloot:
Okay, _now_ it's merged.

PR #3977, opened 2025-10-24
Add @Singular support for Optional types

### LINKED ISSUE #3976: [FEATURE] Allow `@Singular` to be used on Optional
**Describe the feature**
It would be really nice if we could annotated Optional's with @Singular too

```
    @Singular
    Optional<Connectors> connectors,
```

Doing so, the builder would create `connectors(Connectors)` and it would create the Optional instance from that parameter.

velo:
## Summary
This PR adds support for using `@Singular` with `Optional` types in `@Builder`.

## Changes
- Added `JavacJavaUtilOptionalSingularizer` for javac compiler support
- Added `EclipseJavaUtilOptionalSingularizer` for Eclipse compiler support  
- Added comprehensive test cases for Optional singularization

## Behavior
For a field annotated with `@Singular Optional<T>`, the builder generates:
- A setter method accepting the unwrapped type `T`
- A clear method to reset the optional
- Build code that wraps the value in `Optional.ofNullable()` or returns `Optional.empty()`

## Example
```java
@Builder
class Example {
    @Singular
    private Optional<String> name;
}

// Generated methods:
builder.name("John")    // Sets the value
builder.clearName()     // Clears the value
builder.build()         // Returns Optional.ofNullable(value) or Optional.empty()
```

## Test Plan
- [x] Compiles successfully with ant dist
- [x] Added test cases in `test/transform/resource/`
- [ ] Tests should be run with `ant test`

Fixes #3976

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,276 @@
+# AGENTS.md
+
+This file provides guidance to Claude Code (claude.ai/code) when working with code in this repository.
+
+## Project Overview
+
+**Project Lombok** is a Java library that uses annotation processing and bytecode manipulation to eliminate boilerplate code. It operates as a Java agent and annotation processor, supporting multiple compilers (javac, ECJ) and IDEs (Eclipse, IntelliJ).
+
+**License:** MIT
+
+## Build System
+
+### Apache Ant (Primary)
+Lombok uses Ant with Ivy for dependency management. All development tasks use Ant.
+
+**Essential Commands:**
+```bash
+ant quickstart          # Quick setup and help
+ant help               # Detailed help system
+ant help.IDE           # IDE setup help
+ant help.test          # Testing help
+ant help.compile       # Compilation help
+ant help.packaging     # Packaging help
+ant dist               # Build lombok.jar (main artifact)
+ant compile            # Compile source code
+ant compile.support    # Compile build support tools
+ant test               # Run default test suite
+ant test.broad         # Comprehensive testing across platforms
+ant maven              # Package for Maven Central
+```
+
+**IDE Setup:**
+```bash
+ant eclipse            # Generate Eclipse project files (recommended)
+ant intellij           # Generate IntelliJ project files (experimental)
+```
+
+**Note:** Eclipse is strongly recommended for development. IntelliJ support is experimental.
+
+**Platform-Specific Testing:**
+```bash
+ant test.javacCurrent  # Test on current JVM
+ant test.javac11       # Test against JDK 11
+ant test.javac17       # Test against JDK 17
+ant test.ecj11         # Test Eclipse compiler
+ant test.eclipse-oxygen # Test Eclipse Oxygen integration
+```
+
+**Docker Integration Tests:**
+All Docker tests run without local Maven/Gradle/Bazel installation. Tests are in `/docker/` directory organized by build tool and JDK version.
+
+### Build System Details
+
+**Stubs Compilation:**
+Lombok uses a sophisticated compilation process with "stubs" - signature-only versions of library classes. Compilation order:
+1. Compile stubs (signature-only classes from various library versions)
+2. Compile lombok core with stubs on classpath
+3. Package lombok without stubs
+
+Different parts of lombok are compiled with different `-release` targets for compatibility.
+
+**Multi-JDK Testing:**
+- `javac6` and `javac8`: Don't require installed JDKs - downloads complete runtime classes and uses module limits to exclude the current VM's javac
+- `javac11+`: Requires actual JDK installation - build auto-discovers or prompts for path
+- **JVM Configuration:** Create `jvm.locations` file to specify JDK paths:
+  ```
+  j11 = /path/to/jdk-11
+  j17 = /path/to/jdk-17
+  ```
+  Build will auto-create this file when you provide paths on first prompt.
+
+## Contribution Workflow
+
+### Before Starting Development
+**IMPORTANT:** New features must be discussed on the Project Lombok Forum first. Pull requests for new features without prior discussion are unlikely to be accepted.
+
+When proposing features:
+- Provide code examples showing both annotation usage ("with lombok") and generated output ("vanilla Java")
+- Understand existing design considerations to avoid duplication
+- Include proof-of-concept implementations for better consideration
+
+### Development Process
+1. Discuss feature/change on Project Lombok Forum
+2. Fork repository and create feature branch
+3. Implement changes following architecture patterns
+4. Add comprehensive test coverage
+5. Submit pull request via GitHub
+
+## Architecture
+
+### Multi-Compiler Handler Pattern
+Lombok implements platform-specific handlers for each annotation:
+
+- **Annotations:** `src/core/lombok/` (e.g., `@Getter`, `@Setter`, `@Builder`)
+- **Javac Handlers:** `src/core/lombok/javac/handlers/` (e.g., `HandleGetter.java`)
+- **Eclipse Handlers:** `src/core/lombok/eclipse/handlers/` (e.g., `HandleGetter.java`)
+
+Each annotation has separate implementations for javac and Eclipse Compiler (ECJ).
+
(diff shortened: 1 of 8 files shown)


ExE-Boss:
Since this is using [`Optional​::ofNullable(…)`], the `…$set` subfield is redundant.

[`Optional​::ofNullable(…)`]: https://docs.oracle.com/en/java/javase/25/docs/api/java.base/java/util/Optional.html#ofNullable(T)

velo:
> Since this is using [`Optional​::ofNullable(…)`](https://docs.oracle.com/en/java/javase/25/docs/api/java.base/java/util/Optional.html#ofNullable(T)), the `…$set` subfield is redundant.

done

rzwitserloot:
I feel the pain of wanting to live in a `null` less world.

And maybe we can do _something_ here, but we have a problem: We aren't allergic to `null`. We kinda like it, really. Hence we're just not sure what the null-allergic crowd wants.

Lombok can't help you much - java fundamentally has `null` and while I understand your desire to be rid of it, it's not that simple. This PR is vetoed. I'm going to explain our thinking because it is certainly possible we've misunderstood; if you point out where our explanation is wrong, this PR or something similar to it might no longer get vetoed, and en passant there are things we would consider changing when `Optional` params/fields are part of a `@Builder`:

## The terrible sin of operator overloading

(This is not the fundamental issue with this PR. It is _a_ fundamental problem, but one that has a very easy fix!)

Why do we hate operator overloading? I'm speaking half rhetorical here, I'm assuming lots of java programmers don't like it. But, I will focus on why _I_ do not like it:

Because folks will take something that is semantically well defined, such as 'add 2 numbers together, in a commutative, associative, and distributive way', and then take a sidestep into common english to end up using it for 'concatenate two strings'. Oof, I guess we're stuck with that one, given that it's part of the JLS. But, the same thinking leads to defining the `<<` operator on `InputStream` to accept strings which are then turned into ascii bytes and shoved into the stream, or `++` to mean 'add to a collection', immediately causing lots of confusion about whether that produces a new list, or modifies an existing one.

__This proposal commits the exact same sin__. `@Singular` on `Optional` is, in a word, a terrible idea. It makes no sense. Only if you first wear the baggage of knowing what `@Singular` does, and then add 'I find how `@Builder` interacts with `j.u.Optional` annoying', you end up with this proposal.

So that's fundamental problem 1. But that is easily fixed - __if__ this is a good idea, lombok should just always exhibit the behaviour this PR adds for any `Optional`. That would be backwards incompatible which is a new and separate issue.

In other words, any proposal simply cannot use `@Singular` as some sort of hack to fix backwards compatibility or to be some modal switch to decide how we deal with it. Instead, any solution is going to have to go whole hog. __ALL__ uses of Optional get the update or none of em do.

## Semantic confusion

When I _just_ look at the API of a builder and I see 2 methods:

```
Builder name(String newName) { ... }

Builder clearName() { .... }
```

I start thinking 'ternary state'. I start thinking that this system allows and can differentiate __3__ different states:

* `.name("John")`
* `.name(null)`
* `.clearName()`

This isn't nuts - there really can be a semantic difference between 'explicitly `null`' and 'not set at all'. For example, any hierarchical system. For example, maven poms: I may want to express:

* Please copy over whatever defaults are present in my parent in the hierarchy.
* I want to override this setting to the value "John", disregard what my parent sets.
* I explicitly want to override this setting to have no value, disregard what my parent sets.

Is that relevant to `Builder`? Probably not, but it's nice if APIs are as clear as can be, and that is my first thought when I see `clearName` for a non-collection property.

Said differently, what is the difference between:

```java
builder.name(null);
 // -AND-
builder.clearName();
```

Possibly: The first one throws an exception, the second one doesn't. This is not a satisfactory answer to this dilemma. COBOL died for a reason - libraries _generally_ (this rule has limits) should be written to be flexible, and the language (and lombok acts more like 'programming language' than 'library') should incentivise this. In other words, 'throws an exception' means if you write any intermediate glue, that would have a lot of this kinda code in it:

```java
String name = findName();
if (name == null) builder.clearName();
else builder.name(name);
```

Instead of just `builder.name(findName())`. And that's bad.

Given that `Optional` doesn't even allow `Optional.of(null)`, this ternary state thing cannot be the answer. The problem is: But it _seems_ like that is a possible answer given just an API of some builder. Knowledge of the intricacies of lombok is not meant to be a requisite for understanding APIs written _with_ lombok.

## Semantic confusion reprise

When you have a property of type `Optional<String>` then __fundamentally__ that is __not__ a String. That is the _point_ of `Optional`. You __must__ accept these semantics, or, alternatively, if you disagree, then we have a new problem: Then `Optional` is conceptually broken. Going with the benefit of the doubt then, these are the semantics:

An `Optional<String>` fundamentally isn't itself a string. It's the concept of 'I am a placeholder; either I am a string and you can ask me for it, or I hold the place of "No value applies / no value has been set / no value has been found / no value is available". In this semantic model, you __want__ to call `.name(Optional.of("John"))` to make this clear.

Let me try to approach this from a different angle:

Imagine you file a JEP with the request that a string should silently autobox to a `Supplier<String>`. In other words, that given: `void foo(Supplier<String> supplier)`, you can just write `foo("Hello")` and that javac compiles this as if you wrote `foo(() -> "Hello")`. And of course, this JEP goes all the way: You can have a `List` that contains a mix of suppliers of strings and strings, and you can just write `for (String s : list)`, with javac automatically generating the code required to call `get()` exactly once on any element that so happens to be a `Supplier`.

Good idea? I hope you said no. Because this is going to be difficult if you feel this is a good change. __It is important not to confuse the semantics of these two things__ - a string is not a supplier of strings, and any attempt to write a library or a language that conflates these ideas is going to cause _vastly_ more problems than it solves.

The _exact same problems_ that such a conflation has, I feel, are problems you get when you conflate a `String` with an `Optional<String>`.

With that in mind, I think I found the problem: Being annoyed at having to write `.name(Optional.of("John"))`? __That__ is the problem. You shouldn't be annoyed at that. That's what your class property _is_.

I do accept I might hold a minority opinion on that one, but, I have very little room for PRs like this given this stance.

## Alternatives: 1 - overload

Workable alternative 1 is to simply overload optional with its own type. Instead of generating only:

```java
public BuilderType name(Optional<String> name) {
  this.name = name;
  return this;
}
```

We generate:

```java
public BuilderType name(@NonNull Optional<String> name) {
  if (name == null) throw new NullPointerException("name");
  this.name = name;
  return this;
}

public BuilderType name(String name) {
  this.name = Optional.ofNullable(name);
  return this;
}
```

This would then allow you to 'clear' the value either by calling `.name(Optional.none())` or even `.name((String) null)`.

It's backwards incompatible in only 2 ways:

1. Currently if you don't set any value for your optional, your `Optional<String>` field would contain `null`, whereas in the new model it would contain `Optional.none()`. While backwards incompatible and we're away of [XKCD 1172](https://xkcd.com/1172/), I have no problem telling anybody that complains about this to take a hike.

2. Before you could call `name(null)`, which then sets your optional field __explicitly to `null`__ which is weird. After this update, the compiler would complain that your method call is now ambiguous. See previous bullet: Don't care.

## Alternative 2: Just silently 'fix' optionality

instead of generating this:

```java
public BuilderType name(Optional<String> name) {
  this.name = name;
  return this;
}
```

We generate:

```java
public BuilderType name(String name) {
  this.name = Optional.ofNullable(name);
  return this;
}
```

This is much more backwards incompatible. Essentially all uses of `@Builder` on anything with a field of type `Optional` - we break. It also means you `null` allergic folk __have to__ write `.name(null)` if you want to clear the name.

## Alternative 2b

As a slight update to the above, we could add:

```java
public BuilderType clearName() {
  this.name = Optional.none();
  return this;
}
```

Because we love you and we know your computer explodes when you type the sequence `n`, `u`, `l`, `l` in your code editor.

## Concerns

Any proposal or feature needs to take into account:

* Backwards compatibility
* `@SuperBuilder`

I'm going to close this issue. If you have some basic feedback just drop it as a comment here. If you have a serious proposal that addresses the above concerns, I think the best route is to file a new issue (not a PR) that refers to this one and which lays out the proposal.

PR #4006, opened 2026-01-26
#3974 refactored to ShadowClassLoader use Collections::enumeration instead of using Vector

### LINKED ISSUE #3974: `ShadowClassLoader​::getResources(…)` can use `Collections​::enumeration(…)` instead of `java.util.Vector`
The static [`java.util.Collections​::enumeration(…)`][java.util.Collections::enumeration(java.util.Collection)] method allows getting a [`java.util.Enumeration`] from an arbitrary `Collection`, and can be used to avoid the legacy `java.util.Vector` class. This method dates back to **Java SE 1.2**, so it’s supported in every **Java** version supported by **Lombok**.

[java.util.Collections::enumeration(java.util.Collection)]: https://docs.oracle.com/en/java/javase/25/docs/api/java.base/java/util/Collections.html#enumeration(java.util.Collection)
[`java.util.Enumeration`]: https://docs.oracle.com/en/java/javase/25/docs/api/java.base/java/util/Enumeration.html

skysrd:
Fixes #3974 modified the code to use Collections::enumeration instead of Java.util.Vector

### DIFF
--- a/src/launch/lombok/launch/ShadowClassLoader.java
+++ b/src/launch/lombok/launch/ShadowClassLoader.java
@@ -43,7 +43,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Set;
-import java.util.Vector;
 import java.util.WeakHashMap;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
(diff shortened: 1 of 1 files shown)


rzwitserloot:
I really, really thought `Collections::enumeration` was too new to use in lombok, but apparently it's old enough to buy alcohol (well, in the cultured climes of Europe, not the colonies). Neat! Thanks for the PR :)

This has been merged - github just doesn't know how git works.

PR #4009, opened 2026-02-09
Add support for installing into IBM Rational Developer for i

fallingrock:
IBM Rational Developer for i (or RDi) is a Eclipse based IDE for RPG & COBOL on IBM i (formerly AS/400, System i, & iSeries) systems.

Although it's primarily intended for RPG & COBOL development, it does include the complete Java dev tools features.

I found it frustrating that lombok's stand alone installer couldn't find or properly update RDi, so I added the necessary support.


rzwitserloot:
There's a logo png in this commit. Searching for "ibm rational logo" does not find that logo anywhere. Reverse image searching this logo does not result in any hit for 'IBM' or 'Rational'.

I don't mind adding support for this (I'm going to say 'lombok supports COBOL!' and cackle like a madman, because, funny), but that's a bit of a showstopper. Can you clarify the logo situation? Admittedly IBM is not doing a great job making clear what its logo even is. If this really is the logo that the RDP app itself has, ookay. If you testify to that, I guess we'll install it and have a look. If not - that's the best logo to use here.

PR #4010, opened 2026-02-12
Feat: 'Default' `@ExtensionMethod` extensions

screret:
# Description

This PR introduces the `lombok.extensionMethod.defaultExtensions` and `lombok.extensionMethod.suppressBaseMethods` configuration keys, allowing extension method usage without explicitly applying `@ExtensionMethod({MyExtensions.class, MyExtensions2.class, ...})` to every class via lombok.config.
Using the annotation still requires explicitly defining an extension, but it can be left empty if one (for whatever reason) only wants to change the `suppressBaseMethods` value for a class.

If `suppressBaseMethods` isn't defined, the configuration value is applied (default: true).
Extensions in the `lombok.extensionMethod.suppressBaseMethods` configuration are applied _in addition to_ explicitly defined ones. As far as I can tell, defining an extension twice doesn't have any adverse effects.

# Changes
- Added EXTENSION_METHOD_SUPPRESS_BASE_METHODS key in ConfigurationKeys.java.
- Added EXTENSION_METHOD_DEFAULT_EXTENSIONS key in ConfigurationKeys.java.
- Made `HandleExtensionMethod` an AST adapter.

# Additional Information
The code is loosely based on `HandleFieldDefaults`.
I did _not_ fix the Eclipse handler, because I don't know how, and I assume you know better than I do.

Getting the project to compile was surprisingly annoying, though that might just be me not wanting to use Eclipse. IntelliJ wasn't happy with the java 6 requirement and duplicate dependency classes, but I got it working.
Sometimes I had to delete a stub class, recompile, and add the stub back. That made it mysteriously work 10/10 times. It worked, though, so I'm not complaining.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -65,5 +65,6 @@ Till Brychcy <till.brychcy@mercateo.com>
 Victor Williams Stafusa da Silva <victorwssilva@gmail.com>
 Yonatan Sherwin <yonatansherwin@gmail.com>
 Yun Zhi Lin <yun@yunspace.com>
+Minttu Stenberg <screret@screret.dev>
 
 By adding your name to this list, you grant full and irrevocable copyright and patent indemnity to Project Lombok and all use of Project Lombok in relation to all commits you add to Project Lombok, and you certify that you have the right to do so.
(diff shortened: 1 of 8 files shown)


nealeu:
This is definitely going to need tests.

The `lombok.config` settings can be varied in the `before` fixture as done in `NonNullWithAssertion`.

I'd also be interested in what performance impact there could potentially be having something apply by default.

Rawi01:
I implemented something similar years ago: https://github.com/Rawi01/lombok/commit/516e0c774a0c5ea0d25685658a1d9c51b641a761

If I remember correctly it slows down compilation and @rzwitserloot was not happy about that.

nealeu:
> I implemented something similar years ago: [Rawi01@516e0c7](https://github.com/Rawi01/lombok/commit/516e0c774a0c5ea0d25685658a1d9c51b641a761)
> 
> If I remember correctly it slows down compilation and @rzwitserloot was not happy about that.

This is what I would suspect.  Doing extension functions is probably way more efficient as part of the compile stage rather than as an annotation processing step.

PR #4012, opened 2026-02-22
[ci] GitHub Actions workflow to create Github Releases from git tags

timtebeek:
Hi! This a quick request to make it easier to track new releases right from GitHub. I'm a fan of the "Watch > Custom > Releases" option, and I imagine I'm not the only one.

image

This workflow would run whenever a tag is pushed to create a matching release on this page of the repository:

- https://github.com/projectlombok/lombok/releases

No further changes to the release process needed; this just runs on tags and uses a built in support from GitHub tokens & CLI.

No harm of course if this is not to your liking; it was just as much effort to open a PR as an issue would have been, so feel free to adopt or close.

To see it in action: I'd previously added the same in
- https://github.com/apache/camel-upgrade-recipes/pull/53
- https://github.com/apache/camel-upgrade-recipes/releases

### DIFF
--- a/.github/workflows/github-release.yml
+++ b/.github/workflows/github-release.yml
@@ -1,0 +1,21 @@
+name: Create GitHub Release
+# This workflow creates a GitHub release when a tag is pushed, such that folks can subscribe to releases in GitHub
+
+on:
+  push:
+    tags:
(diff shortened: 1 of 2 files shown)


rspilker:
We're moving away from GitHub specific features, most notably all Actions.

PR #4013, opened 2026-02-22
Add `@lombok.CheckReturnValue` to `@With` and `@Builder build()` methods for static analysis

### LINKED ISSUE #3984: [FEATURE] Add a `@CheckReturnValue` annotation on generated methods where applicable
**Describe the feature**
Add a `@CheckReturnValue` annotation on generated methods where applicable, to signify the result should be used. The goal is to avoid folks mistakenly calling a method and ignoring the return, for instance on a `@With` method that returns a new instance.

Example taken from: https://projectlombok.org/features/With
## Before
```java
import lombok.AccessLevel;
import lombok.NonNull;
import lombok.With;

public class WithExample {
  @With(AccessLevel.PROTECTED) @NonNull private final String name;
  @With private final int age;
  
  public WithExample(@NonNull String name, int age) {
    this.name = name;
    this.age = age;
  }
}
```
## After
```java
Import lombok.CheckReturnValue;
import lombok.NonNull;

public class WithExample {
  private @NonNull final String name;
  private final int age;

  public WithExample(String name, int age) {
    if (name == null) throw new NullPointerException();
    this.name = name;
    this.age = age;
  }

  @CheckReturnValue // New!
  protected WithExample withName(@NonNull String name) {
    if (name == null) throw new java.lang.NullPointerException("name");
    return this.name == name ? this : new WithExample(name, age);
  }

  @CheckReturnValue
  public WithExample withAge(int age) {
    return this.age == age ? this : new WithExample(name, age);
  }
}
```

Developers using annotations generated from `@With`, `@Builder` `build()` or even `@Getter` and `@ToString`. I think the most value would be there for `@With` those, and propose to start (only) there.

**Additional context**
There is no shortage of CheckReturnValue annotations; since JSR-305 failed folks seem to just define their own, with tooling picking up based on the simple class name. As an example there’s all of these:

1. https://github.com/JetBrains/java-annotations/blob/master/src/jvmMain/java/org/jetbrains/annotations/CheckReturnValue.java - `@Retention(RetentionPolicy.CLASS)` (implied)
2. https://github.com/google/error-prone/blob/master/annotations/src/main/java/com/google/errorprone/annotations/CheckReturnValue.java - `@Retention(RetentionPolicy.RUNTIME)`
3. https://github.com/findbugsproject/findbugs/blob/master/findbugs/src/java/edu/umd/cs/findbugs/annotations/CheckReturnValue.java - `@Retention(RetentionPolicy.CLASS)`
4. https://github.com/qos-ch/slf4j/blob/master/slf4j-api/src/main/java/org/slf4j/helpers/CheckReturnValue.java - `@Retention(RetentionPolicy.RUNTIME)`
5. https://github.com/spotbugs/spotbugs/blob/master/spotbugs-annotations/src/main/java/edu/umd/cs/findbugs/annotations/CheckReturnValue.java - `@Retention(RetentionPolicy.CLASS)`
6. https://github.com/assertj/assertj/blob/main/assertj-core/src/main/java/org/assertj/core/annotation/CheckReturnValue.java - `@Retention(RetentionPolicy.CLASS)` (implied)
7. https://github.com/jOOQ/jOOQ/blob/main/jOOQ/src/main/java/org/jooq/CheckReturnValue.java - `@Retention(RetentionPolicy.CLASS)`

It might be complicated to determine which of these are available to use, or which to use when multiple are available. Lombok does already provide a `lombok.Generated` annotation though, so perhaps there's the option for a `lombok.CheckReturnValue` annotation too?

As an alternative there's also the `@Contract(pure = true)` option, as potentially hinted at here:
- https://github.com/projectlombok/lombok/issues/1109

timtebeek:
As briefly discussed at JFall.

- Fixes https://github.com/projectlombok/lombok/issues/3984


rzwitserloot:
The concept is approved, but we want to release a JDK26 compatible version first.

A few changes to the design:

1. We want to kick this off with, as default, _not_ to generate it. For now, you have to opt in to get em. This way we can paper over a few versions: `lombok.CheckReturnValue` will exist the moment this PR goes live into stable, and we'll then wait a year or two and change the default - that gives a 'window' of two years of lombok releases where the annotation exists but wasn't by default on yet.
2. Whilst clearly it is _very_ trendy to make your own `CheckReturnValue` annotation, and we will, it'd be nice if we can accomodate those who have their own fashion sense: A `lombok.config` setting to tell lombok to use some different common `@CheckReturnValue`.

We can solve both of those in one swoop with a `lombok.config` setting (enum) with the following values:

1. _unset_ - take the lombok default, which for now is __none__ but will later be __lombok__.
2. `none` - do not generate any annotation.
3.  `lombok` - generate the `@lombok.CheckReturnValue` annotation.

Later on we can add more values, such as `jetbrains` or `errorprone` or what not, or even `custom: com.foo.FullyQualifiedNameHere`, or as a separate lombok config key. It leaves doors open.

timtebeek:
Great to hear and thanks for [the new release](https://github.com/projectlombok/lombok/releases/tag/v1.18.46); let me know if you'd want me to make any changes, or if you'd prefer to do so yourself.

timtebeek:
I've made the above changes on my branch, as well as merged `master` to update the doc/changelog.markdown; let me know if there's anything else I can do to help this along!

rzwitserloot:
@timtebeek LGTM, but you haven't added your name to the AUTHORS file. Can you throw one more commit in here with that update?

timtebeek:
Done, thanks! Looking forward to using this in our analyses.

rzwitserloot:
This got complicated in further review and integration!

* The config key model is not semantically correct. Instead of flavour NONE and flavour LOMBOK, it's a list of flavors. It's semantically reasonable to want to generate _both_ lombok flavor, as well as android flavor, as well as checkerframework flavor, for example. It's like JACKSON_VERSION: Usually you just want 1 verison but if you're a library trying to stay stable during a major upgrade process, you probably want both.
* We _already do this_ - for the CheckerFramework variant, which has gone down a different semantic road: CheckerFramework doesn't describe this notion as 'check return value'; it describes it as 'side effect free'. This then _implies_ that invoking the method and ignoring what it returns is therefore a code error. But, the `lombok.config` setting for this is a general 'please apply all relevant logic of my checkerframework version', not on a per-feature basis. So that makes integration complicated.
* __There is a bug in our existing implementation__ - and you made the same mistake, so I guess you and we are in each other's good company! The `@Builder` annotation can be applied to scenarios where the effect of 'executing' the builder invokes a method, and _that method may have side effects_ - in other words, it is not necessarily true that ignoring the returned value of a call to `build()` is a bug at all!

The 3 situations (one of which splits into 2 alternate roads, so 4 total) you can use `@Builder` are:

* Case A: You annotate a type with `@Builder`. This gets you a free implied `@AllArgsConstructor`, and for this case, that means lombok _will_ generate this constructor.
* Case B: Like A, except, you wrote out the all-args-constructor yourself, therefore the implied `@AllArgsConstructor` doesn't actually generate anything.
* Case C: You annotate an explicitly written out constructor.
* Case D: You annotate a method.

In _every case_ except A, it's possible that side effects occur. I think we can all agree that writing a constructor in such a way that it is plausible and possibly sensible to invoke the constructor and then ignore the return type, is __terrible__ java code, but it goes a bit far to just assume that _therefore_ nobody would be stupid enough to try it.

Nevertheless, that's how our _already existing_ implementation (for CheckerFramework's `@SideEffectFree`) already works: We slap that annotation on `build()`, regardless of where this builder came from. We should be copying over any `@SEF` method from the constructor/method that we're building for (and `@AllArgsConstructor`) should slap `@SEF` onto the constructor it makes.

Roel and I are talking about how to proceed.

rzwitserloot:
Okay, some conclusions:

1. We now generate `@SideEffectFree` on any constructors we make.
2. Lombok hard-codes the notion that constructors are inherently 'check return type'. Even though this is not, technically, true (a constructor can have side effects. They can run the whole app, even). checkerframework follows the theory: In checkerframework parlance, a constructor can be, but does not have to be, `@SideEffectFree`. There is even an example situation where side-effect-full constructors (i.e. ones where if you just call them and disregard the value of the expression is fine) are common: The building of two-way traversible hierarchies; the step to make an object is often `new ChildComponent(parent)` in order to enforce that post-construction you can't have an unattached child. UI frameworks do this sometimes. SWT, in particular, works this way. Nevertheless, lombok doesn't cater to them. In any way: We are not going to make a 'negative' annotation (no `@ReturnValueMayBeIgnored`). The workaround is to make a static method that does this, because...
3. When `@Builder` is applied to a method (static or instance), lombok _does not_ assume that this method is therefore 'check return type'... unless it has an annotation: This PR's own `@CheckReturnType`, android's take, or checkerframework's take will be recognized out of the box. This does not count if the `@Builder`-annotated method returns `void`. This is a breaking change, but probably for nobody (we no longer slap checkerframework `@SideEffectFree` on em, but then what SEF enables is disabled by default in checkerframework). It's a bug fix, so doesn't count as intentional backwards incompatibility, XKCD space bar heater haters, do you worst!

Which means:

1. Update for CheckerFramework users: We now slap `@SEF` on generated constructors. This is unrelated to this feature.
2. Bugfix: We no longer slap `@SEF` on `build()` if the builder is generated due to putting `@Builder` on a method. We still generate `@SEF` if you put `@Builder` on a constructor.
3. lombok config: The config key is a list, with 2 flavours (lombok, and android). CheckerFramework is to be applied via the `checkerframework` config key (using CF is an all-or-nothing situation, you can't pick and choose).
4. Your code has been updated with the same logic as item 2: No `@CheckReturnType` if the builder is due to `@Builder`-on-a-method.
5. .. but we _do_ annotate with `@CRT` if using `@Builder`-on-a-method if that method has any of the known SEF/CRT annotations.. unless the method is void, in which case we will generate `@SEF` but not `@CRT`.
6. BIg update to the config system: For listy things you can now just use `=` which is shorthand for 'clear the list then add this one thing', because in the vast majority of case, you'd only want us to generate _one_ flavor.

rzwitserloot:
7. `@lombok.CheckReturnValue` also applies to getters, hashCode, equals...

rzwitserloot:
[Lean Branching](https://github.com/fork-dev/TrackerWin/issues/1432) is great and github is too dumb to know how it works, but, this branch has been merged, great job @timtebeek !

timtebeek:
Thanks a lot for the thoughtful (and late night!) review & merge. Great to see this added, and I look forward to adding it downstream, to figure out where the return value might have been lost.

PR #4016, opened 2026-02-23
500ns per annotation improvement in checking if annotations are NonNull

### LINKED ISSUE #4017: [BUG] Poor performance searching alphabetically ordered lists to find NonNull annotation matches
**Describe the bug**
As demonstrated in #4016  the implementation for checking if NonNull annotations are present (e.g. when generating constructors) is inefficient. 

- The time taken to find a match depends on the order of the array being search, with android NonNull annotations performing best.
- The method implementing the match does work that does not apply to this list of annotations (i.e. looking for `$` within the FQCN) and the fix appears to be the correct lookup.

**To Reproduce**
Run the `benchmark.compile` and `benchmark` ant tasks from branch https://github.com/SolumXplain/lombok/tree/refs/heads/hasNonNullOptimize-jmh against the last and last but one commits of the PR branch.

**Expected behavior**
There should not be a significant difference between matching different NonNull annotations.

 - Lombok version: 1.18.30
 - Platform: all - PoC just demonstrates javac fix for now

nealeu:
See #4017 for more information.

Local benchmarking on a modern 5GHz CPU results in approx 500ns improvement for each annotation that is checked that isn't a match.  For matches then the benefit depends on how far down the list of non-null annotations it is, but as demonstrated, the Android one that is first in the list also improves.

Projects using Spring, JPA or Swagger will probably find this notably improves build and rebuild times due to their heavy use of annotations.

```
Benchmark		Before (ns/op)	After (ns/op)	Improvement
noAnnotations		0.695			0.900		~same
androidNonNullFirst	127.2			112.3		1.1×
singleNonMatching	597.0			92.7		6.4× faster
singleNonNull		650.8			70.5		9.2× faster
manyNoMatch			1911.8			275.1		6.9× faster
manyLastMatch		2367.2			349.3		6.8× faster
```

resolves #4017

### DIFF
--- a/src/core/lombok/core/handlers/HandlerUtil.java
+++ b/src/core/lombok/core/handlers/HandlerUtil.java
@@ -45,6 +45,7 @@ import lombok.core.AST;
 import lombok.core.AnnotationValues;
 import lombok.core.JavaIdentifiers;
 import lombok.core.LombokNode;
+import lombok.core.TypeLibrary;
 import lombok.core.configuration.AllowHelper;
 import lombok.core.configuration.CapitalizationStrategy;
 import lombok.core.configuration.ConfigurationKey;
@@ -80,6 +81,7 @@ public class HandlerUtil {
 	}
 	
 	public static final List<String> NONNULL_ANNOTATIONS, BASE_COPYABLE_ANNOTATIONS, JACKSON_COPY_TO_GETTER_ANNOTATIONS, JACKSON_COPY_TO_SETTER_ANNOTATIONS, JACKSON_COPY_TO_BUILDER_SINGULAR_SETTER_ANNOTATIONS, JACKSON_COPY_TO_BUILDER_ANNOTATIONS;
+	public static final TypeLibrary NONNULL_TYPE_LIBRARY;
 	static {
 		// This is a list of annotations with a __highly specific meaning__: All annotations in this list indicate that passing null for the relevant item is __never__ acceptable, regardless of settings or circumstance.
 		// In other words, things like 'this models a database table, and the db table column has a nonnull constraint', or 'this represents a web form, and if this is null, the form is invalid' __do not count__ and should not be in this list;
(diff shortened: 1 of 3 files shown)


Rawi01:
Have you tried if this adds up to a reasonable amount in a real world project? I had no time to try it yet. If this really helps we could extend it to the other annotation lists in `HandlerUtil` and I also think that we could add a new constructor to the type library to create it without building a temporary array.

nealeu:
I've not benchmarked overall effect in a large project. The one I'm currently working on has over 30,000 annotations, so it's 15ms, so I wouldn't expect this single PR to be noticed on a single build.  It just looked low hanging fruit for engaging with the team.

Edit: Given just how many projects are using Lombok, many builds a day, then this is not trivial!

rzwitserloot:
> It just looked low hanging fruit for engaging with the team.

I'm sure this is a cultural misunderstanding.

But, to me, this is __unbelievably__ rude. A slap in the face. "Let's send some pointless PR that I freely admit has no significant effect.. it's just to test if the open source maintainers I do not pay and have no support contract with are awake".

nealeu:
I think that goes both ways.  You seem to be deliberately taking the most insulting interpretation of something that wasn't even addressed to you, and my comment above was in no way an insult.

What is incredibly rude is that I have put days of effort in to Lombok, and this one PR that is a slam dunk of gaining back some CPU cycles gets closed unmerged with an insult.  That's so unfair.  I took on a big request around JSpecify, and have put days of effort into looking at what is involved.  And there's silence on [my updates on the issue](https://github.com/projectlombok/lombok/issues/3861#issuecomment-4000487654) and on my [initial feature PR](https://github.com/projectlombok/lombok/issues/3861#issuecomment-4000487654)
While doing that I spotted a potential perf issue, I benchmarked it, and I isolated it into a PR.  I'm new to this community, so putting a simple PR in was a good way to see what is involved.  You have closed the PR without saying why!
Not cool to comment like that rather than feed back what you would like to see to consider merging it if it is so inadequate.
That estimated 15ms for one build for me is a drop in the ocean (literally) because this code will be executed potentially billions of times a day. Yes, it totally sucks that you are not paid appropriately for that, but it doesn't justify the rudeness of closing a PR that you've actually commented on with just an insult!

nealeu:
And relating to professional support: you might want to check your links, because I tried: 

"Looking for professional support of Project Lombok? Lombok is now part of a [tidelift subscription](https://tidelift.com/subscription/pkg/maven-org-projectlombok-lombok?utm_source=maven-org-projectlombok-lombok&utm_medium=referral&campaign=website)!"

Professionalism is less arrogant.

rzwitserloot:
Nealeu wrote:

> doesn't justify the rudeness of closing a PR that you've actually commented on with just an insult!

There is absolutely nothing in my comment that was an insult.

I suggest you take a time out; further snipes and jabs will be removed, and any suggestion that FOSS maintainers owe anybody a review based solely on the fact that you sent in a PR or wrote a comment (or the stars forbid, a merge!), will lead to a ban. FOSS maintainers are not obligated to respond. I will brook no disagreement on this matter.

We closed the PR because it's difficult to process a PR written by somebody we either do not understand, or who has unreasonable demands in trade for the PR if our understanding of the contribution is correct. Difficult, but not impossible. However, going by your own feedback the actual performance impact is negligible which tipped the scale.

nealeu:
Sorry @rzwitserloot, but you are telling me that I'm being "unbelievable rude" for daring to want to engage with you, yet I push back and you tell me that it wasn't an insult.

Well. I guess we engaged and the result is clear -  I should not bother with any bigger PRs.

yes. I agree that FOSS maintainers owe people nothing.  I really don't see the point of *anyone* contributing a PR given what you have said.

PR #4018, opened 2026-02-24
Support JSpecify @NullMarked/@Nullable at class level

nealeu:
This adds support for JSpecify @NullMarked when applied to a class.

As demonstrated in the after-delombok and after-ecj snapshots, the constructor and setter will generate null-check code for anything not marked `@Nullable` where the class has `@NullMarked`.

I'd appreciate some review input on this.  I had considered making the `@NullMarked` annotation configurable so that it could potentially support an internal Lombok class too, and with the ability to choose to only apply that annotation on classes where null checks are wanted. So it could be called `@NullCheck`, for example.

TODO:
- [ ] Make `@RequiredArgsConstructor` support an opt-in so that we can get a release out for IDE plugins that use their own bundled Lombok, and then users can opt in.
  [This](https://github.com/JetBrains/intellij-community/blob/master/plugins/lombok/src/main/java/de/plushnikov/intellij/plugin/processor/clazz/constructor/AbstractConstructorClassProcessor.java#L276) is what would need to change in Jetbrains plugin for this to work. They duplicate Lombok's annotation processor.  
- [ ] Additional test scenarios
- [ ] Documentation: Once we have maintainer buy in
- [ ] Later: Support `@NullMarked` on package-info.java (I think I can see the path but that's a separate PR)

### DIFF
--- a/src/core/lombok/javac/handlers/JavacHandlerUtil.java
+++ b/src/core/lombok/javac/handlers/JavacHandlerUtil.java
@@ -131,7 +131,7 @@ public class JavacHandlerUtil {
 	/**
 	 * Find the parent class or interface
 	 */
-  static JavacNode getParentTypeNode(JavacNode childOfType) {
+	static JavacNode getParentTypeNode(JavacNode childOfType) {
 		JavacNode typeNode = childOfType;
 		while (typeNode != null && typeNode.getKind() != Kind.TYPE) typeNode = typeNode.up();
 		return typeNode;
--- a/src/core/lombok/javac/handlers/JavacHandlerUtil.java
+++ b/src/core/lombok/javac/handlers/JavacHandlerUtil.java
@@ -119,8 +119,38 @@ public class JavacHandlerUtil {
 	private JavacHandlerUtil() {
 		//Prevent instantiation
 	}
-	
-  private static class MarkingScanner extends TreeScanner {
+
+	static boolean hasNullableAnnotations(JavacNode fieldNode) {
+		return hasAnnotation("org.jspecify.annotations.Nullable", fieldNode);
+	}
+
+	static boolean isNullMarked(JavacNode typeNode) {
+		return hasAnnotation("org.jspecify.annotations.NullMarked", typeNode);
+	}
+
+	/**
+	 * Find the parent class or interface
+	 */
+  static JavacNode getParentTypeNode(JavacNode childOfType) {
+		JavacNode typeNode = childOfType;
+		while (typeNode != null && typeNode.getKind() != Kind.TYPE) typeNode = typeNode.up();
+		return typeNode;
+	}
+
+	/**
+	 * Return true if the specified field or parameter node is determined as non-null according
+	 * to JSpecify rules.
+	 * Currently supports @NullMarked on class. To support annotations on package-info.java, we
+	 * would need to place each type node under a package node. Supporting NullMarked on the type
+	 * seems adequate for now, and gives the option opt-in to null assertions on a per class basis
+	 * (i.e. have NullMarked be the default but allow a custom annotation to trigger null checks at
+	 * system boundaries only).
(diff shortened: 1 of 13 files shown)


nealeu:
I've given this a go via a [Jitpack build](https://jitpack.io/#SolumXplain/lombok/5eb0ac9a) (which others could try - just swapping to this for your annotation processor would be fine).

This correctly improves a `@Data` class that is `@NullMarked` by now replacing the no-args constructor with one with the fields that are not `@Nullable` - but IDEs!

For the Gradle based project we're using with IntelliJ, the code analysis for the red-ink, still uses a release version of Lombok, so it expects there to be a no-args constructor.  Gradle however correctly uses the Jitpack build.

For `@RequiredArgsConstructor` we're therefore going to need a feature flag to allow an updated release to be included in IDE plugins and then enable this.

I'll add a checklist of outstanding items to the PR desc.

nealeu:
I've been running a variant of this with a number of feature additions (see this draft PR just for easy viewing https://github.com/SolumXplain/lombok/pull/1/changes?w=1).

This gives us some good static analysis with NullAway for our `@NullMarked` classes and some flexibility to find errors in our null marking by being strict in `assumeNonNull` during tests, but `log.error()` only for those in production.
We also avoid having to go adding `@Nullable` in places where frameworks treat various Spring Data fields as nullable for construction with these additions to `lombok.config`

```properties
lombok.addNullAnnotations=jspecify
lombok.nonNull.checkMethod = com.solum.xplain.shared.common.NullSafety.assumeNonNull
lombok.nonNull.skipAnnotations += jakarta.validation.constraints.NotNull
lombok.nonNull.skipAnnotations += jakarta.validation.constraints.NotEmpty
lombok.nonNull.skipAnnotations += jakarta.validation.constraints.NotBlank
# Id is generated by Spring Data, so we're happy to treat it as non-null apart from on construction
lombok.nonNull.skipAnnotations += org.springframework.data.annotation.Id
```

So, with some engagement from you folks we can pick out what you'd like from this.

The `skipAnnotations` setting is the equivalent of adding to the list of `@Nullable` annotations, so that could be made more unified.

src/core/lombok/javac/handlers/JavacHandlerUtil.java
```diff
@@ -119,22 +119,52 @@ public class JavacHandlerUtil {
...
+	}
+
+	static boolean isNullMarked(JavacNode typeNode) {
+		return hasAnnotation("org.jspecify.annotations.NullMarked", typeNode);
+	}
+
+	/**
+	 * Find the parent class or interface
+	 */
+  static JavacNode getParentTypeNode(JavacNode childOfType) {
```
ExE-Boss:
```suggestion
	static JavacNode getParentTypeNode(JavacNode childOfType) {
```

test/transform/resource/after-delombok/NullMarkedPlain.java
```diff
@@ -0,0 +1,47 @@
...
+	@lombok.Generated
+	public void setS(final String s) {
+		if (s == null) {
+			throw new java.lang.NullPointerException("s is marked non-null but is null");
+		}
+		this.s = s;
+	}
+	@java.lang.SuppressWarnings("all")
+	@lombok.Generated
+	public void setO(@org.jspecify.annotations.Nullable final Object o) {
```
JarvisCraft, line 44: Given that `@Nullable` is a `TYPE_USE` annotations the correct form seems to be:


```suggestion
	public void setO(final @org.jspecify.annotations.Nullable Object o) {
```

nealeu: @JarvisCraft I agree.  That would be a good change to make if this gets adopted by the maintainers.

nealeu:
Closing. Withdrawn.

PR #4024, opened 2026-03-18
Fix Jacksonized stops generating JsonProperty/JsonIgnore after first manual annotation (fluent=true)

### LINKED ISSUE #4022: [BUG]  Jacksonized stops generating JsonProperty after first manual annotation (fluent=true)
**Describe the bug**
`@Jacksonized` stops generating `@JsonProperty` annotations once a property with existing `@JsonProperty` annotation is encountered, when using `@Accessors(fluent = true)` with Jackson 3.

**To Reproduce**

With the configuration `lombok.jacksonized.jacksonVersion += 3`,  the following code:

```Java
@Jacksonized
@Builder
@Getter
@Accessors(fluent = true)
public class Example {

    private String a;
    @JsonProperty("B")
    private String b;
    private String c;
}
```
generates:

```Java
@JsonDeserialize(
    builder = ExampleBuilder.class
)
public class Example {
    @JsonProperty("a")
    private String a;
    @JsonProperty("B")
    private String b;
    private String c;

   //....
}
```

**Expected behavior**
`private String c;` annotated with `@JsonProperty("c")`

 - Lombok version: `1.18.44`
 - Jackson version: `3.1.0`

**Additional context**
Github repository with the example

https://github.com/ralemanc/lombok_bug_Jacksonized_JsonProperty/tree/main

**Possible fix**
https://github.com/ralemanc/lombok/commit/cc025dc49e8fe94b67ec037341cb24aa3dad4795

ralemanc:
This PR fixes #4022

### DIFF
--- a/src/core/lombok/javac/handlers/HandleJacksonized.java
+++ b/src/core/lombok/javac/handlers/HandleJacksonized.java
@@ -116,13 +116,13 @@ public class HandleJacksonized extends JavacAnnotationHandler<Jacksonized> {
 		// Add @JsonIgnore to all transient fields. It will be automatically copied to the getter/setters later.
 		for (JavacNode javacNode : tdNode.down()) {
 			if (javacNode.getKind() == Kind.FIELD) {
-				if (!hasAnnotation(javacNode, JacksonAnnotationType.JSON_PROPERTY2) &&
-					!hasAnnotation(javacNode, JacksonAnnotationType.JSON_IGNORE2)) {
-					if (javacNode.isTransient()) {
-						createJsonIgnoreForField(javacNode, annotationNode);
-					} else {
-						createJsonPropertyForField(javacNode, annotationNode);
-					}
+				if (hasAnnotation(javacNode, JacksonAnnotationType.JSON_PROPERTY2) ||
+					hasAnnotation(javacNode, JacksonAnnotationType.JSON_IGNORE2)) {
+					continue;
+				} else if (javacNode.isTransient()) {
+					createJsonIgnoreForField(javacNode, annotationNode);
+				} else {
+					createJsonPropertyForField(javacNode, annotationNode);
 				}
 			}
 		}
--- a/src/core/lombok/javac/handlers/HandleJacksonized.java
+++ b/src/core/lombok/javac/handlers/HandleJacksonized.java
@@ -119,7 +119,8 @@ public class HandleJacksonized extends JavacAnnotationHandler<Jacksonized> {
 				if (hasAnnotation(javacNode, JacksonAnnotationType.JSON_PROPERTY2) ||
 					hasAnnotation(javacNode, JacksonAnnotationType.JSON_IGNORE2)) {
 					continue;
-				} else if (javacNode.isTransient()) {
+				}
+				if (javacNode.isTransient()) {
 					createJsonIgnoreForField(javacNode, annotationNode);
 				} else {
 					createJsonPropertyForField(javacNode, annotationNode);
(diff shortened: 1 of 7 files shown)


src/core/lombok/javac/handlers/HandleJacksonized.java
```diff
@@ -116,14 +116,13 @@ private void handleJacksonizedAccessors(JavacNode annotationNode, JavacNode anno
...
-					createJsonIgnoreForField(javacNode, annotationNode);
-				} else {
-					createJsonPropertyForField(javacNode, annotationNode);
+				if (!hasAnnotation(javacNode, JacksonAnnotationType.JSON_PROPERTY2) &&
+					!hasAnnotation(javacNode, JacksonAnnotationType.JSON_IGNORE2)) {
+					if (javacNode.isTransient()) {
+						createJsonIgnoreForField(javacNode, annotationNode);
+					} else {
+						createJsonPropertyForField(javacNode, annotationNode);
+					}
```
nealeu: I think `return` -> `continue` would have bean more readable.

However, even with that, is this really the intent.

To match what the comments above say, then this would make sense
```java
	if (javacNode.isTransient()) {
		if (!hasAnnotation(javacNode, JacksonAnnotationType.JSON_IGNORE2)) {
			createJsonIgnoreForField(javacNode, annotationNode);
		}
	} else if (!hasAnnotation(javacNode, JacksonAnnotationType.JSON_PROPERTY2)) {
		createJsonPropertyForField(javacNode, annotationNode);
	}
```

Rawi01: I agree that `continue` is more readable.

Your proposal would add `@JsonProperty` even if the field was already annotated with `@JsonIgnore`.

ralemanc: I also agree with the use of `continue`.

Indeed: 
* If the field is `transient` and annotated with `@JsonProperty`, `@JsonIgnore` will be added.
* If the field is not `transient` and annotated with `@JsonIgnore`, `@JsonProperty` will be added.

**Additional context**
The PR where it was introduced, in case it helps provide some context: https://github.com/projectlombok/lombok/pull/3945

ralemanc:
As a side comment, does it make sense to add `@JsonIgnore` if the field is annotated with `@JsonFormat`?

**Additional context**
https://github.com/FasterXML/jackson-databind/issues/3357

src/core/lombok/javac/handlers/HandleJacksonized.java
nealeu: > I agree that `continue` is more readable.
> 
> Your proposal would add `@JsonProperty` even if the field was already annotated with `@JsonIgnore`.

Ah yes.  Good point.

And how long before Github add a workflow that comes back with "your suggestion caused the following tests to fail", which I hope would have been the case.

ralemanc: Updated the code using `continue.` 

There is a warning in the IDE: `'continue' is unnecessary as the last statement in a loop`.

nealeu: Yes. That's because you can remove the `else` that follows when it's expressed that way.

ralemanc: > Yes. That's because you can remove the `else` that follows when it's expressed that way.

Ups ;)  Right. Code updated.

rzwitserloot:
@ralemanc We've reviewed this PR and it's accepted, but we can't merge this PR until you add your name to the AUTHORS file as part of this PR (for copyright/legal reasons).

ralemanc:
> @ralemanc We've reviewed this PR and it's accepted, but we can't merge this PR until you add your name to the AUTHORS file as part of this PR (for copyright/legal reasons).

@rzwitserloot Done. Thanks!

PR #4027, opened 2026-03-23
Fix: Bump EA_JDK to 27 (25 and 26 have been released)

nealeu:
Attempt to use EA logic for JDK 25 was failing.

Naturally there is work to be done before 26 and 27 can be added to the list of JDKs, but this at least sorts JDK 25.

### DIFF
--- a/.github/workflows/ant.yml
+++ b/.github/workflows/ant.yml
@@ -41,8 +41,8 @@ jobs:
     runs-on: ubuntu-24.04
     needs: build
     env:
-      EA_JDK: 25
-    strategy: 
+      EA_JDK: 27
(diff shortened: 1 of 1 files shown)


rzwitserloot:
@Rawi01 Can you have a look at this / does it make sense? I get that 'EA_JDK: 25' is now a lie, but does 'EA_JDK: 27' do anything useful (is there even an EA 27?) - and why is e.g. 22 and 23 in that long list of JDKs (in the line immediately following the change), and 26 is not?

You're the actions expert :)

rzwitserloot:
merged via cherry-pick.

PR #4030, opened 2026-03-27
[Java 11 baseline] Remove lombok.var - It's illegal in Java 10+

nealeu:
This PR removes lombok.var and related processing and testing.

The target branch should naturally not be main, but a lombok-2.0 branch (which would be an expected future release for a Java baseline of at least Java 11).


rzwitserloot:

PR #4031, opened 2026-03-27
[Java 11 baseline] Remove Java < 9 classes and logic

nealeu:
This removes functionality that supports Java below 9.

I've not yet removed anything from CI and Ant build.xml, but this shows that the Java 11+ builds pass with these changes.

The target branch should naturally not be main, but a lombok-2.0 branch (which would be an expected future release for a Java baseline of at least Java 11).


Rawi01:
Thanks for your effort. Have you discussed these changes with one of the core maintainers? 

Given that lombok still supports Java 6 in 2026 I would be supprised if @rzwitserloot and @rspilker will drop Java 8 support which is still in use in a lot of companies.

nealeu:
See https://groups.google.com/g/project-lombok/c/9Hi9k88u5q0.  I'm sure they'll engage when they next have a session

nealeu:
> Thanks for your effort. Have you discussed these changes with one of the core maintainers?
> 
> Given that lombok still supports Java 6 in 2026 I would be supprised if @rzwitserloot and @rspilker will drop Java 8 support which is still in use in a lot of companies.

The big question here is the path to supporting Java 26 and beyond. There are still parts of the code that support a version of Eclipse that runs on Java 1.4 (for the Eclipse agent) - I've done a PR #4032  for that.

rzwitserloot:

PR #4032, opened 2026-03-30
[Java 1.6 baseline] Remove eclipseAgent transformedSources which supported Java 1.4 source/target

nealeu:
As in related comments, Eclipse support has been Java 8+ for some time.

Provides a nice simplification in the understanding of the build.

This causes no test failures because there aren't any CI tests configured for Eclipse that old.

### DIFF
--- a/buildScripts/compile.ant.xml
+++ b/buildScripts/compile.ant.xml
@@ -1,16 +1,16 @@
 <!--
  Copyright (C) 2020-2023 The Project Lombok Authors.
-  
+
   Permission is hereby granted, free of charge, to any person obtaining a copy
   of this software and associated documentation files (the "Software"), to deal
   in the Software without restriction, including without limitation the rights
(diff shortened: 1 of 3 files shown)


rzwitserloot:

For context, the support for 1.4 class files is primarily about older installations of eclipse that were capable of running on java 1.4 for a very long time, which meant that eclipse itself shipped with 1.4 level class files for a _very long time_ and thus we had to support them. I'm still not sure we can leave it by the wayside: Of every class file in every (transitive) dependency across all jar files across your entire ecosystem (laptop, CI servers, test servers, deployed installations, and so on), how sure are you that they are all of a class version that matches with JDK release 17 and above? I really doubt you can say '100%' with a straight face. Hence, dropping support specifically for old class files is something we do not do lightly.

But, having said that, I do think the time has come to drop all support for all aspects of 1.6 and 1.7. We'll eventually get around to doing that.

PR #4052, opened 2026-07-08
Reject ambiguous Builder class name collisions

### LINKED ISSUE #3857: [BUG] builderClassName="Builder" results in code that compiles in Eclipse, but fails to compile in javac (in an edge case)
**Describe the bug**
If a class uses @Builder(builderClassName = "Builder"), Eclipse compiles the code and it works fine.

When we run a Maven build (Java 21), compile fails 'type lombok.Builder does not take parameters' error.

**To Reproduce**

Take this class:

```
import lombok.Builder;

@Builder(builderClassName = "Builder")
public class MyClass<T extends Object> {
	
	public <E extends Object> Builder<E> toSubBuilder(){
		return new MyClass.Builder<>();
	}
	
}
```

and this POM.xml:

```
<project xmlns="http://maven.apache.org/POM/4.0.0"
	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
	<modelVersion>4.0.0</modelVersion>
	<groupId>com.test</groupId>
	<artifactId>sample</artifactId>
	<version>0.0.1-SNAPSHOT</version>

	<properties>
        <maven.compiler.source>21</maven.compiler.source>
        <maven.compiler.target>21</maven.compiler.target>
		<JUNIT_VERSION>5.11.3</JUNIT_VERSION>
	</properties>



	<dependencies>
		<dependency>
			<groupId>org.projectlombok</groupId>
			<artifactId>lombok</artifactId>
			<version>1.18.34</version>
			<scope>provided</scope>
		</dependency>

	</dependencies>
</project>
```

Load in Eclipse and note that it shows no compile errors.  (Note: sometimes, I do see compile errors in Eclipse as I've been working on this, but doing a Maven->Update Project always clears them.)

Now do a Maven clean compile.

The compiler generates the following errors:

```
[ERROR] Failed to execute goal [32morg.apache.maven.plugins:maven-compiler-plugin:3.13.0:compile[m [1m(default-compile)[m on project [36msample[m: [1;31mCompilation failure[m: Compilation failure: 
[ERROR] /...../src/main/java/my/package/MyClass.java:[8,42] type lombok.Builder does not take parameters
[ERROR] /...../src/main/java/my/package/MyClass.java:[9,35] cannot find symbol
[ERROR]   symbol:   class Builder
[ERROR]   location: class my.package.MyClass
```

If I remove the bulderClassName parameter, everything works as expected.

**Expected behavior**
I expect the Eclipse compile and the Maven compile to both fail or both succeed.  I'm assuming b/c the javac compile is failing, that the Eclipse compile probably should be failing also.

 - Lombok version - 1.18.34
 - Platform:  Java 21.0.5.  Eclipse 2024-12 (4.34.0)
 - Lombok 1.18.34 "Envious Ferret" is installed in Eclipse

**Additional context**

For my initial troubleshooting, and recommendation from Jan Rieke to open a Lombok issue, see https://stackoverflow.com/questions/79540332/maven-with-lombok-build-failing-java-21/79542214#79542214

snowykte0426:
## Summary
- reject simple-name `@Builder` when `builderClassName` resolves to `Builder`, suggesting `@lombok.Builder` instead
- apply the same simple-name collision rule to `@SuperBuilder` / `SuperBuilder`
- add javac and ECJ transform fixtures for rejected simple annotations and allowed fully-qualified annotations

Fixes #3857.

## Tests
- `JAVA_HOME=$(/usr/libexec/java_home -v 21) ant -noinput test.javacCurrent`
- `git diff --check`
- `JAVA_HOME=$(/usr/libexec/java_home -v 21) ant -noinput dist test.eclipse-202503` confirms the new Builder/SuperBuilder cases pass; the target still fails later on existing `ecj-ValUndenotable.java` missing expected output.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -59,6 +59,7 @@ Rostislav Krasny <45571812+rosti-il@users.noreply.github.com>
 Samuel Pereira <samuel.p.araujo@gmail.com>
 Sasha Koning <askoning@gmail.com>
 Szymon Pacanowski <spacanowski@gmail.com>
+Taeeun Kim <snowykte0426@naver.com>
 Taiki Sugawara <buzz.taiki@gmail.com>
 Takuya Murakami <tmurakam@tmurakam.org>
 Thomas Darimont <thomas.darimont@gmail.com>
--- a/src/core/lombok/eclipse/handlers/HandleBuilder.java
+++ b/src/core/lombok/eclipse/handlers/HandleBuilder.java
@@ -114,14 +114,25 @@ public class HandleBuilder extends EclipseAnnotationHandler<Builder> {
 	static final char[] BUILDER_TEMP_VAR = {'b', 'u', 'i', 'l', 'd', 'e', 'r'};
 	static final AbstractMethodDeclaration[] EMPTY_METHODS = {};
 	static final String TO_BUILDER_NOT_SUPPORTED = "@Builder(toBuilder=true) is only supported if you return your own type.";
-	
+	static final String BUILDER_CLASS_NAME_ANNOTATION_CONFLICT = "builderClassName cannot be \"%s\" when using @%s; use @%s or choose another builder class name.";
+
 	private static final boolean toBoolean(Object expr, boolean defaultValue) {
 		if (expr == null) return defaultValue;
 		if (expr instanceof FalseLiteral) return false;
 		if (expr instanceof TrueLiteral) return true;
 		return ((Boolean) expr).booleanValue();
 	}
-	
+
+	static boolean checkBuilderClassNameAnnotationConflict(String annotationName, String qualifiedAnnotationName, String builderClassName, Annotation ast, EclipseNode annotationNode) {
+		char[][] typeName = ast.type == null ? null : ast.type.getTypeName();
+		if (annotationName.equals(builderClassName) && (typeName == null || typeName.length == 1)) {
+			annotationNode.addError(String.format(BUILDER_CLASS_NAME_ANNOTATION_CONFLICT, builderClassName, annotationName, qualifiedAnnotationName));
+			return false;
+		}
+
+		return true;
+	}
+
 	static class BuilderJob {
 		CheckerFrameworkVersion checkerFramework;
 		EclipseNode parentType;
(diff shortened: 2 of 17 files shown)


rspilker:
@snowykte0426 Thanks for this contribution as well. Can you add your name to the [AUTHORS](https://github.com/projectlombok/lombok/blob/master/AUTHORS) file, so we are legally covered?

In the meantime, we've promoted `SuperBuilder` out of experimental. Can you replace your references to `lombok.experimental.SuperBuilder` by `lombok.SuperBuilder`. `experimental.SuperBuilder` is not yet deprecated, but it will be in an upcoming release.

snowykte0426:
Thanks, @rspilker! Added myself to AUTHORS.

I also rebased onto current `master` and replaced the remaining `lombok.experimental.SuperBuilder` references — the qualified name in the error message (javac + eclipse) and the test resources — with `lombok.SuperBuilder`. `ant test` passes.

rzwitserloot:
Branch integrated (github frontend does not understand [Lean Branching](https://github.com/fork-dev/TrackerWin/issues/1432) but, it's merged!)

SledgeHammer01:
This change really makes the code look ugly with `@lombok.builder` everywhere tbh. I didn't run into the issue in either IntelliJ or a Jenkins workflow. My team decided to downgrade back to the previous version. Yup, its THAT ugly :). As is the alternative of `MyClass.MyClassBuilder`. Like what else would you call a builder class besides `Builder`? I'd rename the annotation before making us put `@lombok.Builder` everywhere.

rspilker:
@SledgeHammer01 Are you suggestion to rename the `@Builder` annotation? We could rename it to `@LombokBuilder`, but `@lombok.Builder` is a proper solution.

SledgeHammer01:
@rspilker I understand `@lombok.Builder` is the "proper" workaround, it just looks ugly and out of place compared to every other annotation:

```
@Getter
@Setter
@ToString
@lombok.Builder
```

I personally have never come across a single other annotation that requires a qualifier. You probably have more data then I do, but I would suspect many people have been renaming the builder class to "Builder" as "Actor.ActorBuilder" is overly verbose compared to "Actor.Builder".

`@LombokBuilder` definitely looks a million times better and is more in line with all other Java and Spring annotations.

If putting in the block was just intended to solve an "edge case", then maybe a warning would be more appropriate? I have seen a warning in my build logs when I use `@SuperBuilder`, I think related to default values or something.

Just my two cents 😊, but other devs on my team also didn't want to put `@lombok.Builder` every where :). Also, `@LombokBuilder` is one character shorter as a bonus.

But like I said, if it was just intended to prevent an edge case, I'd probably stick with `@Builder` and a warning. Although, the build not working would be a warning too, but it would at least explain what the issue is.

SledgeHammer01:
@rspilker actually, now that I think about it, `@LombokBuilder` is also kind of out of place since we don't have `@LombokGetter`, etc., so I'd say, for me personally:

#1 keep `@Builder` with added warning since its been working this way since day one
#1.5 `@Builder` with error, but have lombok.config flag to allow it for people who have been using it fine for years
(distant) #2 `@LombokBuilder`

#1 is preferable since the other 2 choices are breaking changes to existing code bases (as is `@lombok.Builder`). Although option 1.5 would only need a change in one place.

PR #4053, opened 2026-07-08
Document Ant annotation processing for JDK 23

### LINKED ISSUE #3909: [BUG] Compilation issues with GraalVM 24 CE (GraalVM 21 CE works)
**Describe the bug**
I upgraded my build environment from GraalVM 21 CE to GraalVM 24 CE. With 21 my IVY and ANT based build script worked as intended, but with 24 the compile step does not work anymore:

```
[javac] Compiling 484 source files to /opt/atlassian/pipelines/agent/build/ant-build
    [javac] /opt/atlassian/pipelines/agent/build/src/com/pathmate/pmcp/FirstInitializations.java:31: error: incompatible types: String cannot be converted to val
    [javac] 				.toLowerCase().replaceFirst("^/", "");
    [javac] 				                           ^
    [javac] /opt/atlassian/pipelines/agent/build/src/com/pathmate/pmcp/FirstInitializations.java:33: error: incompatible types: String cannot be converted to val
    [javac] 		val configurationsFileString = System.getProperty(contextPath
    [javac] 		                                                 ^
```

A lot of these errors appear as if lombok is no longer "working/active".

My ant compilation looks like this:

```
	<!-- Compile -->
	<target name="compile" depends="resolve">
		<javac srcdir="src" destdir="ant-build" debug="off" nowarn="on" includeantruntime="false" source="15" target="15">
			<classpath refid="classpath" />
		</javac>
...
```

The classpath of course contains lombok via ivy:

```
		<!-- Lombok -->
		<dependency org="org.projectlombok" name="lombok"
			rev="1.18.38" />
```

I can also find the lombok JAR in the used lib folder while the script is running.

Any ideas?

**To Reproduce**
Complex - but of course I will provide a test environment if no instant ideas how to solve it come up.

**Expected behavior**
It compiles with GraalVM 24 CE the same way as it does with GraalVM 21 CE. :-)

 - Lombok version: 1.18.38
 - Platform: ant started from eclipse on Mac, but **same** behavior also in Bitbucket Pipelines with Docker based Ubuntu

**Additional context**
As you can see in the version information: I can reproduce it locally on the Mac as well as in Docker Ubuntu on Pipelines... the result is the same, the only change is the switch from the older to the newer GraalVM.

snowykte0426:
## Summary
- update the Ant setup page to distinguish classpath-only non-modular builds from JDK 23+ annotation processing requirements
- add a `-processorpath` Ant `compilerarg` example for local lombok jars
- mention the same processor path / `-proc:full` requirement for Ivy-resolved lombok jars

Fixes #3909.

## Tests
- `git diff --check`
- `JAVA_HOME=$(/usr/libexec/java_home -v 21) ant -noinput website.build`

### DIFF
--- a/website/templates/setup/ant.html
+++ b/website/templates/setup/ant.html
@@ -5,7 +5,7 @@
 		<p>
 			This page explains how to compile your code when you use the <a href="https://ant.apache.org/">Apache Ant</a> build tool. We suggest you use <a href="http://ant.apache.org/ivy/">ivy</a>, the ant add-on that lets you fetch dependencies from the internet automatically.
 		</p><p>
-			Lombok just needs to be on the classpath when you compile your code to do its work, so all you have to ensure, is that lombok is on the classpath in your <code>&lt;javac&gt;</code> task.
+			Lombok needs to be available when you compile your code to do its work. For non-modular builds on JDK22 and below, it is enough to put lombok on the classpath in your <code>&lt;javac&gt;</code> task. On JDK23 and above, you also need to enable annotation processing explicitly.
 		</p>
 	</@s.introduction>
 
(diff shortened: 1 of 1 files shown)


rspilker:
Thanks for the contribution.

Issue #4056, opened 2026-07-24
[FEATURE] Supports custom toBuilder method name


bengbengbalabalabeng:
**Describe the feature**

Currently, when using `@Builder(toBuilder = true)`, Lombok automatically generates an instance method named `toBuilder()`. However, there is no way to customize this method's name. This limitation introduces several issues:

1. **Domain/API Naming Conventions**: Some codebases or frameworks prefer different terminologies for creating a builder from an existing instance, such as `mutate()`, `rebuild()`, `modify()`, or `copy()`.
2. **API Inconsistency**: Lombok's `@Builder` already allows customizing other generated method names via `builderMethodName` and `buildMethodName`. Not being able to customize the "to-builder" method name feels like an API inconsistency.

We propose adding a new attribute `toBuilderMethodName` to `@Builder` (And `@SuperBuilder` if needs).

API Usage:

```java
// with lombok
@Builder(
    toBuilder = true,
    // default: "toBuilder"
    toBuilderMethodName = "mutate"
)
public class Person {
    private String name;
    private Integer age;
}

// vanilla Java
public class Person {
    // ......
    
    // Generated with the custom name instead of toBuilder()
    public PersonBuilder mutate() {
        return new PersonBuilder().name(this.name).age(this.age);
    }

    public static class PersonBuilder {
        // ......
    }
}
```

bengbengbalabalabeng:
If the direction is acceptable, I’d be happy to prepare a PR :)

PR #4062, opened 2026-08-06
[FEATURE] Supports custom toBuilder method name

bengbengbalabalabeng:
## Summary

Related: #4056  (Scope: `@Builder`)

Adds a new `toBuilderMethodName` attribute to `@Builder`, allowing users to customize the name of the generated `toBuilder` method.

  - Explicitly specifying a non-empty `toBuilderMethodName` implicitly enables `toBuilder = true` (i.e., users do not need to redundantly specify `@Builder(toBuilder = true, toBuilderMethodName = "...")`).
  - Defaults to `""` in the annotation, which falls back to `"toBuilder"` (backward compatible — existing code is unaffected).
  - Implemented in both the javac and eclipse/ECJ handlers.

 Adds the config key `lombok.builder.toBuilderMethodName`, which sets the global default whenever `@Builder(toBuilder = true)` is used without an explicit `toBuilderMethodName` parameter.

Resolution order: Explicit annotation attribute (`toBuilderMethodName`) → `lombok.builder.toBuilderMethodName` config key → `"toBuilder"` (default fallback).


nealeu:
Nice.  Really simple and the example of e.g. `.mutate()` as an alternative to `.toBuilder()` could be nice for visual ergonomics.

Rawi01:
My initial thought about it was that this should probably be a config flag instead of an individual attribute. This keeps all builders in sync and leads to a better codebase. Different names for different builders doesn't feel right.

bengbengbalabalabeng:
> My initial thought about it was that this should probably be a config flag instead of an individual attribute. This keeps all builders in sync and leads to a better codebase. Different names for different builders doesn't feel right.

The configuration file approach is indeed a more unified way to keep the behavior of all `builder`s consistent (e.g. `lombok.builder.toBuilderMethodName`). However, I think offering both annotation parameters and configuration options gives users more flexibility to choose in different scenarios.

Regarding the priority between the two, my thoughts are:

-  ~~If the `lombok.builder.toBuilderMethodName` setting exists in the configuration file and the annotation still uses the default value `toBuilder`, then the configuration file value should be used;~~
-  If the `lombok.builder.toBuilderMethodName` property is set in the configuration file and the `toBuilderMethodName` annotation parameter is not explicitly specified, then the configuration file value should be used.
-  If `toBuilderMethodName` is explicitly specified in the annotation, the annotation takes precedence.

This way, we maintain global consistency while still allowing overrides on individual `builders`, striking a balance between flexibility and controllability.

nealeu:
Have you tried this out in different IDEs? 

I am aware of the Jetbrains IDEA Lombok plugin being part of the main IDE git tree now, so they'd need to update that given they maintain an internal autocomplete index.  I suspect they'd be willing given it's been active with work on `@ExtensionMethod` in recent releases.

bengbengbalabalabeng:
> Have you tried this out in different IDEs?
> 
> I am aware of the Jetbrains IDEA Lombok plugin being part of the main IDE git tree now, so they'd need to update that given they maintain an internal autocomplete index. I suspect they'd be willing given it's been active with work on `@ExtensionMethod` in recent releases.

I’ve only tried this in IntelliJ IDEA so far.

Regarding the JetBrains Lombok plugin: I’m not fully sure about the coordination process between Lombok and JetBrains. Would it be appropriate to wait for further progress on this PR before opening a corresponding update request on their side, or would parallel tracking be acceptable?

I’m happy to follow whichever workflow is preferred.

nealeu:
If using your changes in IntelliJ doesn't cause any issues then that's great.  I'll defer to the Lombok team on collaboration with Jetbrains.  I personally see no reason it should get in the way.

bengbengbalabalabeng:
> If using your changes in IntelliJ doesn't cause any issues then that's great. I'll defer to the Lombok team on collaboration with Jetbrains. I personally see no reason it should get in the way.

Sorry for the confusion earlier — what I meant is that using IntelliJ IDEA as the development environment does not cause any issues for compiling or testing the newly added functionality.

bengbengbalabalabeng:
Just a gentle follow‑up — please let me know if there’s anything I can improve or clarify for this PR :)

rspilker:
Hi @bengbengbalabalabeng , I'm still not convinced by the use-case.

Yes, we at Lombok understand that in some scenarios a feature adds value to some people. But we always weigh the potential value to the added cognitive load for the users and the added maintenance load. And in this case, although I understand some fringe scenarios it makes sense, I fear they're not common enough.

I did see this pull request earlier, and the reason I didn't yet respond was that I needed some time to let it sink in.

But I'm still on the fence. I do see some value. And I see the added complexity.

rspilker:
> My initial thought about it was that this should probably be a config flag instead of an individual attribute. This keeps all builders in sync and leads to a better codebase. Different names for different builders doesn't feel right.

I don't necessarily agree. To have a project/package level configuration might make sense in some scenarios (we're writing a math calculator library) but in other scenarios it really should be a per type configuration.

rspilker:
[IFF](https://en.wikipedia.org/wiki/If_and_only_if) we're adding this, the presence of `toBuilderMethodName` in the annotation would imply `toBuilder = true`; if it is in a config key, it still would apply to `toBuilder = true`.

rzwitserloot:
I have a rule:

Every feature should have an obvious, slam-dunk example. If that example is fundamentally broken in some way, it's a bad feature.

This is, to be clear, a 'judge a book by its cover' shortcut: It's merely a guideline, not a hard rule.

But this proposal does fall down based on that metric. The one clear slamdunk example provided is to use the word `mutate()` instead of `toBuilder()` and this is _terrible_: `mutate()` is misleading. Nothing is mutated. Mutate as a word, whether you go by dictionary definition or by what your imagination will serve you up, _extremely strongly implies a thing changing into something else_ and that is not what `toBuilder()` does. If you must paint with the brush of prose, the right image you need to put a word to is the concept of a (possibly not quite perfect) clone budding off. Think Mogwai and the Gremlins. Not Species.

(You might need some knowledge of 80s and 90s movies to get that one).

With implementation notes I was positively inclined to this feature (I think if you omit `toBuilder` but include `toBuilderMethodName`, you should generate the toBuilder; in essence, `toBuilder = (omitted)` means: If `toBuilderMethodName` is present than `true` otherwise `false`) - but I'm __really__ missing a clear and concise example as to why this is needed.

I vaguely recall we decided not to make `toBuilder()` (the name) configurable _because_ of this problem: Every name we could come up with was distasteful. It fell afoul of the abuse rule (if an attempt to think of uses of a lombok feature includes mostly things we consider abuse, in the objective sense of: Makes code harder to maintain, harder to understand, more fragile, then we don't add the feature at all).

Floor's open on this one: Give us some clear use cases where it's nice to rename the method, and the name isn't misleading.

bengbengbalabalabeng:
Hi @rspilker @rzwitserloot,

Thank you very much for the detailed feedback. I completely understand and appreciate your caution regarding maintenance overhead and keeping Lombok's API surface minimal.

First of all, regarding the implementation detail: I completely agree. If `toBuilderMethodName` is explicitly specified in the annotation, it should implicitly imply `toBuilder = true` without requiring redundant declaration.

Secondly, regarding the concern over the naming and semantics of `mutate`: `mutate` is not an arbitrary invention. It has been battle-tested in the Spring Framework across several core modules and has become an established idiom throughout modern Spring projects (such as Spring AI):

Spring (WebMVC / WebFlux / AI):

- `org.springframework.web.servlet.mvc.method.RequestMappingInfo.mutate() -> RequestMappingInfo.Builder`
- `org.springframework.web.server.ServerWebExchange.mutate() -> ServerWebExchange.Builder`
- `org.springframework.http.server.reactive.ServerHttpRequest.mutate() -> ServerHttpRequest.Builder`
- `org.springframework.ai.chat.client.ChatClient.mutate() -> ChatClient.Builder`
- `org.springframework.ai.chat.prompt.PromptTemplate.mutate() -> PromptTemplate.Builder`
- `org.springframework.ai.chat.prompt.ChatOptions.mutate() -> ChatOptions.Builder`
- ...

Its official Javadoc explicitly defines the intent: "Return a builder to create a new instance whose settings are initialized from this instance."

Since many Java developers are already well-accustomed to this pattern through Spring, it is not interpreted as an in-place modification; rather, it is naturally understood as spawning a builder initialized with the current instance.

These examples are not meant to suggest that `mutate` is the only right name under the toBuilder pattern, but rather to demonstrate that such naming does not pose a risk of being misleading.

bengbengbalabalabeng:
> [IFF](https://en.wikipedia.org/wiki/If_and_only_if) we're adding this, the presence of `toBuilderMethodName` in the annotation would imply `toBuilder = true`; if it is in a config key, it still would apply to `toBuilder = true`.

> I think if you omit `toBuilder` but include `toBuilderMethodName`, you should generate the toBuilder; in essence, `toBuilder = (omitted)` means: If `toBuilderMethodName` is present than `true` otherwise `false`

I've updated the implementation and PR description, setting `toBuilderMethodName` in `@Builder` now implicitly enables `toBuilder = true` as suggested. Please take another look when your have time!

PR #4063, opened 2026-08-08
Add fair lock policy to @Locked via @Locked.Policy

vulinh64:
Hi Lombok team,

This is my first contribution, and thank you for taking the time to review it.

This PR adds class-level lock fairness configuration through:

```java
@Locked.Policy(fairness = true)
```

When present, Lombok generates fair locks for the class:

- `@Locked` -> `new ReentrantLock(true)`
- `@Locked.Read` / `@Locked.Write` -> a shared `new ReentrantReadWriteLock(true)`

By default, `ReentrantLock` and `ReentrantReadWriteLock` are non-fair, and Lombok’s existing behavior remains unchanged. The policy only affects locks generated by Lombok; explicitly named locks and user-declared lock fields are deliberately left untouched.

With JDK 25 that fixed the pinned `synchronized` issue, the usage of `@Locked` can still be attractive with the optional fairness policy.

The implementation includes javac and Eclipse support, transform tests for generated and named locks, and documentation/usage-example updates.

I would appreciate feedback on the API and implementation choices.


rzwitserloot:
@rspilker and I spent over an hour discussing this one and crafting alternatives. Our fifth attempt at crafting one might barely be good enough to get over the bar.

But before we / you implement this idea, let's discuss why we want an alternative and what we considered:

## `@Locked.Policy` doesn't feel like the answer.

- It's an annotation that doesn't do anything other than configure how other lombok annotations work. That feels like `lombok.config` territory, except `lombok.config` is for package / module / worktree wide settings, and the notion '__all__ the locks we let lombok generate for us should be fair!' sounds like 'common moronics' (I can foresee companies putting this in code standards / policy documents out of a misguided sense; hey, who doesn't want 'fair', that's like asking for world peace, right? Yes, we all want that!)
- The configuration would apply only to 'the unnamed lock' - if you use named `@Locked`, then the policy will just be there, no warnings or errors or any other signage on it - doing nothing, other than misleading everybody.
- If you put the policy annotation in slightly the wrong place it silently does nothing.
- configuring a single use of an annotation is normally done via attributes of that annotation, but...

## Directly parameterise `@Locked` itself?

- This is weird - what if you have 5 `@Locked` annotations, 3 of which are unnamed, 2 of which are named? Should we just disallow the `fair =` attribute entirely on named `@Locked`? What if 2 of the 3 unnamed `@Locked` annotations have `fair = true`, and the other one has no `fair` at all? Should we just state that missing `fair` means you adopt the fairness setting of whatever `@Locked` annotation in the same type does have an explicit `fair`, and if none do, then it's `false`? What if you have one with `fair = true` and another with `fair = false` (easy enough: Error out). This does mean removing one line affects a completely different one.

## Rejected plan A: `@Locked.Default`

This one is almost, but not quite, acceptable: We introduce `@Locked.Default` instead. This is an annotation you stick on a field. This means: _This_ field is __the__ lock - any unnamed `@Locked` will use this field. The field can be _whatever you want_, we won't check anything about it. We're just going to generate code that calls `.lock()` and `.unlock()` on this thing, and it's your responsibility to ensure the type of this field has these methods. Even opens the door to lock-like things that do not `implement` java's lock but do have `lock()` and `unlock()` methods.

It's nearly good enough but not quite: Faced with this file:

```java
class Example {
   @Locked.Default private Lock lock = new ...Lock(true);

   @Locked void foo() {}
}
```

would a casual reader have even a slight chance of understanding what this means?

## Rejected plan B: The field named `lock` is magic

unnamed `@Locked` will look for a field named `lock`. If it exists, that's the lock that will be used. If it does not exist, a field named `$lock` will be made and that one will be used.

Rejected for many reasons:

- It's not quite backwards compatible. (If you have fields named `lock` right now...) This is a hurdle that can be overcome; anybody that wrote an unnamed `@Locked` and has a `lock` field probably has a bug they didn't know about in the first place. But, it's a downside. Imagine that someone has a field: `boolean lock; // If true, when frobnitzing the doohickey, acquire a lock first`, then we break their code.
- Magic names is a downside. Another hurdle that can be overcome but it's not very nice. We can configure the 'magic name' via `lombok.config`, so this doesn't break those who write their java identifiers in not-english, but, still.
- A `lock` field needs a ton of boilerplate, usually: You don't want it included in your `toString`, it shouldn't be part of the hashcode/equals, and so on. We take care of this with the generated `$lock`. We obviously don't, and can't just assume we should, for an explicitly written out field named `lock`.

## Feasible plan C: Lombok already does this!

Just write your own lock, and then use `@Locked("nameOfField")`. You can then make that lock fair if you want. This is a plausible solution to it all (conveniently, zero changes needed; this already works in lombok). It's obviously not what you want, and it does mean you need to add all the shenanigans to exclude the field from toString, equals, hashCode, and so on.

## Feasible plan D: `@Locked.Generate`

This is an annotation that goes on types, and it does what it says: If you add this annotation, we _will_ generate a lock field. Even if you end up not using it.

If you have `@Locked` usage in a type and you haven't annotated that type with `@Locked.Generate`, then `@Locked.Generate` is implied, with default settings (i.e., not fair). Now we have a place where the `fair = true` can go. If you choose to name your lock differently, you can, and all unnamed `@Locked` in that type will use it. It's similar in practical usage to `@Locked.Policy`, except now it's clearer and less ambiguous as to what it does. It 'generates a lock'.

It still has downsides

- it's a bit hard to tell whether a `@Locked` annotation will defer to the lock generated by a `@Locked.Generate` and when it will generate its own (for example, if you place `@Locked.Generate` on a superclass expecting subclass `@Locked` to use it, you'll be confused, as it won't).

```java
@Locked.Generate(fair = true)
class Example {
  class Inner {
    @Locked void foo() { }
  }
}
```

- You HAVE TO tell `@Locked.Generate` whether it's a readwrite lock, or a normal lock. And I can't come up with a good name for this parameter. `readwrite = true`? As in, this will be a mandatory parameter.
- 
- If you want  `@Locked.Generate` to generate a static lock you have to tell it. Not doing so and using `@Locked` on a static method will cause compiler errors. But this is actually a good thing: You can now mix `@Locked.Generate(staticLock = true)` on a class and stick `@Locked` on an instance method, and now you have a static lock, which is a plausible scenario and is now a lot easier (before you'd have to write an explicit lock and use named `@Locked`, though because the explicit lock is static you don't have to put in as much effort to exclude it from toString and company). It also avoids confusion - right now if you stick `Locked` on both instance and static methods you get two completely unrelated locks. This would be an exotic backwards break, but one we will accept (because it's likely that any code we break was already buggy. Even if it isn't, it's quite confusing).

----

Our current thinking is option C - this is a lot of bother for a fairly exotic feature that has a ton of caveats.

Issue #408, opened 2015-07-14
@Getter/@Setter website page doesn't match behaviour


lombokissues:
_Migrated from Google Code (issue 335)_

PR #4067, opened 2026-08-17
[#3107] Copy @Deprecated(since, forRemoval) onto generated accessors

### LINKED ISSUE #3107: [BUG] forRemoval parameter in Deprecated annotation is not propagated
**Describe the bug**
forRemoval parameter in Deprecated annotation is not propagated

**To Reproduce**
Just annotate a property with @Deprecated(forRemoval=true) and check annotation in related class and builder methods.

**Expected behavior**
Propagated annotation should have annotation parameters

 - Lombok version 1.18.22

arimu1:
Fixes #3107

## Summary
- Generated getters, setters, withers, and `@Builder`/`@SuperBuilder` setters (including `@Singular` methods) now copy `since` and `forRemoval` from a field's `@Deprecated` annotation instead of always emitting a marker `@Deprecated`.
- javac copies the existing member values; Eclipse copies only `since` / `forRemoval`. Generated Eclipse member pairs use a zero-length source range and `setGeneratedBy` (same approach as other synthetic annotations; see #408). Marker `@Deprecated` and javadoc-only `@deprecated` are unchanged.
- Both `TestEclipse` and `test.ecj*` boot JDK 8 `java.lang.Deprecated` (no members), so the new cases are `//platform javac` only.

## Test plan
- [x] `ant test.javacCurrent` on Temurin 21 (`GetterDeprecatedMembers`, `SetterDeprecatedMembers`, `BuilderWithDeprecatedMembers`, plus the rest of the javac transform suite)
- [x] Neighbor Eclipse cases still pass (`GetterDeprecated`, `WithMethodMarkedDeprecated`)
- [ ] Maintainer: confirm Eclipse JDT on Java 9+ sources shows `@Deprecated(since=…, forRemoval=…)` on generated accessors without AST⇔DOM range errors


rzwitserloot:
The idea LGTM. We'll review it... except you've missed one crucial step: You need to add your name to the `AUTHORS` file. Could you add/update the PR? Or if you're already in there, comment which name is you? Thanks, @arimu1

arimu1:
Added `arimu1 <19286898+arimu1@users.noreply.github.com>` to `AUTHORS` and rebased onto current `master` (kept the new `CheckReturnValue` path in `HandleGetter`, still using `generateDeprecatedAnnotation` so `since`/`forRemoval` copy).

PR #4071, opened 2026-08-21
Fix SuperBuilder for non-ASCII superclass simple names

### LINKED ISSUE #4070: [BUG] @SuperBuilder fails when a superclass has a non-ASCII (Korean/Hangul) name
## Describe the bug

`@SuperBuilder` fails during javac compilation when a child class extends a superclass whose Java identifier contains non-ASCII characters, such as Korean (Hangul).

Both classes are in the same package and both are annotated with `@SuperBuilder`.

The inheritance itself is valid Java and compiles without `@SuperBuilder`. It also compiles when the child references the superclass using its fully qualified class name.

## To reproduce

Create a Gradle Java project using UTF-8.

### `build.gradle`

```groovy
plugins {
    id 'java'
}

repositories {
    mavenCentral()
}

java {
    toolchain {
        languageVersion = JavaLanguageVersion.of(21)
    }
}

dependencies {
    compileOnly 'org.projectlombok:lombok:1.18.46'
    annotationProcessor 'org.projectlombok:lombok:1.18.46'
}

tasks.withType(JavaCompile).configureEach {
    options.encoding = 'UTF-8'
}
```

### `src/main/java/repro/부모.java`

```java
package repro;

import lombok.experimental.SuperBuilder;

@SuperBuilder
public class 부모 {
    protected Long a;
}
```

### `src/main/java/repro/자식.java`

```java
package repro;

import lombok.experimental.SuperBuilder;

@SuperBuilder
public class 자식 extends 부모 {
    private String b;
}
```

Run:

```shell
./gradlew compileJava
```

Image

## Workaround

Using the fully qualified superclass name in the `extends` clause compiles successfully:

```java
@SuperBuilder
public class 자식 extends repro.부모 {
    private String b;
}
```

Adding an import does not help because the generated code still fails when the superclass is referenced by its simple name.

## Expected behavior

`@SuperBuilder` should generate the superclass builder reference correctly when the valid Java superclass identifier contains non-ASCII characters.

The simple-name and fully qualified-name forms should behave identically.

## Version info

- Lombok version: 1.18.46
- Platform: javac 21.0.7
- Java runtime: Java 21.0.7
- Gradle: 8.14.3
- OS: Windows

## Additional context

Other `@SuperBuilder` hierarchies whose superclass names contain only ASCII characters compile successfully in the same project.

This may be related to the different superclass handling paths in `HandleSuperBuilder`. The simple-name path converts the extends clause to a string and rebuilds the type reference, while the fully qualified-name path preserves the AST structure:

kalayciburak:
## Summary
- `@SuperBuilder` on a subclass whose superclass is referenced by a non-ASCII simple name (for example Hangul) failed during javac compilation.
- The javac handler built the superclass builder type from `extendsClause.toString()`. javac pretty-prints non-ASCII identifiers as unicode escapes, so the generated builder type used those escape characters as the name.
- The qualified-name path already used `JCFieldAccess.getIdentifier()`. The simple-name path now uses `JCIdent.getName()` the same way.

Fixes #4070

## Tests
- `ant test.javacCurrent` (JDK 17): `javac-SuperBuilderNonAscii.java` PASS
- same run: `javac-SuperBuilderAbstract.java` and `javac-SuperBuilderWithNonNull.java` PASS (ASCII simple-name / existing SuperBuilder regression)

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -6,6 +6,7 @@ Amine Touzani <ttzn.dev@gmail.com>
 Andre Brait <andrebrait@gmail.com>
 Anshuman Mishra <a.mishra@uber.com>
 Bulgakov Alexander <buls@yandex.ru>
+Burak KALAYCI <kalayciburak1996@gmail.com>
 Caleb Brinkman <floralvikings@gmail.com>
 Christian Nüssgens <christian@nuessgens.com>
 Christian Schlichtherle <christian-schlichtherle@users.noreply.github.com>
--- a/src/core/lombok/javac/handlers/HandleSuperBuilder.java
+++ b/src/core/lombok/javac/handlers/HandleSuperBuilder.java
@@ -261,6 +261,13 @@ public class HandleSuperBuilder extends JavacAnnotationHandler<SuperBuilder> {
 			String builderClassNameTemplate = BuilderJob.getBuilderClassNameTemplate(annotationNode, null);
 			String superclassBuilderClassName = job.replaceBuilderClassName(superclassName.toString(), builderClassNameTemplate);
 			superclassBuilderClass = parent.getTreeMaker().Select(cloneType(maker, (JCFieldAccess) extendsClause, annotationNode), parent.toName(superclassBuilderClassName));
+		} else if (extendsClause instanceof JCIdent) {
+			// Use the identifier Name, not Tree.toString(): javac pretty-prints non-ASCII
+			// names as \\uXXXX escapes, which then become the builder type's simple name.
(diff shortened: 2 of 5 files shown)


rzwitserloot:
Added some tests for the complex case which already worked. @rspilker is working on a simple system that 'knows' we are in lombok tests; the `else` clause still uses `toString()` and thus will break: Your addition means we should never get there, but we're leaving it in as a fallback mechanism. Still, if we ever run into the fallback during tests, the test should just fail.

standard disclaimer: github does not understand our branching policies so thinks this PR was rejected. It's in master.

PR #4073, opened 2026-08-23
Fix parameterized return type inference for @ExtensionMethod in Eclipse

### LINKED ISSUE #4060: [BUG] ExtensionMethod does not have the correct parameterized return type (Eclipse-only)
**Description**

In Eclipse, an `@ExtensionMethod` that has a parameterized return type does not yield the correct concrete return type when used. Instead, `Object` is used as type argument, resulting in compilation errors.

`javac` works fine.

**To Reproduce**
```
@ExtensionMethod({Usage.MapperExtensionMethods.class})
public class Usage {
	public void test() {
		List<String> stringList = List.of("");
		process(stringList.mapAllToList(s -> s + ";"));
	}

	private void process(List<String> list) {
	}

	public class MapperExtensionMethods {
		public static <V, R> List<R> mapAllToList(Collection<? extends V> values, Function<V, R> mapper) {
			if (values == null) {
				return new ArrayList<>();
			}
			return values.stream()
				.map(mapper)
				.toList();
		}
	}
}
```

results in the error `The method process(List<String>) in the type Usage is not applicable for the arguments (List<Object>)`.

**Expected behavior**
Compiles correctly.

**Version info**
 - Lombok v1.18.46
 - Eclipse 2026-06 (4.40.0)

**Additional context**
Workaround: Adding a manual cast fixes the error: `process((List<String>)stringList.mapAllToList(s -> s + ";"));`

cfcromn:
Fixes #4060

### What changed

Fix parameterized return type inference for `@ExtensionMethod` calls in Eclipse.

When an extension method has a parameterized generic return type, such as `List<R>`, ECJ may resolve the provisional return type as `List<Object>` before the surrounding target type can participate in type inference.

This change preserves the invocation as a poly expression so ECJ can continue inference using the outer target type.

A regression test based on the reproduction from #4060 is included.

### Testing

- `ant test.eclipse-202503`
- `ant test`
- `ant dist`

All passed.

The relevant `@ExtensionMethod` tests also pass against the latest Eclipse integration build. The full integration-build test suite still reports unrelated existing expected-output differences.

### DIFF
--- a/src/eclipseAgent/lombok/eclipse/agent/PatchExtensionMethod.java
+++ b/src/eclipseAgent/lombok/eclipse/agent/PatchExtensionMethod.java
@@ -319,6 +319,7 @@ public class PatchExtensionMethod {
 				}
 				
 				MethodBinding fixedBinding = scope.getMethod(extensionMethod.declaringClass, methodCall.selector, argumentTypes.toArray(new TypeBinding[0]), methodCall);
+				boolean polyExpression = false;
 				if (fixedBinding instanceof ProblemMethodBinding) {
 					methodCall.arguments = originalArgs;
 					// Sometimes the declaring class is null, in that case we have to create a new ProblemMethodBinding using the extension method's declaring class
@@ -327,11 +328,12 @@ public class PatchExtensionMethod {
 					}
 					PostponedInvalidMethodError.invoke(scope.problemReporter(), methodCall, fixedBinding, scope);
 				} else {
+					polyExpression = fixedBinding.returnType.isParameterizedType() && methodCall.isPolyExpression(fixedBinding);
 					// If the extension method uses varargs, the last fixed binding parameter is an array but 
 					// the method arguments are not. Even thought we already know that the method is fine we still
 					// have to compare each parameter with the type of the array to support autoboxing/unboxing.
 					boolean isVarargs = fixedBinding.isVarargs();
-					for (int i = 0, iend = arguments.size(); i < iend; i++) {
+					for (int i = 0, iend = polyExpression ? 0 : arguments.size(); i < iend; i++) {
 						Expression arg = arguments.get(i);
 						TypeBinding[] parameters = fixedBinding.parameters;
 						TypeBinding param;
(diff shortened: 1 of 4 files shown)


Rawi01:
I already had a short debugging session and ended up with returning a poly bindung too. I'm short on time right now but should be able to review it in the next few days.

src/eclipseAgent/lombok/eclipse/agent/PatchExtensionMethod.java
```diff
@@ -327,11 +328,12 @@ public static TypeBinding resolveType(TypeBinding resolvedType, MessageSend meth
 					}
 					PostponedInvalidMethodError.invoke(scope.problemReporter(), methodCall, fixedBinding, scope);
 				} else {
+					polyExpression = fixedBinding.returnType.isParameterizedType() && methodCall.isPolyExpression(fixedBinding);
```
Rawi01: We need to use reflection to call `MessageSend::isPolyExpression(MethodBinding)` to support old ecj versions

src/eclipseAgent/lombok/eclipse/agent/PatchExtensionMethod.java
```diff
@@ -327,11 +328,12 @@ public static TypeBinding resolveType(TypeBinding resolvedType, MessageSend meth
...
 					// If the extension method uses varargs, the last fixed binding parameter is an array but 
 					// the method arguments are not. Even thought we already know that the method is fine we still
 					// have to compare each parameter with the type of the array to support autoboxing/unboxing.
 					boolean isVarargs = fixedBinding.isVarargs();
-					for (int i = 0, iend = arguments.size(); i < iend; i++) {
+					for (int i = 0, iend = polyExpression ? 0 : arguments.size(); i < iend; i++) {
```
Rawi01: Skipping the argument handling by adding a tenary expression might be the shortest solution but feels like a hack. Can this be moved to a more expicit if statement (and maybe merged with if at the end of this method)

Rawi01 (changes requested):

src/eclipseAgent/lombok/eclipse/agent/PatchExtensionMethod.java
cfcromn: Updated. The ternary loop bound has been removed and the argument handling is now guarded by an explicit if (!polyExpression) branch.

src/eclipseAgent/lombok/eclipse/agent/PatchExtensionMethod.java
cfcromn: Updated to call MessageSend::isPolyExpression(MethodBinding) through Permit.permissiveGetMethod, with a fallback for ECJ versions where the method is unavailable. I also verified this against the Eclipse Oxygen test target.

cfcromn:
Addressed both review comments in 5aeafcfd.

* Switched the isPolyExpression(MethodBinding) call to reflection for compatibility with older ECJ versions.
* Reworked the argument handling to use an explicit conditional instead of the ternary loop bound.

Validation:

* ant test.eclipse-202503 passed
* ant test.eclipse-oxygen passed
* git diff --check passed

Thanks for the review!

cfcromn:
@Rawi01  gentle ping when you have a chance 👀

PR #4074, opened 2026-08-24
Support JDK 27: end positions moved from EndPosTable to JCTree.endpos

### LINKED ISSUE #4072: [BUG] Lombok 1.18.46 fails on JDK 27: ClassNotFoundException com.sun.tools.javac.tree.EndPosTable
**Describe the bug**

Lombok 1.18.46 does not work on JDK 27. `LombokProcessor.init()` throws `ExceptionInInitializerError`, caused by a `ClassNotFoundException` for `com.sun.tools.javac.tree.EndPosTable`, which was removed in JDK 27:

- JDK 26: https://github.com/openjdk/jdk/blob/jdk26/src/jdk.compiler/share/classes/com/sun/tools/javac/tree/EndPosTable.java
- JDK 27: https://github.com/openjdk/jdk/blob/jdk27/src/jdk.compiler/share/classes/com/sun/tools/javac/tree/EndPosTable.java (404)

Since this happens during processor initialization, all compilation fails — including files with no Lombok annotations.

**To Reproduce**

`Example.java`:

```java
import lombok.Getter;

public class Example {
    @Getter private String name;
}
```

```
javac -proc:full -cp lombok.jar Example.java
```

```
An annotation processor threw an uncaught exception.
Consult the following stack trace for details.
java.lang.ExceptionInInitializerError
	at lombok.javac.apt.LombokProcessor.placePostCompileAndDontMakeForceRoundDummiesHook(LombokProcessor.java:174)
	at lombok.javac.apt.LombokProcessor.init(LombokProcessor.java:96)
	at lombok.core.AnnotationProcessor$JavacDescriptor.want(AnnotationProcessor.java:160)
	at lombok.core.AnnotationProcessor.init(AnnotationProcessor.java:213)
	at lombok.launch.AnnotationProcessorHider$AnnotationProcessor.init(AnnotationProcessor.java:64)
	at jdk.compiler/com.sun.tools.javac.processing.JavacProcessingEnvironment$ProcessorState.<init>(JavacProcessingEnvironment.java:633)
	at jdk.compiler/com.sun.tools.javac.processing.JavacProcessingEnvironment$DiscoveredProcessors$ProcessorStateIterator.next(JavacProcessingEnvironment.java:757)
	at jdk.compiler/com.sun.tools.javac.processing.JavacProcessingEnvironment.discoverAndRunProcs(JavacProcessingEnvironment.java:852)
	at jdk.compiler/com.sun.tools.javac.processing.JavacProcessingEnvironment$Round.run(JavacProcessingEnvironment.java:1183)
	at jdk.compiler/com.sun.tools.javac.processing.JavacProcessingEnvironment.doProcessing(JavacProcessingEnvironment.java:1304)
	at jdk.compiler/com.sun.tools.javac.main.JavaCompiler.processAnnotations(JavaCompiler.java:1265)
Caused by: java.lang.ClassNotFoundException: com.sun.tools.javac.tree.EndPosTable
	at java.base/java.lang.ClassLoader.findClass(ClassLoader.java:685)
	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:569)
	at lombok.launch.ShadowClassLoader.loadClass(ShadowClassLoader.java:565)
	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:502)
	at java.base/java.lang.Class.forName0(Native Method)
	at java.base/java.lang.Class.forName(Class.java:479)
	at java.base/java.lang.Class.forName(Class.java:469)
	at lombok.javac.Javac.getMethod(Javac.java:260)
	at lombok.javac.Javac.<clinit>(Javac.java:223)
	... 16 more
```

No class file is produced.

**Expected behavior**

`Example.java` compiles and `Example` gets a generated `getName()`, as it does on JDK 26 and earlier.

- Lombok version: **1.18.46**
- Platform: **javac**, `javac 27` (`openjdk version "27" 2026-09-15`, build `27+35-2325`, `openjdk-27_macos-aarch64_bin.tar.gz` from download.java.net), macOS aarch64

**Additional context**

- JDK 25 and 26 are unaffected; the same file compiles and the getter is generated.
- Also fails under Gradle 9.7 with a Java 27 toolchain.

danishnawab:
Fixes #4072.

JDK 27 removed `com.sun.tools.javac.tree.EndPosTable` and stores end positions directly on tree nodes ([JDK-8372948](https://bugs.openjdk.org/browse/JDK-8372948)). Lombok resolved that class in `Javac`'s static initializer and rethrew on failure, so `LombokProcessor.init()` died with an `ExceptionInInitializerError` and every compilation on JDK 27 failed, including files with no lombok annotations.

### Changes

- `Javac`: probe for `EndPosTable` instead of failing when it is missing. When absent, read end positions via the no-arg `DiagnosticPosition.getEndPosition()` and store them by writing `JCTree.endpos`. Detection is by class presence rather than version number, so 27-EA builds from before and after the javac change both work.
- `CommentCatcher`: only set `compiler.genEndPos` on javacs that still have that field. JDK 27 always records end positions.
- New `lombok.javac.java27` parser factory and parser: JDK 27 dropped the `keepEndPositions` parameter from `JavacParser`'s constructors and `ParserFactory.newParser`, so the `java9` variants no longer link. Follows the existing `java6`/`java7`/`java8`/`java9` pattern.
- `DocCommentIntegrator`: implement the new no-arg `getEndPosition()` on the anonymous `DiagnosticPosition`.
- `PrettyPrinter`: removed a reflective `getEndPosition` lookup that would fail class init on JDK 27. The looked-up `Method` was never invoked, so this is a dead code removal rather than a behavior change.
- Added an opt-in `test.javac27` ant target. Not wired into `test.broad` since 27 is not yet released.

### Testing

- `test.javacCurrent`: 862 pass / 0 fail on JDK 27 and JDK 26, 858 pass / 0 fail on JDK 21
- delombok output on JDK 27 is byte identical to 26/21/17, including javadoc reattachment and a `module-info.java` round trip (the `parseModuleInfo` constructor argument has no existing test coverage)
- annotations verified against JDK 8, 11, 17, 21, 25, 26 and 27

### Note

javac clamps stored end positions to `errorEndPos`. Lombok cannot see that parser state on JDK 27, so `storeEnd` writes the raw position. This only matters for generated nodes in files that already have syntax errors.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -11,6 +11,7 @@ Christian Nüssgens <christian@nuessgens.com>
 Christian Schlichtherle <christian-schlichtherle@users.noreply.github.com>
 Christian Sterzl <christian.sterzl@gmail.com>
 Christoph Dreis <christoph.dreis@freenet.de>
+Danish Nawab <danish.nawab@sixt.com>
 DaveLaw <project.lombok@apconsult.de>
 Dave Brosius <dbrosius@mebigfatguy.com>
 Dawid Rusin <dawidrusin90@gmail.com>
--- a/buildScripts/tests.ant.xml
+++ b/buildScripts/tests.ant.xml
@@ -167,6 +167,11 @@ This buildfile is part of projectlombok.org. It takes care of compiling and runn
 		<test.javacX version="26" />
 	</target>
 	
+	<!-- 27 is non-LTS, not yet released -->
+	<target name="test.javac27" depends="test.compile, test.formatter.compile" description="runs the tests on your default VM, using javac27 as underlying compiler">
+		<test.javacX version="27" />
+	</target>
+	
 	<target name="test.javacCurrent" depends="test.compile, test.formatter.compile" description="runs the tests on your default VM, using its javac as underlying compiler">
 		<echo>Running TestJavac on JVM${ant.java.version}, with the javac built into your VM distributon.</echo>
 		<junit haltonfailure="yes" fork="true" forkmode="once">
--- a/doc/changelog.markdown
+++ b/doc/changelog.markdown
@@ -2,6 +2,7 @@ Lombok Changelog
 ----------------
 
 ### v1.18.47 "Edgy Guinea Pig"
+* PLATFORM: JDK27 support added [#4072](https://github.com/projectlombok/lombok/issues/4072).
 * BUGFIX: `@SneakyThrows` usage on JDK26 no longer results in class files that require `lombok.jar` to be on the runtime classpath (which should not be neccessary). [#4040](https://github.com/projectlombok/lombok/issues/4022).
 * FEATURE: New [config key](https://projectlombok.org/features/configuration) `lombok.checkReturnValueAnnotation` (values: `none`, `lombok`; default: `none`) lets lombok generate `@lombok.CheckReturnValue` on generated methods where the return value should not be ignored, such as `@With` methods and `@Builder.build()`. A future lombok release may flip the default to `lombok`. [#4013](https://github.com/projectlombok/lombok/pull/4013).
 * PROMOTION: `@SuperBuilder` has been promoted to the main package. Otherwise, no changes have been made to the annotation. The old experimental annotation will remain for a few versions, at which point it will be marked as a deprecated annotation. Eventually it'll be removed. If you had `lombok.config` configuration for this annotation, the configuration keys for this feature have been renamed.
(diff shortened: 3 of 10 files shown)


rzwitserloot:
This was awesome work, @danishnawab. I stuck one more commit on top that cleaned up some of the code (mostly what was there, not even what you added). The tests on current JDK27-ea (on my machine, Your kilometerage may vary) are 100% pass, so the rather ambitious change log message that this 'fixes lombok to be compatible for JDK27' seems fair enough!

